PR #3, opened 2015-10-26
middleware.Cancel

VojtechVitek:
Something for consideration:

`middleware.Timeout()` and `middleware.CloseNotify` don't need to `cancel()` the context when the underlying handlers return. We might want to use explicit `middleware.Cancel` instead.

This way, all handlers would have `context.Background()` as a base context and it'd be up to developers to specify when the "context flow" ends. Timeout should only cancel underlying contexts on timeout. CloseNotify should only cancel underlying contexts on `<-w.CloseNotify()`. Explicit over magic.

If we wanted to have strict boundaries for context lifetime in the handler chain, we'd need to use `context.WithCancel()` instead of `context.Background()` in [mux.ServeHTTPC()](https://github.com/pressly/chi/blob/6cdb0fa9e298c912a8a9382e461066d2fa5fc36e/mux.go#L184).

### DIFF
--- a/middleware/cancel.go
+++ b/middleware/cancel.go
@@ -1,0 +1,24 @@
+package middleware
+
+import (
+	"net/http"
+
+	"github.com/pressly/chi"
+	"golang.org/x/net/context"
+)
+
+// Cancel is a middleware that cancels ctx when the underlying handlers
+// return. It enforces any running goroutines (or ctxhttp clients)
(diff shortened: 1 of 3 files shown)


VojtechVitek:
Closing this PR -- I don't think we need this anymore, since Timeout and CloseNotify close their contexts in defer functions, which is imho the correct behavior.

PR #7, opened 2015-10-31
Add Makefile

vladdy:
I'm preparing another pull request with basic instructions on how to contribute to this project and the makefile should make it easier to work with it for those who develop in command line.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,26 @@
+# Makefile for Chi (https://github.com/pressly/chi)
+#
+# Targets:
+#	all (default): Builds and tests the code
+#	build: Builds the code
+#	test: Runs the tests
+#	fmt: Formats the source files
+
+# Build settings
+GOCMD?=go
+# Allow setting of Go flags via the command line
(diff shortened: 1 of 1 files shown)


VojtechVitek:
Even though I really like Makefiles for larger projects, I don't think it's necessary for project like `chi`. Is there any other simple router out there with Makefile?.

We can run `go build`, `go test` or `go fmt` directly in the single directory. There's no hidden magic that'd require us to run make. Thoughts?

vladdy:
Yes, we can just use go commands and chi is not that big project (yet), but, for example, "make" can be set to default `go build && go test && go fmt` later when you notice a lot of issues with formatting and it won't be required to update the build instructions or change workflow then. 

Anyway, your point also makes sense, so if you are still not sold for this workflow, just close this pull request and I'll update  #8 with raw go commands.

Thanks for quick response!

pkieltyka:
@vladdy thanks for the PR, but we'd like to keep things as lean as possible.

PR #9, opened 2015-10-31
Support custom Not Found handler

vladdy:
Basic support for custom NotFound handler. Custom handler is set via  NotFoundHandler method:

```Go
m = chi.NewRouter()
m.NotFoundHandler(func(rw http.ResponseWriter, r *http.Request) {
   rw.WriteHeader(200)
})

```
It is yet to decide how this will work for additional mounted routers.

### DIFF
--- a/mux.go
+++ b/mux.go
@@ -13,7 +13,7 @@ type Mux struct {
 	routes      map[methodTyp]*tree
 
 	notFound http.HandlerFunc
-	
+
 	// can add rules here for how the mux should work..
 	// ie. slashes, case insensitive, 500 error handler etc.. like httprouter
 }
@@ -109,9 +109,9 @@ func (mx *Mux) Options(pattern string, handlers ...interface{}) {
 	mx.handle(mOPTIONS, pattern, handlers...)
 }
 
-// NotFoundHandler sets a custom handler for the case when no routes matched
+// NotFound sets a custom handler for the case when no routes matched
 // the given URL
-func (mx *Mux) NotFoundHandler(handler http.HandlerFunc) {
+func (mx *Mux) NotFound(handler http.HandlerFunc) {
 	mx.notFound = handler
 }
 
--- a/mux.go
+++ b/mux.go
@@ -12,8 +12,10 @@ type Mux struct {
 	middlewares []interface{}
 	routes      map[methodTyp]*tree
 
+	notFound http.HandlerFunc
+	
 	// can add rules here for how the mux should work..
-	// ie. slashes, case insensitive, notfound handler etc.. like httprouter
+	// ie. slashes, case insensitive, 500 error handler etc.. like httprouter
 }
 
 type methodTyp int
@@ -107,6 +109,12 @@ func (mx *Mux) Options(pattern string, handlers ...interface{}) {
 	mx.handle(mOPTIONS, pattern, handlers...)
 }
 
+// NotFoundHandler sets a custom handler for the case when no routes matched
+// the given URL
+func (mx *Mux) NotFoundHandler(handler http.HandlerFunc) {
+	mx.notFound = handler
+}
+
 func (mx *Mux) handle(method methodTyp, pattern string, handlers ...interface{}) {
 	// Build handler from middleware stack, inline middlewares and handler
 	h := chain(mx.middlewares, handlers...)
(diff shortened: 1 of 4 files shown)


mux_test.go
```diff
@@ -393,4 +393,31 @@ func TestMux(t *testing.T) {
...
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode != 200 {
+		t.Error("expecting custom NotFound handler when it is set")
+	}
+
+
```
VojtechVitek: Did you run go fmt?

mux_test.go
```diff
@@ -393,4 +393,31 @@ func TestMux(t *testing.T) {
...
+		t.Fatal(err)
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode != 404 {
+		t.Error("expecting default NotFound error when custom NotFound handler is not set")
+	}
+
+	// Custom NotFound handler
+	m.NotFoundHandler(func(rw http.ResponseWriter, r *http.Request) {
```
VojtechVitek: Just a nit: Can you use `w` instead of `rw`, as in the rest of the codebase?

mux.go
```diff
@@ -107,6 +109,12 @@ func (mx *Mux) Options(pattern string, handlers ...interface{}) {
 	mx.handle(mOPTIONS, pattern, handlers...)
 }
 
+// NotFoundHandler sets a custom handler for the case when no routes matched
+// the given URL
+func (mx *Mux) NotFoundHandler(handler http.HandlerFunc) {
```
VojtechVitek: Maybe we could call this just `NotFound(handler http.HandlerFunc)`, as goji or httprouter do.

@pkieltyka what do you think about the API?

VojtechVitek:
LGTM, besides the minor comments. Thanks for you contribution!

mux.go
pkieltyka: I prefer `NotFound()` as well.. however, the handler arg should be a http.Handler, chi.Handler or interface{} where we can support various handler types. The other option is to have custom NotFound handlers as a middleware instead.

vladdy: I could make it to be just interface{} but I'm concerned about providing context for URL that will be missing which can be confusing. 

Also, if NotFound is made as middleware do you have idea how it will know about non existing addresses for middleware chained after it? 

Alternatively, I see NotFound handler to be a part of context. In such case NotFound middleware can change it from some default value.

vladdy:
I could make the handler to accept interface{} but I'm concerned about providing a context for URL that will be missing which can be confusing.

Also, if NotFound is made as a middleware, do you have idea how it will know about non existing addresses for middlewares chained after it?

Alternatively, I see NotFound handler to be a part of the context. In such case NotFound middleware can change it from some default value. What do you think?

VojtechVitek:
@pkieltyka I'm not 100% convinced that `ctx` is useful here. If the router doesn't find any matching route -- then there's no context whatsoever, is it? This 404 handler represents `"route not found"`, which is a different thing than end-handler's `"resource not found"` (which has `ctx` of course). We're not trying to solve the latter one.

I think that all the sub-routers should inherit their parent's `NotFound` handler. Thinking of it... I'd actually prefer doing pkg-wide global variable instead of putting handler into every `Mux`. This would be very close to stdlib's [`http.NotFound`](https://github.com/golang/go/blob/master/src/net/http/server.go#L1449-L1454) - lightweight, but configurable.

Something like:

chi.go:
```go
var NotFound func(w http.ResponseWriter, r *http.Request) = http.NotFound // defaults to stdlib's NotFound
```

users-end-handler.go:
```go
chi.NotFound = MyNotFoundHandler // router 404 will serve this handler instead
```

pkieltyka:
@VojtechVitek cool ideas. A package level handler is interesting and does simply the case where there is a single service powered by a stack of routers. But if for some reason someone wanted to have multiple "services" running defined by a few indepdent chi.Router's .. then they're screwed because they are all using the same notfound handler. So I tend to think thats not ideal in the end.

Btw, the ctx is passed to even routes that aren't matched.

for example..

```go
r := chi.NewRouter()
r.Use(a)
r.Use(b)
r.Get("/dontmatchthis", ...)
```

.. for any not-found route, the request will run through the "a" and "b" middlewares. This is supported, by design. Remember we came across this for other things? ... so meaning, there could be a ctx passed down to NotFound ..

edigu:
any news on this?

pkieltyka:
@edigu have a look at https://github.com/pressly/chi/pull/30 and let me know what you think. Let's move the discussion to that PR.

pkieltyka:
Closed in b9ba24fa6c5ac5b949cc7beaea5f80e4c5c5d3a0

PR #10, opened 2015-11-02
Fixed panics on empty paths

utrack:
Mounting leaves with empty paths like ```r.Get("",func...)``` panics with index out of range errors.
Sometimes it happens even with sane paths w/ many subgroups, I'm not sure why; however, this commit fixes that.

This commit converts empty paths to single slashes, so "" becomes "/".

### DIFF
--- a/mux.go
+++ b/mux.go
@@ -111,6 +111,9 @@ func (mx *Mux) handle(method methodTyp, pattern string, handlers ...interface{})
 	// Build handler from middleware stack, inline middlewares and handler
 	h := chain(mx.middlewares, handlers...)
 
+	if len(pattern) == 0 {
+		pattern = "/"
+	}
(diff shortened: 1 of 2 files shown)


pkieltyka:
thanks for this, however, I'd like to require that at a minimum "/" is specified for the routing pattern.

VojtechVitek:
@utrack can you change it to `panic()` instead?

pkieltyka:
yea, I think `if len(pattern) == 0 || pattern[0] != '/' {` is good

utrack:
@VojtechVitek I tried to fix the panic :) next if threw panic because of index out of range, which was kinda counter-intuitive for an error. @pkieltyka's solution should work though

@pkieltyka: OK, I'll try to isolate case when it happens with non-empty initial paths and submit a separate issue

PR #14, opened 2015-11-05
small benchmark (chi vs httprouter)

sergeyt:
the code is extracted from https://github.com/julienschmidt/go-http-routing-benchmark

### DIFF
--- a/bench_test.go
+++ b/bench_test.go
@@ -1,0 +1,435 @@
+package chi
+
+import (
+	"log"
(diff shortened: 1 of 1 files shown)


pkieltyka:
thank you for this! it inspired me to make a full benchmark suite at https://github.com/pkieltyka/go-http-routing-benchmark and posted the results at https://github.com/pressly/chi#benchmarks

PR #16, opened 2015-11-11
Use a single log line per request, very similar to Apache-style logs

gdm85:
This also means that count of bytes written is introduced in the log line.

The choice of colors is pretty much arbitrary, so if you have better ideas please say so :blush:

### DIFF
--- a/middleware/logger.go
+++ b/middleware/logger.go
@@ -6,6 +6,7 @@ package middleware
 import (
 	"bufio"
 	"bytes"
+	"fmt"
 	"io"
 	"log"
 	"net"
@@ -25,8 +26,7 @@ import (
 func Logger(next chi.Handler) chi.Handler {
 	fn := func(ctx context.Context, w http.ResponseWriter, r *http.Request) {
 		reqID := GetReqID(ctx)
-		printStart(reqID, r)
-
+		prefix := requestPrefix(reqID, r)
 		lw := wrapWriter(w)
 
 		t1 := time.Now()
@@ -34,7 +34,7 @@ func Logger(next chi.Handler) chi.Handler {
 		defer func() {
 			t2 := time.Now()
 
-			printEnd(reqID, lw, t2.Sub(t1))
+			printEnd(prefix, reqID, lw, t2.Sub(t1))
 		}()
 
 		next.ServeHTTPC(ctx, lw, r)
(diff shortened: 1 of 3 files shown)


pkieltyka:
@gdm85 cool, the single log line is definitely cleaner.

The only upside to 2 lines is you can see when the initial request came in and then the request response. In case of a request+panic, one wouldn't see when the initial request and its details came in. What do you think?

gdm85:
@pkieltyka I think the Recoverer middleware should work as much as possible stand-alone without the Logger, maybe the request context could be added to the panic message written by the recoverer as well?

GeertJohan:
@pkieltyka If we can switch the Logger and Recoverer middleware, then the Logger will still print the panic outcome before the Recoverer recovers the request.

middleware/logger.go
```diff
@@ -43,34 +42,27 @@ func Logger(next chi.Handler) chi.Handler {
...
 	if reqID != "" {
 		cW(&buf, nYellow, "[%s] ", reqID)
 	}
-	buf.WriteString("Started ")
+	cW(&buf, nBlue, "\"")
 	cW(&buf, bMagenta, "%s ", r.Method)
-	cW(&buf, nBlue, "%q ", r.URL.String())
-	buf.WriteString("from ")
-	buf.WriteString(r.RemoteAddr)
+	cW(&buf, nBlue, fmt.Sprintf("%q ", r.URL.String()+" "+r.Proto)[1:])
```
VojtechVitek: > `../../middleware/logger.go:54: undefined: fmt in fmt.Sprintf`

Can you run `goimports`?

Also, what's the `[1:]` slice for?

middleware/logger.go
```diff
@@ -25,17 +25,16 @@ import (
...
 		t1 := time.Now()
 
-		defer func() {
+		defer func(p string) {
 			t2 := time.Now()
 
-			printEnd(reqID, lw, t2.Sub(t1))
-		}()
+			printEnd(p, reqID, lw, t2.Sub(t1))
+		}(prefix)
```
VojtechVitek: you can use the `prefix` variable directly inside of `defer`

VojtechVitek:
@gdm85 @GeertJohan Logger should be used before Recoverer -- Recoverer responds with `HTTP 500` and Logger should be able to pick this value up. If you reverse the order, Logger won't know the status code.

Logger logs requests and responses. Recoverer prints panic+stack, as it is a critical error. I think of it as Apache's Error Log vs. Access Log.

gdm85:
@VojtechVitek my bad..I had only built `chi` and not `chi/middleware`. I will process your other feedback

middleware/logger.go
gdm85: The `[1:]` is to skip the first `"` added by the `%q` formatter. It's a trick to be able to put the URI and the protocol within the same quotation marks, while we're giving the method a different color.

gdm85:
@VojtechVitek yes I agree with that and in line with that view of Recoverer, I will add request context to the panic message in a separate commit

gdm85:
So, when Logger is not added or added after Recoverer (for example in the simple example):
```
2015/11/11 17:47:25 [::1]:42520 [delver/rUYSwm1LYq-000001] "GET /accounts HTTP/1.1" panic: test
/home/gdm/chi/_gopath/src/github.com/pressly/chi/middleware/recoverer.go:28 (0x4b3ebc)
	Recoverer.func1.1: debug.PrintStack()
/usr/local/go/src/runtime/asm_amd64.s:438 (0x45b76e)
	call64: CALLFN(·call64, 64)
/usr/local/go/src/runtime/panic.go:423 (0x42c179)
[...]
^Cexit status 2
```

When Logger is before Recoverer (best practice):
```
2015/11/11 17:47:43 [::1]:42526 [delver/71wjK689mh-000001] "GET /accounts HTTP/1.1" panic: test
/home/gdm/chi/_gopath/src/github.com/pressly/chi/middleware/recoverer.go:28 (0x4b3ebc)
	Recoverer.func1.1: debug.PrintStack()
/usr/local/go/src/runtime/asm_amd64.s:438 (0x45b76e)
	call64: CALLFN(·call64, 64)
/usr/local/go/src/runtime/panic.go:423 (0x42c179)
[...]
/usr/local/go/src/runtime/asm_amd64.s:1696 (0x45d9f1)
	goexit: BYTE	$0x90	// NOP
2015/11/11 17:47:25 [::1]:42520 [delver/rUYSwm1LYq-000001] "GET /accounts HTTP/1.1" 500 22B in 9.163817ms
```

It could be argued that the panic line is "duplicating" what is later on reported by Logger as 500, but it adds context to the panic message...I think it's okay to have it this way as panics should really be exceptional anyway, and context is exactly what you want (whether you have or not Logger enabled there).

gdm85:
The last commit I added adds a not to clarify how it's best to use Recoverer together with Logger, it can be dropped or squashed depending on what's decided.

pkieltyka:
@gdm85 thanks, I'm going to merge this and make some small changes after it. Great addition!

VojtechVitek:
Nice, this is so much better now. Thanks!

PR #20, opened 2015-11-13
Now saving matched router's path to the context

utrack:
Added ```fullPath``` param to the tree node, which contains full path for
this node. Each time request is processed, full non-interpolated path is put 
to ctx's ```const MatchedPathCtxKey``` key, which can be retrieved by logging
middlewares.

### DIFF
--- a/mux.go
+++ b/mux.go
@@ -62,6 +62,7 @@ type ctxKey int
 const (
 	URLParamsCtxKey ctxKey = iota
 	SubRouterCtxKey
+	MatchedPathCtxKey
 )
 
 func NewMux() *Mux {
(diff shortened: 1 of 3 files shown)


pkieltyka:
Hmm. I don't see why this is useful. The logger works fine. And full path is in r.URL.Path

utrack:
@pkieltyka It's useful for logging endpoints' metrics. Say, request timings and counts per endpoint.
For endpoint ```/users/:id``` r.URL.Path contains
```
/users/1
/users/2
/users/3
```
and so on. If you'd export stats outside (with Prometheus or similar software) all these would be viewed as separate endpoints, while using path from ```MatchedPathCtxKey``` would present a single path per endpoint, so all these requests will be exported as calls to ```/user/:id```.

pkieltyka:
I see, that's a narrow use case I don't plan to support in the core. 

An option you have is to implement the chi.Router interface with your own mux where you can make it send those stats or automatically attach a middleware for the same path, etc. You'd make a struct, embed chi.Mux, and change it how you like.

Thanks for the submission tho.

PR #21, opened 2015-11-26
Implement middleware.Compress

VojtechVitek:
Let's discuss the API before we merge this.

### DIFF
--- a/middleware/compress.go
+++ b/middleware/compress.go
@@ -1,0 +1,186 @@
+package middleware
+
+import (
+	"compress/flate"
+	"compress/gzip"
+	"io"
+	"net/http"
+	"strings"
+)
+
+type encoding int
+
+const (
+	encodingNone encoding = iota
+	encodingGzip
+	encodingDeflate
+)
+
+var defaultContentTypes = map[string]struct{}{
+	"text/html":                struct{}{},
+	"text/css":                 struct{}{},
+	"text/plain":               struct{}{},
+	"text/javascript":          struct{}{},
+	"application/javascript":   struct{}{},
+	"application/x-javascript": struct{}{},
+	"application/json":         struct{}{},
+	"application/atom+xml":     struct{}{},
+	"application/rss+xml ":     struct{}{},
+}
+
+// DefaultCompress is a middleware that compresses response
+// body of predefined content types to a data format based
+// on Accept-Encoding request header. It uses a default
+// compression level.
+func DefaultCompress(next http.Handler) http.Handler {
+	return Compress(flate.DefaultCompression)(next)
+}
+
+// CompressLevel is a middleware that compresses response
+// body of a given content types to a data format based
+// on Accept-Encoding request header. It uses a given
+// compression level.
+func Compress(level int, types ...string) func(next http.Handler) http.Handler {
+	contentTypes := defaultContentTypes
+	if len(types) > 0 {
+		contentTypes = make(map[string]struct{}, len(types))
+		for _, t := range types {
+			contentTypes[t] = struct{}{}
+		}
+	}
+
+	return func(next http.Handler) http.Handler {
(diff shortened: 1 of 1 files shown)


middleware/compress.go
```diff
@@ -0,0 +1,186 @@
...
+// compression level.
+func DefaultCompress(next http.Handler) http.Handler {
+	return Compress(flate.DefaultCompression)(next)
+}
+
+// CompressLevel is a middleware that compresses response
+// body of a given content types to a data format based
+// on Accept-Encoding request header. It uses a given
+// compression level.
+func Compress(level int, types ...string) func(next http.Handler) http.Handler {
```
VojtechVitek: I'm thinking of `func Compress(levelFunc func(r *http.Request) int, types ...string) func(next http.Handler) http.Handler {` instead, so we could change the compression level based for example on User-Agent. Mobile usually need better compression.

middleware/compress.go
```diff
@@ -0,0 +1,186 @@
...
+	"text/html":                struct{}{},
+	"text/css":                 struct{}{},
+	"text/plain":               struct{}{},
+	"text/javascript":          struct{}{},
+	"application/javascript":   struct{}{},
+	"application/x-javascript": struct{}{},
+	"application/json":         struct{}{},
+	"application/atom+xml":     struct{}{},
+	"application/rss+xml ":     struct{}{},
+}
```
VojtechVitek: Since there are not so many Content-Types, I'm thinking that slice `[]string` with a predefined order would be better. Haven't decided yet.

middleware/compress.go
pkieltyka: I think that could be interesting. Maybe try that in another commit later with some examples

middleware/compress.go
```diff
@@ -0,0 +1,186 @@
...
+		return http.HandlerFunc(fn)
+	}
+}
+
+func selectEncoding(h http.Header) encoding {
+	enc := h.Get("Accept-Encoding")
+
+	switch {
+	// TODO:
+	// case "br":    // Brotli, experimental. Firefox 2016, to-be-in Chromium.
```
pkieltyka: do we need these?

middleware/compress.go
```diff
@@ -0,0 +1,186 @@
...
+		// case "zopfli":   // Too slow on-the-fly.
+		// case "xz":       // Too slow on-the-fly.
+	}
+
+	return encodingNone
+}
+
+type maybeCompressResponseWriter struct {
+	http.ResponseWriter
+	w            io.Writer
```
pkieltyka: Do we need this since we embed http.ResponseWriter ...?

middleware/compress.go
VojtechVitek: Maybe in the future. But since these encodings are supposed to be faster/better and will have browser support, I wanted to leave the boilerplate in the code.

middleware/compress.go
VojtechVitek: I believe we do. I'm not able to "change" the underlying Writer otherwise.

middleware/compress.go
```diff
@@ -0,0 +1,186 @@
+package middleware
+
+import (
+	"compress/flate"
+	"compress/gzip"
```
VojtechVitek: I'm thinking of exposing the writers via API, so users could inject 
https://github.com/klauspost/compress or https://github.com/klauspost/pgzip

pkieltyka: I think good idea. You could take a Compressor or something, or use a default one if its not passed

middleware/compress.go
henrikbjorn: the structs are empty, why not have a `map[string]bool` instead?

VojtechVitek: Empty structs are more memory-efficient than bools.

PR #22, opened 2015-12-07
Fixes panic on HTTP method not in methodMap

c2h5oh:
This is just a panic fix, but IMO we should allow custom HTTP methods (HTTP is extensible) instead
of relying on a static map. methodMap should be only used for the standard methods defined in HTTP RFCs - for the `Mux.Get` short methods.

### DIFF
--- a/mux.go
+++ b/mux.go
@@ -3,6 +3,7 @@ package chi
 import (
 	"fmt"
 	"net/http"
+	"strings"
 
 	"golang.org/x/net/context"
 )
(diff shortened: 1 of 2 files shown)


mux.go
```diff
@@ -224,8 +225,25 @@ func (tr treeRouter) ServeHTTPC(ctx context.Context, w http.ResponseWriter, r *h
...
+	// Check if method is supported by chi
+	method, ok := methodMap[r.Method]
+	if !ok {
+		w.WriteHeader(405)
+
+		// Return allowed methods, as required by RFC2616 for 405 Method not allowed
+		methods := make([]string, len(methodMap))
+		i := 0
+		for m := range methodMap {
+			methods[i] = m // stil faster than append to array with capacity
```
VojtechVitek: typo in comment

mux.go
```diff
@@ -224,8 +225,25 @@ func (tr treeRouter) ServeHTTPC(ctx context.Context, w http.ResponseWriter, r *h
...
+			i++
+		}
+		w.Header().Add("Allow", strings.Join(methods, ","))
```
VojtechVitek: You need to add headers before `w.WriteHeader(405)`

mux_test.go
```diff
@@ -200,6 +200,11 @@ func TestMux(t *testing.T) {
 	if string(body) != "catchall" {
 		t.Error("expecting response body: 'catchall'")
 	}
+
+	// Custom http method DIE /admin/catch-this
+	if resp := testRequest(t, ts, "DIE", "/ping/1/woop", nil); resp != "Method Not Allowed" {
```
VojtechVitek: check `resp.Code` instead

c2h5oh: yeah, not provided by testRequest helper. I'll open a new ticket for that one.

VojtechVitek:
LGTM

mux.go
```diff
@@ -224,8 +225,24 @@ func (tr treeRouter) ServeHTTPC(ctx context.Context, w http.ResponseWriter, r *h
 		routePath = r.URL.Path
 	}
 
+	// Check if method is supported by chi
+	method, ok := methodMap[r.Method]
+	if !ok {
+		// Return allowed methods, as required by RFC2616 for 405 Method not allowed
+		methods := make([]string, len(methodMap))
```
pkieltyka: thanks for fixing this panic.

I think we should move this procedure to its own function, like "writeMethodNotAllowed(w); return" .

PR #26, opened 2015-12-09
Adding backlog to throttler middleware

xiam:


middleware/throttle.go
```diff
@@ -34,21 +54,46 @@ type token struct{}
...
+		case tok := <-t.tokens:
+			defer func() {
+				t.tokens <- tok
+			}()
+			t.h.ServeHTTPC(ctx, w, r)
+		}
 	}
 }
+
+func httpStatus(w http.ResponseWriter, statusCode int) {
```
VojtechVitek: use http.Error() instead.

middleware/throttle.go
```diff
@@ -2,23 +2,43 @@ package middleware
...
-// processed requests at a time.
-func Throttle(limit int) func(chi.Handler) chi.Handler {
-	if limit <= 0 {
+var (
+	defaultTimeout = time.Second * 60
+)
+
+// Throttle is a middleware that limits number of currently processed requests
+// at a time.
+func Throttle(limit int, backloglimit int, timeout time.Duration) func(chi.Handler) chi.Handler {
```
VojtechVitek: So far, it looks good! But before we merge this PR, I'd like to leave the original Throttle API backward-compatible. Backlog+timeout should be optional.

```go
middleware.Throttle(10) // hard limit on number of processed requests
middleware.Throttle(10, middleware.Backlog(1000, time.Second)) // use backlog+timeout too
```

@xiam @pkieltyka thoughts?

middleware/throttle.go
```diff
@@ -34,21 +54,46 @@ type token struct{}
...
-	case tok := <-t.tokens:
+	case btok := <-t.backlogtokens:
 		defer func() {
-			t.tokens <- tok
+			t.backlogtokens <- btok
 		}()
-		t.h.ServeHTTPC(ctx, w, r)
+		select {
+		case <-timer.C:
+			httpStatus(w, http.StatusGatewayTimeout)
```
c2h5oh: This should still be a 503 - there is no gateway. 504 should only be returned by proxy if proxied service times out.

middleware/throttle.go
```diff
@@ -34,21 +54,46 @@ type token struct{}
 
 // throttler limits number of currently processed requests at a time.
 type throttle struct {
```
VojtechVitek: @pkieltyka what was wrong with `throttler` variable name?

middleware/throttle.go
```diff
@@ -34,21 +54,46 @@ type token struct{}
...
 		defer func() {
-			t.tokens <- tok
+			t.backlogtokens <- btok
 		}()
-		t.h.ServeHTTPC(ctx, w, r)
+		select {
+		case <-timer.C:
+			httpStatus(w, http.StatusGatewayTimeout)
+			return
+		case <-ctx.Done():
```
VojtechVitek:
```go
if ctx.Err() == context.DeadlineExceeded {
    httpStatus(w, http.StatusGatewayTimeout)
    return
}
```

middleware/throttle.go
```diff
@@ -2,23 +2,43 @@ package middleware
...
+// Throttle is a middleware that limits number of currently processed requests
+// at a time.
+func Throttle(limit int, backloglimit int, timeout time.Duration) func(chi.Handler) chi.Handler {
+	if limit < 1 {
 		panic("middleware.Throttle expects limit > 0")
 	}
 
+	if backloglimit < limit {
+		panic("middleware.Throttle expects backloglimit > limit")
+	}
```
c2h5oh: This needs to be explained. Seeing backlog in a throttler I assumed that backlog is the number of requests that you can queue on top of the limit, but if I correctly understand the code starting here https://github.com/pressly/chi/pull/26/files#diff-31daf09af7ba24641f7fce5bb0918d1dR76 backlog is the part of the limit that is queued (which explains why `backlog < limit` has to be enforced)

middleware/throttle.go
VojtechVitek: There are several places such like this. I think that all timeouts should trigger http.StatusGatewayTimeout instead of http.StatusServiceUnavailable -- objections?

middleware/throttle.go
xiam: Good point, I think it would be better if I change this part and put the backlog number on top of the limit.

middleware/throttle.go
xiam: If error messages are different clients will be able to debug what's happening by themselves, or a least the will provide more accurate debugging information to a developer. If all errors are the same developers will not know where is their code failing. We can also use custom statuses if the standard ones fail to represent the error accurately.

middleware/throttle.go
```diff
@@ -34,21 +54,46 @@ type token struct{}
...
 }
 
 // TODO: add support for a backlog
 
 // ServeHTTPC implements chi.Handler interface.
 func (t *throttle) ServeHTTPC(ctx context.Context, w http.ResponseWriter, r *http.Request) {
+	timer := time.NewTimer(t.timeout)
+
 	select {
+	case <-timer.C:
```
VojtechVitek: Thinking of the behavior...

Imho, if the backlog is full, we should return Service Unavailable response as soon as possible. Only if the request can fit into the backlog, then we should start the timeout -- while waiting for the throttle token.

So, I'd remove timer from this select and move the initialization down:

```diff
- case btok := <-t.backlogtokens:
+ case btok := <-t.backlogtokens:
+    timer := time.NewTimer(t.timeout)
      // ... case for time.C, ctx.Done() and t.tokens
+ default:
+    httpStatus(w, http.StatusServiceUnavailable)
+    return
```

middleware/throttle.go
```diff
@@ -2,23 +2,43 @@ package middleware
...
 	"github.com/pressly/chi"
 	"golang.org/x/net/context"
 )
 
-// Throttle is a middleware that limits number of currently
-// processed requests at a time.
-func Throttle(limit int) func(chi.Handler) chi.Handler {
-	if limit <= 0 {
+var (
+	defaultTimeout = time.Second * 60
```
pkieltyka: is this the backlog timeout? I think then it should be named `defaultBacklogTimeout`

middleware/throttle.go
VojtechVitek: Or we could simply have two middlewares:
```go
middleware.Throttle(N)
middleware.ThrottleBacklog(N, M, T)
```

middleware/throttle.go
VojtechVitek: I don't even think we need a default value.

middleware/throttle.go
pkieltyka: nothing, I am okay with throttler too, but rename the file too if you guys think its better. I consider that "throttle" is an action on a request.. and a "throttler" is a thing that throttles requests, so, depends on context and what sounds better.. I dunno :P both work

middleware/throttle.go
```diff
@@ -34,21 +54,45 @@ type token struct{}
...
 type throttle struct {
-	h      chi.Handler
-	tokens chan token
+	h             chi.Handler
+	tokens        chan token
+	backlogtokens chan token
+	timeout       time.Duration
 }
 
 // TODO: add support for a backlog
```
pkieltyka: lets remove this..

middleware/throttle.go
pkieltyka: I agree. If backlog is full, then 503.

middleware/throttle.go
VojtechVitek: Since it's a convention to use "-er" type names in Go (writer, reader, encoder, scanner, replacer etc.), I'd prefer throttler.

pkieltyka: that convention is for interfaces.. not for middlewares. I don't buy that argument :)

VojtechVitek: for both interfaces and structs ... https://golang.org/src/strings/replace.go?s=305:341#L1,  https://golang.org/src/net/http/server.go?s=1517:2599#L2094, https://golang.org/src/bufio/scan.go?s=1184:1728#L20 etc.

middleware/throttler.go
```diff
@@ -0,0 +1,99 @@
...
+	case btok := <-t.backlogtokens:
+		timer := time.NewTimer(t.timeout)
+
+		defer func() {
+			t.backlogtokens <- btok
+		}()
+
+		select {
+		case <-timer.C:
+			httpStatus(w, http.StatusGatewayTimeout)
```
xiam: Probably `http.StatusGatewayTimeout` is not the best choice if we try to be strict, but I rather have an error message that helps us spot a problem easily. In this case the `http.StatusGatewayTimeout` is associated to operations that take too much time to complete, while `http.StatusServiceUnavailable` is only associated to capacity problems.

c2h5oh: In that case return the correct code, but modify the message.

middleware/throttler.go
```diff
@@ -0,0 +1,105 @@
...
+)
+
+const (
+	errCapacityExceeded = "Server capacity exceeded."
+	errTimedOut         = "Timed out while waiting for a pending request to complete."
+	errContextCanceled  = "Context was canceled."
+)
+
+var (
+	defaultThrottleTimeout = time.Second * 60
```
pkieltyka: can we call this `defaultBacklogTImeout` ..? this is what it is

middleware/throttler.go
```diff
@@ -0,0 +1,105 @@
...
+
+// throttler limits number of currently processed requests at a time.
+type throttler struct {
+	h             chi.Handler
+	tokens        chan token
+	backlogtokens chan token
+	timeout       time.Duration
+}
+
+// TODO: add support for a backlog
```
pkieltyka: please remove this comment

middleware/throttler.go
```diff
@@ -0,0 +1,105 @@
...
+
+// ServeHTTPC implements chi.Handler interface.
+func (t *throttler) ServeHTTPC(ctx context.Context, w http.ResponseWriter, r *http.Request) {
+	select {
+	case <-ctx.Done():
+		http.Error(w, errContextCanceled, http.StatusServiceUnavailable)
+		return
+	case btok := <-t.backlogtokens:
```
pkieltyka: I think we should call it backlogTokens

middleware/throttler.go
```diff
@@ -0,0 +1,105 @@
...
+// Throttle is a middleware that limits number of currently processed requests
+// at a time.
+func Throttle(limit int) func(chi.Handler) chi.Handler {
+	return ThrottleBacklog(limit, 0, defaultThrottleTimeout)
+}
+
+// ThrottleBacklog is a middleware that limits number of currently processed
+// requests at a time and provides a backlog for holding a finite number of
+// pending requests.
+func ThrottleBacklog(limit int, backloglimit int, timeout time.Duration) func(chi.Handler) chi.Handler {
```
pkieltyka: I kind of prefer `backlogLimit` over `backloglimit`

Issue #29, opened 2016-01-06
Custom not found handler


pkieltyka:
Related to #9

pkieltyka:
Solved in b9ba24fa6c5ac5b949cc7beaea5f80e4c5c5d3a0

PR #30, opened 2016-01-06
NotFound() method on Router

pkieltyka:
Solves #9 and #29

### DIFF
--- a/mux_test.go
+++ b/mux_test.go
@@ -212,6 +212,10 @@ func TestMuxPlain(t *testing.T) {
 	r.Get("/hi", func(ctx context.Context, w http.ResponseWriter, r *http.Request) {
 		w.Write([]byte("bye"))
 	})
+	r.NotFound(func(ctx context.Context, w http.ResponseWriter, r *http.Request) {
+		w.WriteHeader(404)
+		w.Write([]byte("nothing here"))
+	})
 
 	ts := httptest.NewServer(r)
 	defer ts.Close()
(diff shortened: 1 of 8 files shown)


pkieltyka:
This implementation is alright, but it would be nice if the NotFound() handler could cascade upwards back up the chain.

ie.
```go
sub := chi.NewRouter()
sub.Get("/hello", ...)

root := chi.NewRouter()
root.Get("/", ...)
root.NotFound(func(ctx context.Context, w http.ResponseWriter, r *http.Request) {
  w.WriteHeader(404)
  w.Write([]byte("custom 404 page"))
})
root.Mount("/subrouter", sub)
```

a 404 on root render the custom error, but in the subrouter it will be the standard http.NotFound(), unless you define it again on that router.

This should be solvable. WIP.

pkieltyka:
technically, this PR already supports #3 because if you set http.NotFound to some handler, chi will use that by default

pkieltyka:
This PR completes the NotFound() handler support, along with cascading up the tree looking for a not found handler.

mux_test.go
```diff
@@ -212,6 +212,10 @@ func TestMuxPlain(t *testing.T) {
 	r.Get("/hi", func(ctx context.Context, w http.ResponseWriter, r *http.Request) {
 		w.Write([]byte("bye"))
 	})
+	r.NotFound(func(ctx context.Context, w http.ResponseWriter, r *http.Request) {
+		w.WriteHeader(404)
+		w.Write([]byte("nothing here"))
```
c2h5oh: Defaulting to https://golang.org/pkg/net/http/#NotFound would be better

PR #37, opened 2016-04-20
Mux ServeHTTPC supports context of generic implementation

mrcpvn:
ServeHTTPC  method will embed a chi.Context inside an existing different implementation of context.Context

### DIFF
--- a/mux.go
+++ b/mux.go
@@ -273,16 +273,19 @@ func (mx *Mux) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 // ServeHTTPC is chi's Handler method that adds a context.Context argument to the
 // standard ServeHTTP handler function.
 func (mx *Mux) ServeHTTPC(ctx context.Context, w http.ResponseWriter, r *http.Request) {
-	rctx, _ := ctx.(*Context)
-	if rctx == nil {
-		rctx = mx.pool.Get().(*Context)
-		rctx.Context = ctx
-		defer func() {
-			rctx.reset()
-			mx.pool.Put(rctx)
-		}()
+	rctx, ok := ctx.(*Context)
+	if !ok || rctx == nil {
+		rctx, ok = ctx.Value(routeCtxKey).(*Context)
+		if !ok {
+			rctx = mx.pool.Get().(*Context)
+			defer func() {
+				rctx.reset()
+				mx.pool.Put(rctx)
+			}()
+			ctx = context.WithValue(ctx, routeCtxKey, rctx)
+		}
 	}
-	mx.handler.ServeHTTPC(rctx, w, r)
+	mx.handler.ServeHTTPC(ctx, w, r)
 }
 
 // A treeRouter manages a radix trie prefix-router for each HTTP method and passes
--- a/mux.go
+++ b/mux.go
@@ -273,7 +273,16 @@ func (mx *Mux) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 // ServeHTTPC is chi's Handler method that adds a context.Context argument to the
 // standard ServeHTTP handler function.
 func (mx *Mux) ServeHTTPC(ctx context.Context, w http.ResponseWriter, r *http.Request) {
-	mx.handler.ServeHTTPC(ctx, w, r)
+	rctx, _ := ctx.(*Context)
+	if rctx == nil {
+		rctx = mx.pool.Get().(*Context)
+		rctx.Context = ctx
+		defer func() {
+			rctx.reset()
+			mx.pool.Put(rctx)
+		}()
+	}
+	mx.handler.ServeHTTPC(rctx, w, r)
 }
 
 // A treeRouter manages a radix trie prefix-router for each HTTP method and passes
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 .idea
 *.sw?
+.vscode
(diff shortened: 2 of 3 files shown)


pkieltyka:
@mrcpvn thanks for the PR! just wondering, did you run into this because of a bug..?

mrcpvn:
@pkieltyka yes, if the input of the (*Mux).ServeHTTPC function was not a chi.Context there was a panic on the chi.(*treeRouter).ServeHTTPC

panic: interface conversion: interface {} is nil, not *chi.Context [recovered]
        panic: interface conversion: interface {} is nil, not *chi.Context

goroutine 41 [running]:
panic(0x3adc40, 0xc8201228c0)
        /usr/local/go/src/runtime/panic.go:464 +0x3e6
testing.tRunner.func1(0xc820139320)
        /usr/local/go/src/testing/testing.go:467 +0x192
panic(0x3adc40, 0xc8201228c0)
        /usr/local/go/src/runtime/panic.go:426 +0x4e9
github.com/pressly/chi.treeRouter.ServeHTTPC(0xc820103f80, 0xc820104150, 0x14c00e0, 0xc820103fb0, 0x14c0288, 0xc8200feea0, 0xc82012b5e0)
        /Users/marco/Programmazione/go/src/github.com/pressly/chi/mux.go:329 +0x301
github.com/pressly/chi.(*treeRouter).ServeHTTPC(0xc820120ca0, 0x14c00e0, 0xc820103fb0, 0x14c0288, 0xc8200feea0, 0xc82012b5e0)
        <autogenerated>:18 +0xca
github.com/pressly/chi.(*Mux).ServeHTTPC(0xc820127080, 0x14c00e0, 0xc820103fb0, 0x14c0288, 0xc8200feea0, 0xc82012b5e0)
        /Users/marco/Programmazione/go/src/github.com/pressly/chi/mux.go:288 +0x6c
github.com/pressly/chi.TestMuxServeHTTPC(0xc820139320)
        /Users/marco/Programmazione/go/src/github.com/pressly/chi/mux_test.go:325 +0x992
testing.tRunner(0xc820139320, 0x6112c8)
        /usr/local/go/src/testing/testing.go:473 +0x98
created by testing.RunTests
        /usr/local/go/src/testing/testing.go:582 +0x892
exit status 2
FAIL    github.com/pressly/chi  0.052s

mux.go
```diff
@@ -273,6 +273,18 @@ func (mx *Mux) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 // ServeHTTPC is chi's Handler method that adds a context.Context argument to the
 // standard ServeHTTP handler function.
 func (mx *Mux) ServeHTTPC(ctx context.Context, w http.ResponseWriter, r *http.Request) {
+	rctx, ok := ctx.(*Context)
+	if !ok || rctx == nil {
```
VojtechVitek: Personally, I'd `panic()` on nil.

pkieltyka:
this PR is interesting, but I'll probably go with a different approach to solving it.

pkieltyka:
Thanks for the PR @mrcpvn .. I've closed it with 79fa9188c5d924a6408a8b8ef4f267acec735a84

mrcpvn:
Hello @pkieltyka. Can you please explain why my implementation was not a good solution for this issue?
I think that an incoming context should be handled by the router.

pkieltyka:
hey @mrcpvn. Definitely, I agree the incoming context should be handled by the router and it is, but by ServeHTTP instead of ServeHTTPC like you suggested in the PR. The typical connection point of a chi.Mux is as an http.Handler, passed to http.ListenAndServe(). 

When would you call mux.ServeHTTPC()? I suppose one example could be if you're writing tests and testing specific handlers via httptest.NewRecorder(). I've been writing my http tests with httptest.NewServer(mux). If the reason is for tests, one way I could do it is export the `newContext()` method in chi so someone could make a new routing context for purposes of tests. Let me know if you have other reasons to call ServeHTTPC directly on a mux object. In order to support both methods, it just makes the code a bit messy and I'd rather avoid it by convention. If there are specific needs for it, please let me know and there might be a different way to solve it.

Second and most importantly, chi v2.0 will be ready for Go 1.7, which no longer needs the ServeHTTPC() method because http.Request now has a Context field of its own :) yeay... so again, this problem goes away. See https://github.com/pressly/chi/blob/v2 -- this version is going to be a major step forward, still very minimal in philosophy but static checking everywhere in the router, and I'm going to add a few more perks now that I can lean on context.Context even further. At that point chi v1 will go into a stable branch where < 1.7 users can still use it until they upgrade. But still, v1 should be stable an very refined.

Third, in relation to your PR specifically, the defer statement does have a cost, and I'd rather not use it per request.

mrcpvn:
Thank you for the explanation @pkieltyka. I found the issue with mux.ServeHTTPC because I was calling the method from a middleware handler that is context aware and will pass a context.Context to the chi.Router (calling directly the ServeHTTPC method). This is probably an edge case but should not be ignored. 
I know that with Go 1.7 this problem will go away but for the moment I need to support a context in the ServeHTTPC method.

pkieltyka:
Can you show example of middleware / router? I'll get a proper fix for this to make sure v1 is stable

mrcpvn:
I pushed the example in this repository:
https://github.com/mrcpvn/chi_middleware

pkieltyka:
@mrcpvn @VojtechVitek let me know what you guys think of https://github.com/pressly/chi/pull/40

All tests pass and benchmarks remain the same.

mrcpvn:
@pkieltyka @VojtechVitek the implementation looks good to me.
I have also done some performance tests and found no issues.

PR #38, opened 2016-05-03
Request/Response headers debugging middleware

VojtechVitek:
Helpful middleware for debugging Request and Response Headers.

I found myself re-implementing similar feature multiple times already. Note that `httputil.DumpRequest()` (and similar funcs) are not easily pluggable to complex routes and their output is not that readable.

We can discuss if we want this in middleware pkg, or in debug pkg or something. Or if we want it at all?
If we do, I can add `LogRequestBody` and `LogResponseBody` too.

### DIFF
--- a/middleware/debug.go
+++ b/middleware/debug.go
@@ -1,0 +1,32 @@
+package middleware
+
+import (
+	"encoding/json"
+	"log"
+	"net/http"
+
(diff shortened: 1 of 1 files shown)


pkieltyka:
I think it would be better as a goware/debugmw package

PR #46, opened 2016-06-09
logger will always print time in microseconds

mrcpvn:
I added the elapsed time in µs, it will be easier to parse the log automatically

### DIFF
--- a/middleware/logger.go
+++ b/middleware/logger.go
@@ -85,14 +85,20 @@ func printRequest(buf *bytes.Buffer, reqID string, w writerProxy, dt time.Durati
 
 	buf.WriteString(" in ")
 	if dt < 500*time.Millisecond {
-		cW(buf, nGreen, "%s", dt)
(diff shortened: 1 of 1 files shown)


pkieltyka:
its an interesting point to normalize all request times to a single unit, but, visually its nicer to read the unit of time as us, ms, s. @mrcpvn is this for instrumentation..? perhaps you should check the elapsed time and have a middleware that emits the data? .. or, perhaps we should have a formatter for the Logger to make this configurable

mrcpvn:
The logger will print the time in us, ms, s but it will always report a us measure:

2016/06/14 16:44:58 [XXXXXXXX/i4yumvPyxk-000001] "GET http://127.0.0.1:9000/test HTTP/1.1" from 127.0.0.1:53425 - 400 32B in 59.893µs (59.893µs)

A formatter for the logger would be a handy solution.

PR #47, opened 2016-06-13
Render: Presenter

VojtechVitek:
Present your data models and/or create API object versioning!

```go
func handler(ctx context.Context, w http.ResponseWriter, r *http.Request) {
	obj := &Object{
		ID:       1,
		AuthData: "secret data for auth'd users only",
		Data:     []string{"one", "two", "three", "four"},
	}

	render.Respond(ctx, w, obj)
}
```

```go
v2 := render.NewPresenter()
v2.Register(func(ctx context.Context, from *Object) (*v2.Object, error) {
	to := &v2.Object{
		Object:	from,
		URL:	fmt.Sprintf("http://api.example.com/objects/%d", from.ID),
		Data:	map[string]bool{},
	}
	for _, item := range from.Data {
		to.Data[item] = true
	}
	return to, nil
})
```

```go
r.Get("/", handler) // renders raw Object
r.Get("/v2", render.UsePresenter(v2), handler) // renders Object in version 2
```


render/present.go
```diff
@@ -0,0 +1,100 @@
...
+		panic(err)
+	}
+	for _, fn := range conversionFuncs {
+		if err := p.register(fn); err != nil {
+			panic(err)
+		}
+	}
+}
+
+func (p *presenter) RegisterFrom(presenter *presenter, presenters ...*presenter) {
```
VojtechVitek: This copies Presenter conversion functions from other presenters.. Should we call it .Copy()?

middleware/value.go
```diff
@@ -0,0 +1,19 @@
...
+
+import (
+	"net/http"
+
+	"github.com/pressly/chi"
+	"golang.org/x/net/context"
+)
+
+// WithValue is a middleware that sets a given key/value in a context chain.
+func WithValue(key interface{}, val interface{}) func(next chi.Handler) chi.Handler {
```
pkieltyka: this is pretty cool, I like the short-hand

_examples/render/main.go
```diff
@@ -0,0 +1,153 @@
...
+	"time"
+
+	"golang.org/x/net/context"
+
+	"github.com/pressly/chi"
+	"github.com/pressly/chi/middleware"
+	"github.com/pressly/chi/render"
+)
+
+type RuntimeObject struct {
```
pkieltyka: in the example, I think it would be nice to use a "real-world" analogy. this way people reading the example can relate to it some kind of domain .. for example, maybe use the idea of Articles and a simple blog presenting other data

_examples/render/main.go
```diff
@@ -0,0 +1,153 @@
...
+	render.DefaultRespond(ctx, w, v)
+}
+
+func main() {
+	r := chi.NewRouter()
+
+	r.Use(middleware.RequestID)
+	r.Use(middleware.Logger)
+	r.Use(middleware.Recoverer)
+	r.Use(render.ParseContentType)
```
pkieltyka: I've never been a fan of the ParseContentType middleware, something we should rename or consider differently.

render/present.go
```diff
@@ -0,0 +1,100 @@
+package render
```
pkieltyka: I think rename this file "presenter.go" since it better reflects the naming in the code

_examples/render/main.go
VojtechVitek: We can remove this mw for good if we start passing `*http.Request` to render.Respond:

```go
render.Respond(w, r, Object)
```
--- render would have access to `r.Context()`, `r.URL` etc.

_examples/render/main.go
VojtechVitek: good point

_examples/render/main.go
pkieltyka: yea, that is true, and perhaps we should.. what do you think? it's a bit verbose, but it is quite helpful.

pkieltyka: we may want to support render.Status(200) or something, so we can override the status code for a response. the http.ResponseWriter has w.WriteHeader(statusInt), but, this will write the header right then, which isn't what we want.

the consideration is, what should be the signature of render.Respond(), I do believe we want to give control

perhaps we have.. render.Respond(w, r, Object) and render.RespondStatus(w, r, status, Object)

render/render.go
```diff
@@ -1,81 +1,9 @@
...
-		v = reflect.MakeSlice(val.Type(), 0, 0).Interface()
-	}
-
-	JSON(w, status, v)
-}
+const (
+	statusCtxKey ctxKey = iota
+	presenterCtxKey
+	contentTypeCtxKey
+)
```
c2h5oh: This should not be `iota` - `string` corresponding to the actual HTTP header would be better. This is too easy to be overwritten by an uninitialized int.

render/respond.go
```diff
@@ -0,0 +1,111 @@
...
+
+	// Format data based on Content-Type.
+	switch contentType, _ := ctx.Value(contentTypeCtxKey).(ContentType); contentType {
+	case ContentTypeJSON:
+		JSON(ctx, w, v)
+	case ContentTypeXML:
+		XML(ctx, w, v)
+	default:
+		JSON(ctx, w, v)
+	}
```
c2h5oh: `HTML`, `TXT`

render/respond.go
```diff
@@ -0,0 +1,111 @@
...
+	w.WriteHeader(status)
+	w.Write([]byte(v))
+}
+
+func JSON(ctx context.Context, w http.ResponseWriter, v interface{}) {
+	// Force to return empty JSON array [] instead of null in case of zero slice.
+	val := reflect.ValueOf(v)
+	if val.Kind() == reflect.Slice && val.IsNil() {
+		v = reflect.MakeSlice(val.Type(), 0, 0).Interface()
+	}
```
c2h5oh: IMO this is not the right default behavior.

pkieltyka: do you think we should just leave it as is..? whats your thoughts on why/why not?

render/render.go
pkieltyka: @c2h5oh actually not. Each constant would be stored as a different value in the context because the type of ctxKey is different than something like an int. The only thing that could overwrite it would be another ctxKey type of the same value

pkieltyka: however, I think I'm going to move to a slightly different pattern so its easier to debug.. like: https://github.com/pressly/chi/blob/v2/context.go#L12-L22  .. this is what net/http in 1.7 does for its request context keys, the advantage is debugging the values you can see strings, and again, not to worry about being overridden by other string's with the same value cuz the type is different

render/respond.go
pkieltyka: definitely we need those as well. perhaps look to https://github.com/unrolled/render for some ideas

render/render.go
VojtechVitek: @pkieltyka Good pattern, I like it. Putting Header name into the struct, as Maciej suggested would make it easier to debug.

@c2h5oh when you define private type and put its instance into a context, noone else can override it from outside. Interesting feature of context.

render/respond.go
```diff
@@ -0,0 +1,111 @@
...
+	if err != nil {
+		http.Error(w, err.Error(), http.StatusInternalServerError)
+		return
+	}
+
+	if len(b) > 0 {
+		b = bytes.Replace(b, []byte("\\u003c"), []byte("<"), -1)
+		b = bytes.Replace(b, []byte("\\u003e"), []byte(">"), -1)
+		b = bytes.Replace(b, []byte("\\u0026"), []byte("&"), -1)
+	}
```
c2h5oh: This encoding has been added for security reasons and should not be stripped by default.

And even if we do decide we want to strip it by default it definitely should not do it by allocating 3 copies of a possibly already large response. Do it with pass-through writer.

render/render.go
pkieltyka: @VojtechVitek it's not a feature of context, it's a feature of Go's type system and map[interface{}]interface{}

render/respond.go
```diff
@@ -0,0 +1,111 @@
...
+}
+
+func JSON(ctx context.Context, w http.ResponseWriter, v interface{}) {
+	// Force to return empty JSON array [] instead of null in case of zero slice.
+	val := reflect.ValueOf(v)
+	if val.Kind() == reflect.Slice && val.IsNil() {
+		v = reflect.MakeSlice(val.Type(), 0, 0).Interface()
+	}
+
+	b, err := json.Marshal(v)
```
c2h5oh: https://golang.org/pkg/encoding/json/#NewEncoder and stream to `w` (possibly through the encoding stripping writer - see next comment)

render/respond.go
VojtechVitek: It's tricky -- for example if someone passes `map[string]string{...}` or another complex object into the responder, how do you want to render it into HTML or TXT? It's not really possible, unless you accept only `string` type, which I did:

```
render.HTML(string)
render.String(string) // actually I'd like render.PlainText(string) better
```

HTML and PlainText is not really something that could accept `interface{}`.

render/respond.go
```diff
@@ -0,0 +1,111 @@
...
+		// No header found. Print it out first.
+		w.Write([]byte(xml.Header))
+	}
+
+	w.Write(b)
+}
+
+func Status(ctx context.Context, status int) context.Context {
+	return context.WithValue(ctx, statusCtxKey, status)
+}
```
c2h5oh: `StatusCode` `StatusMessage`?

render/respond.go
pkieltyka: yea, the interface for those methods, HTML and PlainText/String would have to be specific to the kind of data they want to receive

render/respond.go
```diff
@@ -0,0 +1,111 @@
...
+	case ContentTypeJSON:
+		JSON(ctx, w, v)
+	case ContentTypeXML:
+		XML(ctx, w, v)
+	default:
+		JSON(ctx, w, v)
+	}
+}
+
+func String(ctx context.Context, w http.ResponseWriter, v string) {
```
VojtechVitek: @pkieltyka @c2h5oh would you like `render.PlainText()` instead of `render.String()`?

render/respond.go
pkieltyka: hmm.. why message? .. we can use stdlib messages

render/present.go
```diff
@@ -0,0 +1,100 @@
...
+	var contextZeroValue context.Context
+	if !fnType.In(0).Implements(reflect.TypeOf(&contextZeroValue).Elem()) {
+		return fmt.Errorf("expected context.Context as first argument, got: %v", fnType)
+	}
+	var errorZeroValue error
+	if !fnType.Out(1).Implements(reflect.TypeOf(&errorZeroValue).Elem()) {
+		return fmt.Errorf("expected error as second return value, got: %v", fnType)
+	}
+
+	p.ConversionFnStore[fnType.In(1)] = reflect.ValueOf(conversionFunc)
```
c2h5oh: Should check for cycles and limit max depth of a chain

render/respond.go
pkieltyka: I'm good either way, its def more clear about the ContentType

render/present.go
```diff
@@ -0,0 +1,100 @@
...
+		if !ok {
+			return obj
+		}
+		resp := fn.Call([]reflect.Value{reflect.ValueOf(ctx), reflect.ValueOf(obj)})
+		if !resp[1].IsNil() {
+			return resp[1].Interface()
+		}
+		obj = resp[0].Interface()
+	}
+	panic(fmt.Sprintf("render: Present(%T): too many converts", from))
```
c2h5oh: Should happen on register - see comment below.

render/respond.go
VojtechVitek: I think I can remove this completely -- there's a new function in go1.7 to replace these characters by default :)

render/respond.go
VojtechVitek: I agree

render/respond.go
pkieltyka: sounds good. lets try to remove it and see how it goes

render/respond.go
VojtechVitek: We can have certain defaults that we feel are right for REST responder.

I think instead of responding with `nil` for array type, `[]` makes perfect sense and simplifies life of API consumers. But never mind -- if you guys overvote me, we can always put this outside of render pkg into custom `Render()` function that we're gonna use anyway. What do you think?

render/respond.go
```diff
@@ -0,0 +1,111 @@
...
+
+	// Try to find <?xml header in first 100 bytes (just in case there're some XML comments).
+	findHeaderUntil := len(b)
+	if findHeaderUntil > 100 {
+		findHeaderUntil = 100
+	}
+	if bytes.Index(b[:findHeaderUntil], []byte("<?xml")) == -1 {
+		// No header found. Print it out first.
+		w.Write([]byte(xml.Header))
+	}
```
c2h5oh: When can this happen?

_examples/render/main.go
VojtechVitek: Currently, I support:
```go
render.Respond(render.Status(ctx, 409), "Conflict!!")
```
 right now -- do you like it?

VojtechVitek: I may change the signatures a bit for chi v2 -- maybe it can be nicer at the end:

```go
render.Respond(w, r, "Success by default")

// Custom 409:
r = render.Status(r, 409)
render.Respond(w, r, "Conflict!")

// or:
render.Respond(w, render.Status(r, 409), "Conflict!")
```

pkieltyka: oh cool.. so you put 409 in the ctx and then pick it up when you respond..? that is cool. 

one other thing still, is how do we get the request Content-Type to the responder?

render/present.go
VojtechVitek: I check the limit max depth on runtime Present() -- your suggestion sounds better, let me try to improve it

mutex = yes

render/render.go
VojtechVitek: true :)

render/respond.go
VojtechVitek: https://tip.golang.org/pkg/encoding/json/#Encoder.SetEscapeHTML

render/respond.go
VojtechVitek: On
```go
render.XML(w, "<nothing></nothing>")
```

This would respond with
```
<?xml version="1.0" encoding="UTF-8"?>
<nothing></nothing>
```

_examples/render/main.go
VojtechVitek: yea -- custom StatusCode can be stored in ctx --- so any middleware can change the default behavior, which is 200 for GET/PUT, 201 for POST and 204 for DELETE :)

For chi v2, I'm in favor of passing whole request to the Responder.

```go
render.Respond(w, r, "hey")
```

this way, we can grab Content-Type from r.Header, context from r.Context() etc.

render/present.go
VojtechVitek: +1

render/respond.go
```diff
@@ -0,0 +1,111 @@
...
+	"encoding/xml"
+	"net/http"
+	"reflect"
+
+	"golang.org/x/net/context"
+)
+
+var Respond = DefaultRespond
+
+func DefaultRespond(ctx context.Context, w http.ResponseWriter, v interface{}) {
```
c2h5oh: Responder should be a middleware and get it's body and headers from context.

This is go 1.7 middleware with context:
```go
func someMiddleware(next http.Handler) http.Handler {
        // before code
        next.ServeHTTP(w, r)
        // after code
    })
}
```
By having responder write to response writer in handler you prevent all middlewares in chain to set HTTP headers in `// after code`
IMO having `"response_headers": *http.Header` and `"response_body": interface{}` in context + responder as a middleware mounted very high (top level by default?) up the chain would be more flexible.

VojtechVitek: I don't agree.

I definitely don't want to store Body into context (pointer to an extra []byte buffer?) in any way. I want the render pkg to be able to stream data directly from `io.Reader` to `ResponseWriter` without any performance penalty.

The only header that we need to postpone is the last one - StatusCode. There's no need of postponing any other headers, since you can write them directly to `ResponseWriter`.

To your point: You can always set any headers in the `// before code` part of each middleware. After you `.Write()` to the `ResponseWriter`, you can't set any headers anymore -- that's how stdlib works too because of HTTP design.

render/respond.go
c2h5oh: See my comment about renderer being a middleware - makes sense together

render/respond.go
pkieltyka: I think thats a pretty interesting idea. We could even make it a requirement of the responder. 

```golang
func routes() http.Handler {
  r := chi.NewRouter()
  r.Use(middleware.Logger)
  r.Use(render.Responder)
  r.Get(....)
  return r
}

  // a call to Respond() would just put the response status/body on the r.Context()
  // and leave it up to the Responder to return..
  // in fact, Respond() could detect that Responder was set in the chain, because
  // it could check for a context value, and either respond right there, or defer
  // to the Responder middleware
  render.Respond(w, r, 200, "something")
}
```

pkieltyka: @VojtechVitek does make a good point about performance and loading far too much data into the context

pkieltyka: as well.. @c2h5oh middlewares can set headers before they call the next handler, not after. In practice, this has worked just fine

render/respond.go
c2h5oh: Go array zero value is `nil` not `[]type{}` for a reason. We rely on that distinction for child posts.

pkieltyka: yea I think we should remove it, and leave it for someone else to wrap their preference like you're both describing

render/respond.go
pkieltyka: interesting, there we go.. so.. I think, either call that method since this subpackage is for chi v2 which is Go 1.7-only, or copy the tip stdlib code into this function

render/respond.go
VojtechVitek: makes sense, will do

render/respond.go
c2h5oh: If you pass a pointer it's going to be 4 bytes - non issue.

Middlewares being unable to add headers after blocks us from changing the headers in middlewares based on response content (like for example generic pager middleware - which would be super neat :D) 

Another limitation: right now middlewares are forced to write headers to the response, so a middleware can't override a header set by a previous middleware (or strip the header).

render/render.go
c2h5oh: ah, good point.

render/respond.go
c2h5oh: I prefer plain text to math ContentType

render/respond.go
VojtechVitek: ### 1
it's not about size of the buffer/pointer in ctx -- it's more about when you're ready to start sending first bytes to `ResponseWriter` and thus to the client

I want to support this for example:
```go
render.Respond(r, w, slowS3FileIOReader) // io.Reader
render.Respond(r, w, slowChannelOfData) // chan Data
```
- this should start streaming bytes to the client right away
- and most importantly, it should **block the handler**, so you can use `defer` in the handler for clean-up
- if you don't, you'll end up buffering lot of data in the context, or you'll loose defer functionality in handlers

### 2
`generic pager` code can live in our custom `Respond()` function logic - there's no need to wait for a middleware layer down the chain on a way back from the handler to do the same thing.

### 3
This is how stdlib works too. Imho, no need to be more fancy. I don't see any problem with overriding the headers:
```go
w.Header().Set("Header", "first value")
w.Header().Set("Header", "no, sorry, this is the right value")
```

pkieltyka: I'd much rather defer something that bold for a second concept in another PR

c2h5oh: Ad. 1. Matters only for a streaming response of which we have 1 and that could be handled as a special case.

Ad. 2. Also an option

Ad. 3 This will only work until the 1st byte of the body is written, so it will not work anywhere in middleware chain in the code run after `next.ServeHTTP`

c2h5oh:
```go
w.Header().Set("Header", "first value")
w.Header().Set("Header", "no, sorry, this is the right value")
```
equals
`w.Header().Set("Header", []string{"first value", "no, sorry, this is the right value")}`
and not
`w.Header().Set("Header", "no, sorry, this is the right value")`

render/respond.go
VojtechVitek: Added TODO, will enable streaming+SetEscapeHTML in chi v2 only.

VojtechVitek:
Closing this PR and will open new one against v2.

PR #59, opened 2016-07-04
v2: Render + Presenter

VojtechVitek:
Example:
https://github.com/pressly/chi/tree/v2_presenter/_examples/render


_examples/render/main.go
```diff
@@ -0,0 +1,107 @@
...
+	"github.com/pressly/chi/render"
+)
+
+func main() {
+	r := chi.NewRouter()
+
+	r.Use(middleware.RequestID)
+	r.Use(middleware.Logger)
+	r.Use(middleware.Recoverer)
+	r.Use(render.ParseContentType)
```
pkieltyka: do we still need render.ParseContentType() ..? perhaps we can rename this or do something else to get the requested content-type?

VojtechVitek: Getting rid of it right now

_examples/render/main.go
```diff
@@ -0,0 +1,107 @@
...
+	r.Use(render.UsePresenter(v3.Presenter)) // API version 3 (latest) by default.
+
+	// API version 3.
+	r.Get("/", redirectV3)
+	r.Route("/v3", func(r chi.Router) {
+		r.Use(render.UsePresenter(v3.Presenter))
```
pkieltyka: this is awesome! very cool

render/render.go
```diff
@@ -1,84 +1,9 @@
...
-		v = reflect.MakeSlice(val.Type(), 0, 0).Interface()
-	}
-
-	// TODO: support other types based on request Accept header?
-	// render.Respond(ctx, w, 200, data)
-
-	JSON(w, status, v)
-}
+const (
+	statusCtxKey ctxKey = iota
```
pkieltyka: perhaps we should do something more like.. https://github.com/pressly/chi/blob/v2/context.go#L9 ? it's still very efficient since its just a pointer, and will make debugging / printing the request context more clear

VojtechVitek: sure

_examples/render/presenter/v1/article.go
```diff
@@ -0,0 +1,35 @@
...
+	Data map[string]bool `json:"data" xml:"data"`
+}
+
+var Presenter = render.NewPresenter(ArticleV2ToV1)
+
+func init() {
+	Presenter.RegisterFrom(v2.Presenter)
+}
+
+func ArticleV2ToV1(ctx context.Context, from *v2.Article) (*Article, error) {
```
VojtechVitek: I'm gonna change this signature to

```go
func(r *http.Request, from *v2.Article) (*Article, error) {}
```

so we can use `r.Host`, `r.URL`, `r.RequestURI`, `r.RemoteAddr`, `r.TSL` etc. in the presenter funcs

pkieltyka: I think a good idea.

render/respond.go
```diff
@@ -0,0 +1,113 @@
...
+	// 	return
+	// }
+
+	b, err := json.Marshal(v)
+	if err != nil {
+		http.Error(w, err.Error(), http.StatusInternalServerError)
+		return
+	}
+
+	if len(b) > 0 {
```
pkieltyka: should we get rid of this, etc...? since we're go 1.7 only?

VojtechVitek: definitely

pkieltyka:
btw, is there a way to stream the resp body?

render/respond.go
```diff
@@ -0,0 +1,114 @@
...
+	}
+	if bytes.Index(b[:findHeaderUntil], []byte("<?xml")) == -1 {
+		// No header found. Print it out first.
+		w.Write([]byte(xml.Header))
+	}
+
+	w.Write(b)
+}
+
+func Status(r *http.Request, status int) *http.Request {
```
VojtechVitek: Thinking of renaming this func to `WithStatus()` so it's clear it only changes request context

_examples/render/main.go
```diff
@@ -0,0 +1,148 @@
...
+			render.Respond(w, r, errors[rand.Intn(len(errors))])
+			return
+		}
+		next.ServeHTTP(w, r)
+	})
+}
+
+func init() {
+	// custom responder that sets response status code based on Error value/type.
+	render.ModifyResponse = func(r *http.Request, v interface{}) interface{} {
```
VojtechVitek: `render.Respond()` is not a variable anymore -- I like `render.ModifyResponse()` better, just to give users ability to act on data object, nothing more

VojtechVitek: hm.... thinking of it, the signature might be the same as for Presenter conversion funcs, let me check if I can improve the API one step further

nhooyr:
What exactly does this PR accomplish? What is render and a presenter?

VojtechVitek:
@nhooyr it lets you present objects in a desired format (based on Accept header) and let's you version API responses. Check this example: https://github.com/pressly/chi/tree/v2_presenter/_examples/render

Does this answer your Q?

PR #60, opened 2016-07-05
Subrouters refactor, build a single tree

pkieltyka:


util.go
```diff
@@ -23,5 +23,5 @@ func chain(middlewares []func(http.Handler) http.Handler, endpoint http.Handler)
 // 405 Method not allowed.
 func methodNotAllowedHandler(w http.ResponseWriter, r *http.Request) {
 	w.WriteHeader(405)
-	w.Write([]byte(http.StatusText(405)))
+	w.Write(nil)
```
VojtechVitek: w.Write(nil) is not needed, is it?

_examples/todos-resource/main.go
```diff
@@ -0,0 +1,46 @@
...
+		w.Write([]byte("."))
+	})
+
+	// r.Mount("/", usersResource{}.Routes()) // TODO: never gets seen...
+	// r.Mount("/", todosResource{}.Routes())
+
+	r.Append(
+		usersResource{}.Routes(),
+		todosResource{}.Routes(),
+	)
```
VojtechVitek: vojtech [4:54 PM]  
btw: I kind of think that we shouldn't add more methods to the router, we should keep it minimalistic and simple

ie. can't we go with `Mount("/", ...chi.Router)` instead of Append()?

_examples/rest/main.go
```diff
@@ -89,11 +89,8 @@ func main() {
 
 	// RESTy routes for "articles" resource
 	r.Route("/articles", func(r chi.Router) {
-		r.Group(func(r chi.Router) {
-			r.Use(paginate)
-			r.Get("/", listArticles) // GET /articles
-		})
-		r.Post("/", createArticle) // POST /articles
+		r.Get("/", chi.Use(paginate).Handler(listArticles)) // GET /articles
```
VojtechVitek: Trying to see if these alternatives would look better:
```diff
- r.Get("/", chi.Use(paginate).Handler(listArticles)) // GET /articles
+ r.Get("/", chi.Use(paginate).On(listArticles)) // GET /articles
+ r.Get("/", chi.Use(paginate)(listArticles)) // GET /articles
```

VojtechVitek:
👍

Issue #67, opened 2016-08-05
Mux ServeHTTP doesn't supports context of generic implementation


mrcpvn:
Hello, if the input request of ServeHTTP contains an existing context (but not a chi.Context), the context is discarded and a new chi.Context taken from the pool is set in the request. The data of the previous context is lost.

pkieltyka:
See https://github.com/pressly/chi/pull/69

PR #68, opened 2016-08-05
Existing context

mrcpvn:
#67 
this solves the issue, it is the same change made in v1


pkieltyka:
Thanks for the PR and report, I've gone in this direction which is a bit more efficient: https://github.com/pressly/chi/pull/69

PR #69, opened 2016-08-05
Set parent context on a route context

pkieltyka:
Fixes https://github.com/pressly/chi/issues/67 and alternate solution to https://github.com/pressly/chi/pull/68

### DIFF
--- a/context.go
+++ b/context.go
@@ -3,6 +3,7 @@ package chi
 import (
 	"context"
 	"net/http"
+	"time"
 )
 
 var (
@@ -11,10 +12,11 @@ var (
 
 var _ context.Context = &Context{}
 
-// A Context is the default routing context set on the root node of a
+// Context is the default routing context set on the root node of a
 // request context to track URL parameters and an optional routing path.
 type Context struct {
-	context.Context
+	// Parent context
+	parent context.Context
 
 	// URL routing parameter key and values.
 	Params params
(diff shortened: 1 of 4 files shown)


context.go
```diff
@@ -11,10 +12,11 @@ var (
 
 var _ context.Context = &Context{}
 
-// A Context is the default routing context set on the root node of a
+// Context is the default routing context set on the root node of a
 // request context to track URL parameters and an optional routing path.
 type Context struct {
-	context.Context
+	// Parent context
+	parent context.Context
```
mrcpvn: this will make Context unexported. 
the context implementations of the standard library has Context exported and this will also break compatibility with chi v1. For example:
c := chi.Context{}
c.Context = context.Background()
will not compile.

pkieltyka: @mrcpvn thanks for the note, however, chi v2 already breaks a lot of compatibility with v1, so that is okay. the chi.Context is itself a context.Context object, because it implements the context.Context interface, and uses the parent context.Context for behaviour (see implementations of the methods on chi.Context). 

@mrcpvn do you want to do this for testing..? we could export the "Parent", but I'd rather not unless we have to... this works https://github.com/pressly/chi/blob/master/mux_test.go#L833  .. is that not enough? let me know

mrcpvn:
@pkieltyka thank you. 
Exporting "Parent" will resolve the compatibility problem but I agree with you that chi.Context is a context.Context and that is what is needed.
I used to set the parent context to mock the router but I can solve easily this problem testing with httptest package.
The v1 compatibility is not a problem so for me no code change is needed in chi v2.

PR #70, opened 2016-08-10
Automatic doc generation from Router

pkieltyka:


chi.go
```diff
@@ -45,7 +45,7 @@ type Router interface {
 	Use(middlewares ...func(http.Handler) http.Handler)
 
 	// Yep..
```
VojtechVitek: Yup :)

Issue #84, opened 2016-09-08
Render: managing request and response payload as types


pkieltyka:
The render subpkg is pretty cool, and its great for managing the different kinds of content types a response may need, including streams. However, one of the important parts for a maintainable API is to manage the request and response payloads (the inputs and the outputs of an endpoint).

For request payloads, one simple idea is to have a Bind() middleware used in chi's `r.With()` inline middleware routing, that will take a request body and unmarshal it to a kind of struct. Perhaps.

``` golang
type ArticleRequest {
  ID int64 `json:"id"`
  Name string `json:"name"`
}
// ...
r.Get("/", index)
r.With(render.Request(&ArticleRequest{})).Post("/articles", newArticle)
```

.. something like that.. the render.Request() would make a new &ArticleRequest{} object and put it on the context under render.RequestKey or something.. its a decent plan, except it would require some kind of reflection to make the new &ArticleRequest{} object.

As for response payloads, that should be much simpler, just..

``` golang
type ArticleResponse {
  *data.Article // embed the core Article type
  // add more methods on it..
}
//...
render.Respond(w, r, &ArticleResponse{a})
```

if someone has other ideas too that would be awesome!

elithrar:
Thanks for raising this! Some thoughts:

* Right now there are two popular libs (amongst many others) that tackle this, with different (but large) APIs: https://github.com/mholt/binding and http://www.gorillatoolkit.org/pkg/schema - thus: *what would render.Bind` do differently here*?
* I like that—from the example—it's integrated with chi, but you will undoubtedly get requests around skipping fields in the request body, up/down-casing fields, etc (take a look at the issues/PRs for gorilla/schema...). 
* `reflect` is unavoidable if this is going to be plug-and-play.

pkieltyka:
hey @elithrar @VojtechVitek and others, I did a bit of work on the render subpkg today. Ive written request body decoding to be a bit nicer and I introduced render.Bind() as a middleware to automatically bind a request body to a struct value/object.

https://github.com/pressly/chi/pull/85/files

look at CreateArticle and CreateArticle2 for both versions and check the router definition too. After doing it though, I like the Decode() stuff, but I'm probably going to remove the auto-binding stuff since it doesnt help that much.

elithrar:
@pkieltyka 

Overall - like it a lot. It's simple, although we should expose decoding errors to make it clear when the POST body was empty vs. we couldn't decode (for whatever reason).

pkieltyka:
[...]

PR #86, opened 2016-09-17
[docs] Document public symbols; satisfy lint/vet.

elithrar:
WIP.

- [x] Satisfied golint for most of the *render* package.
- [x] Removed unreachable code: noting that although you had `panic("unreachable"), govet still complained. If there's a reason not to trust in go vet, we should document that + fix go vet.

Remaining:
- [ ] Understand the `Presenter` interface more so we can document it.
- [ ] Bind is an exported package-level variable. What is the use-case beyond `defaultBind`?

Otherwise: *great* package.

### DIFF
--- a/render/content_type.go
+++ b/render/content_type.go
@@ -13,6 +13,7 @@ var (
 // A ContentType is an enumeration of common HTTP content types.
 type ContentType int
 
+// ContentTypes handled by this package.
 const (
 	ContentTypePlainText = iota
 	ContentTypeHTML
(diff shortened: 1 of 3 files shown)


ChrisHines:
The `panic("unreachable")` lines were required by Go prior to v1.1, see https://golang.org/doc/go1.1#return.

pkieltyka:
thanks for the PR! I agree we don't need the panic("unreachable"), this project only works with go 1.7+ anyways.

pkieltyka:
if you're looking at the render subpkg, I'd love your opinions on #84. I plan to revise `render` in parts to better consider well-defined request and response payload types. #84 would also likely change some of the Presenter stuff as well.

PR #87, opened 2016-09-18
[WIP] Routes Walker + RAML generator

VojtechVitek:
Auto-generated developer docs for _examples/versions:
screen shot 2016-09-26 at 7 29 14 pm


walker.go
```diff
@@ -0,0 +1,57 @@
...
+			}
+			continue
+		}
+
+		hall := route.Handlers["*"]
+		for method, handler := range route.Handlers {
+			// TODO: Hrm.. this shouldn't be necessary.
+			// Imho, r.Routes() shouldn't give us '*' method plus all the methods.
+			// And, we shouldn't call fn() with '*' method at all.
+			if method != "*" && hall != nil && fmt.Sprintf("%v", hall) == fmt.Sprintf("%v", handler) {
```
VojtechVitek: @pkieltyka I'd like to get rid of `'*'` routes that are generated by `r.Mount()`.

walker.go
```diff
@@ -0,0 +1,57 @@
...
+				continue
+			}
+
+			// TODO: Hrm.. Most of the time, the route.Pattern ends with an
+			// extra '/', even though it doesn't match the actual route, why?
+			fullRoute := strings.TrimRight(parentRoute+route.Pattern, "/")
```
VojtechVitek: `route.Pattern` returns route ending with slash, even though there are no other subroutes. I'd like to change this, if there's no reason behind it.

walker.go
```diff
@@ -0,0 +1,57 @@
+package chi
+
+import (
+	"fmt"
+	"net/http"
+	"strings"
+)
+
+type WalkFn func(method string, route string, handler http.Handler, middlewares ...func(http.Handler) http.Handler) error
+
```
VojtechVitek: @pkieltyka `Walk()` uses native stdlib signatures only.

If anyone needs extra info about a handler or middleware, they can call `chi.GetFuncInfo(handler)`

walker_test.go
```diff
@@ -0,0 +1,17 @@
+package chi
+
+import (
+	"net/http"
+	"testing"
+)
+
+func TestWalker(t *testing.T) {
+	r := MuxBig()
```
VojtechVitek: This test can test table with expected routes for MuxBig.

info.go
```diff
@@ -0,0 +1,113 @@
...
+	"go/parser"
+	"go/token"
+	"os"
+	"path/filepath"
+	"reflect"
+	"runtime"
+	"strings"
+)
+
+type FuncInfo struct {
```
pkieltyka: I'd prefer to keep these function-introspective functions in the `docgen` package.

pkieltyka: as well, remove info.go and move it to docgen/ package, and converge existing logic that is there

walker.go
```diff
@@ -0,0 +1,57 @@
...
+func Walk(r Routes, fn WalkFn) error {
```
pkieltyka: Cool - btw, I didn't get my head so deep to trace this logic - can you tell me why this walk method is better then just using the chi.Routes interface on a router? just like docgen does.. is there some extra data you're able to reach into?

pkieltyka: I dont really want to add more logic to the core `chi` package

VojtechVitek: `Walk()` gives you:
- full expanded flat routes with their middleware stack (with `chi.Routes`, developer would have to traverse all the nested routes himself and build the full path and middleware stack via recursion/loop -- which would be duplicated work)
- is very convenient for 1. custom docgen generation; 2. writing automatic tests (eg. testing all `/admin/users` endpoints and making sure that there's a certain middleware applied)
- gives more power to developers, since they can customize docgen generation process easily... for example they can filter out routes that are not public, remove TODOs from comments etc. on the fly -- they can even create two different docs: 1. Internal Developer Docs (with details about Handlers, Middlewares and file structure) and 2. Public API Docs (with less info for the API consumers)
- the signature is very similar to https://golang.org/pkg/path/filepath/#Walk

btw: I used to have Walk() func in a separate walker pkg in the beginning. But it makes a lot of sense to me that it should be part of the core.

info.go
VojtechVitek: Agreed, makes sense

walker.go
VojtechVitek: Example - print all routes in the following simple format:
```
POST /users
GET /users
GET /users/:id
PUT /users/:id
GET /users/:id/books
```

With `Walk()` fn, it's as simple as:
```go
chi.Walk(r, func(method string, route string, handler http.Handler) error {
	fmt.Printf("%v %v\n", method, route)
})
```

zbindenren:
Any news on this? Does that have any chance to get meged?

BTW: gorilla.mux also provides a walk function.

pkieltyka:
@zbindenren what are you trying to achieve?

btw, chi has https://github.com/pressly/chi/blob/master/chi.go#L82-L90 for very simple walking, you can see example use in `docgen` package.

zbindenren:
@pkieltyka I would like to see the RAML generator in docgen. Would be great!

VojtechVitek:
@zbindenren I'll finish this PR in the upcoming days.

zbindenren:
Great! Thanks for sharing this project.

zet4:
It's about 2 months now, any updates on this? Would love to have RAML in my API's too.

VojtechVitek:
Update: We're still discussing the internals, since the Walk() function, as proposed in this PR, is quite similar to walk() function used internally in the mux.

zbindenren:
Is there a reason why walk() is not public?

VojtechVitek:
Split to https://github.com/go-chi/chi/pull/222 and a subsequent work in https://github.com/go-chi/docgen

PR #93, opened 2016-10-14
Add URLParam* helper functions

kanocz:
Just few functions to convert URLParam to int, uint, float and boolean

### DIFF
--- a/context.go
+++ b/context.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"net"
 	"net/http"
+	"strconv"
 )
 
 var (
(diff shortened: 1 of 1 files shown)


kanocz:
In case of just URLParam you also receive "" without any errors if param doesn't exists and in many cases (for example: item id) you're expecting positive integer and zero indicates error itself

and I think that code like

```
  item := chi.URLParamUint(r, "item")
  timestamp := chi.URLParamUint(r, "timestamp")
  if 0 == item || 0 == timestamp { /* error */ }
```

looks just readable and simple

and this makes possible to write simple construction like

```
func httpGetItem(w http.ResponseWriter, r *http.Request) {
  data, err := getItem(chi.URLParamUint(r, "item"))
  if nil != err {
     // ......
  }
}
```
where we really don't need to verify - database doesn't contain "zero" item usualy

P.S.: my opinion is just "simpler code = less errors"

pkieltyka:
@kanocz thanks a lot for the PR! I think though these sort of helper functions belong in the application code, or in an app subpkg.

"Perfection is finally attained not when there is no longer anything to add, but when there is no longer anything to take away." -Antoine de St. Exupery

PR #96, opened 2016-10-20
Various linter patches

mvdan:


chain.go
```diff
@@ -33,7 +33,7 @@ func (c *ChainHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 // handler in the order they are passed.
 func chain(middlewares []func(http.Handler) http.Handler, endpoint http.Handler) http.Handler {
 	// Return ahead of time if there aren't any middlewares for the chain
-	if middlewares == nil || len(middlewares) == 0 {
+	if len(middlewares) == 0 {
```
pkieltyka: thanks for the PR but Im surprised by this one.. can you send over the linter output in a comment in this PR?

mvdan: This was `gosimple`: https://github.com/dominikh/go-simple

In particular:

> Don't check if slices, maps or channels are nil before checking their length, it's redundant. len is defined as zero for those nil values.

pkieltyka:
I think its good to run chi through a linter, but I'd like it to be the standard one, but one author's version of it, it's a whack-a-mole game

chain.go
VojtechVitek: makes sense, the nil check is not necessary
https://play.golang.org/p/rj8xKahgRx

middleware/terminal.go
```diff
@@ -4,30 +4,24 @@ package middleware
...
 var (
 	// Normal colors
-	nBlack   = []byte{'\033', '[', '3', '0', 'm'}
-	nRed     = []byte{'\033', '[', '3', '1', 'm'}
-	nGreen   = []byte{'\033', '[', '3', '2', 'm'}
-	nYellow  = []byte{'\033', '[', '3', '3', 'm'}
-	nBlue    = []byte{'\033', '[', '3', '4', 'm'}
-	nMagenta = []byte{'\033', '[', '3', '5', 'm'}
-	nCyan    = []byte{'\033', '[', '3', '6', 'm'}
-	nWhite   = []byte{'\033', '[', '3', '7', 'm'}
```
VojtechVitek: can we comment out these unused vars instead of removing them?

mvdan:
The standard linter does some things, other linters do others. This PR isn't to always force all the linters. I simply ran some of them and applied some sane improvements and cleanups.

middleware/terminal.go
mvdan: Never a fan of leaving commented out code. I can simply remove the commit from the PR.

Issue #97, opened 2016-10-22
Customizable Logger middleware output


yargevad:
It's not currently possible to customize the output of the Logger middleware at all. I spent a little time on that, will send a PR shortly.

pkieltyka:
Done! chi's middleware.Logger now supports quite well logging customization. Here is an example structured logger that can be adapted to any other backend: https://github.com/pressly/chi/blob/master/_examples/logging/main.go

PR #98, opened 2016-10-22
allow customizing log format

yargevad:
Addresses #97 - comments welcome

### DIFF
--- a/middleware/logger.go
+++ b/middleware/logger.go
@@ -13,13 +13,17 @@ import (
 	"time"
 )
 
-// DefaultLogger contains all the data needed to construct a log message
-type DefaultLogFormatter struct{}
+// DefaultLogFormatter contains all the data needed to construct a log message
+type DefaultLogFormatter struct {
+	buf bytes.Buffer
+}
 
 // LogFormatter allows you to customize this middleware's output format.
 type LogFormatter interface {
-	FormatRequest(r *http.Request) *bytes.Buffer
-	FormatResponse(buf *bytes.Buffer, status, bytes int, elapsed time.Duration)
+	FormatRequest(r *http.Request) LogFormatter
+	FormatResponse(status, bytes int, elapsed time.Duration)
+	Log()
+	Recover(err error)
 }
 
 var defaultLogFormatter *DefaultLogFormatter
@@ -38,14 +42,14 @@ func Logger(next http.Handler) http.Handler {
 // for each request/response.
 func FormattedLogger(f LogFormatter, next http.Handler) http.Handler {
 	fn := func(w http.ResponseWriter, r *http.Request) {
-		buf := f.FormatRequest(r)
+		ctx := f.FormatRequest(r)
 		lw := wrapWriter(w)
 
 		t1 := time.Now()
 		defer func() {
 			t2 := time.Now()
-			f.FormatResponse(buf, lw.Status(), lw.BytesWritten(), t2.Sub(t1))
-			log.Print(buf.String())
+			ctx.FormatResponse(lw.Status(), lw.BytesWritten(), t2.Sub(t1))
+			ctx.Log()
 		}()
 
 		next.ServeHTTP(lw, r)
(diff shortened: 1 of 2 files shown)


pkieltyka:
@yargevad thanks for this PR! looks good, I am going to dive into it and put some brain power behind it before I merge it. Can you also provide an example format you use that drove the reason for the addition to customize the output?

yargevad:
The reason was to be able to emit structured logs a la [uber-go/zap](https://github.com/uber-go/zap) - here's [my zap implementation](https://github.com/yargevad/dengo/blob/master/201610/errors.go).

pkieltyka:
@yargevad nice, this is awesome! exactly what we need

pkieltyka:
@yargevad btw, https://github.com/yargevad/dengo/blob/master/201610/routes.go#L10 you can return `http.Handler` or `chi.Router` instead of a `*chi.Mux`

VojtechVitek:
This is awesome. 👍 

But before we merge, I'd like to discuss the `LogFormatter interface`. I will take a closer look later today / tmrw.

yargevad:
There are a few improvements I should still make here, for example `defaultLogFormatter` doesn't really need to be a struct, and `Recoverer` calls `debug.PrintStack()` which doesn't have a chance to use the customized logger.

Let me know what other changes you have in mind and I'll work on those too.

middleware/logger.go
```diff
@@ -37,59 +58,64 @@ func Logger(next http.Handler) http.Handler {
 	return http.HandlerFunc(fn)
 }
 
-func requestPrefix(reqID string, r *http.Request) *bytes.Buffer {
-	buf := &bytes.Buffer{}
+// FormatRequest stores info from the request, for use by Log()
+func (l *DefaultLogFormatter) FormatRequest(r *http.Request) LogFormatter {
+	var ctx DefaultLogFormatter
```
VojtechVitek: why do you call it `ctx`? that's rather confusing, since `ctx := r.Context()` is commonly used for request context

middleware/logger.go
```diff
@@ -13,22 +13,43 @@ import (
...
+type DefaultLogFormatter struct {
+	buf bytes.Buffer
+}
+
+// LogFormatter allows you to customize this middleware's output format.
+type LogFormatter interface {
+	FormatRequest(r *http.Request) LogFormatter
+	FormatResponse(status, bytes int, elapsed time.Duration)
+	Log()
+	Recover(err error)
```
VojtechVitek: Why do we need both FormatRequest and FormatResponse methods? Can't we just call one method only, on a way back from the underlying handlers? I don't see the reason why we'd need to store the intermediate log into a bytes.Buffer.

What's the reason for Log() and Recover()? Can't we live without them too?

pkieltyka:
Yea I've been prototyping some slight changes on the side too. The LogFormatter is a good idea and I've made a few improvements, I'll be done hopefully next week but I'm pretty busy with normal work too. In the mean time you could make your zap logger outside of chi/middleware

middleware/logger.go
yargevad: You're right, this could be simplified down to one method, like:

    FormatLog(r *http.Request, status, bytes int, elapsed time.Duration, err error)

With `err == nil` meaning "log normal req/res info", otherwise ignore `status, bytes, elapsed` and show the error and stack trace.

yargevad: Question: would you prefer one dual-purpose method, or two methods, one called from Recoverer when there's a panic, and one from Logger during normal operation?

jiop:
Hi, is there any update about this pull request? I'd love to use it ;)

yargevad:
Hi @jiop, I've been using [my forked version](https://github.com/yargevad/chi/blob/master/middleware/logger.go) until we work through any other changes.

pkieltyka:
@yargevad thanks for inspiring a lot of ideas! I've just pushed https://github.com/pressly/chi/commit/b228440f3ccdc3cc4a4f528fc4ed6700e705ade2 and https://github.com/pressly/chi/commit/7c51c558b51f5740cec86b1b69634f72f17ae72f

PR #107, opened 2016-11-10
MiddlewareFunc prototype for easier middleware functions

VojtechVitek:
I'd like to discuss and get some feedback on my proposal for `MiddlewareFunc` type that lets you write HTTP Middlewares in a single function without the need for a wrapper func (see below). The concept is very similar to stdlib's `http.HandlerFunc`.

This proposal lets you use

```go
func ArticleCtx(w http.ResponseWriter, r *http.Request, next http.HandlerFunc) {
	articleID := chi.URLParam(r, "articleID")
	article, err := dbGetArticle(articleID)
	if err != nil {
		render.Status(r, http.StatusNotFound)
		render.JSON(w, r, http.StatusText(http.StatusNotFound))
		return
	}
	ctx := context.WithValue(r.Context(), "article", article)
	next(w, r.WithContext(ctx))
}
```

instead of

```go
func ArticleCtx(next http.Handler) http.Handler {
		articleID := chi.URLParam(r, "articleID")
		article, err := dbGetArticle(articleID)
		if err != nil {
			render.Status(r, http.StatusNotFound)
			render.JSON(w, r, http.StatusText(http.StatusNotFound))
			return
		}
		ctx := context.WithValue(r.Context(), "article", article)
		next.ServeHTTP(w, r.WithContext(ctx))
	})
}
```

If that make any sense to you guys, I might try to submit these signatures to stdlib `net/http` pkg. Having standardized middleware signatures would be great addition to the language.

Feedback welcome!

### DIFF
--- a/chi.go
+++ b/chi.go
@@ -1,5 +1,5 @@
 //
-// Package chi is a small, idiomatic and composable router for building HTTP services.
+// chi is a small, idiomatic and composable router for building HTTP services.
 //
 // chi requires Go 1.7 or newer.
 //
@@ -45,8 +45,14 @@ type Router interface {
 	// Use appends one of more middlewares onto the Router stack.
 	Use(middlewares ...func(http.Handler) http.Handler)
 
+	// UseFunc appends one of more middleware functions onto the Router stack.
+	UseFunc(fns ...MiddlewareFunc)
+
 	// With adds inline middlewares for an endpoint handler.
 	With(middlewares ...func(http.Handler) http.Handler) Router
+
+	// WithFunc adds inline middleware functions for an endpoint handler.
+	WithFunc(fns ...MiddlewareFunc) Router
 
 	// Group adds a new inline-Router along the current routing
 	// path, with a fresh middleware stack for the inline-Router.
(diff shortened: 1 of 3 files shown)


chi.go
```diff
@@ -89,6 +95,21 @@ type Routes interface {
...
+type Middleware func(http.Handler) http.Handler
+
+// Middlewares type is a slice of middleware handlers with methods
 // to compose middleware chains and http.Handler's.
 type Middlewares []func(http.Handler) http.Handler
+
+// MiddlewareFunc type is an adapter to allow the use of ordinary
+// functions as middleware handlers. If f is a function with the
+// appropriate signature, MiddlewareFunc(f) is a Handler that calls f.
+type MiddlewareFunc func(w http.ResponseWriter, r *http.Request, next http.HandlerFunc)
```
VojtechVitek: Note how similar this is to `http.HandlerFunc`:
https://github.com/golang/go/blob/b20c055230a20663c75e3099f672c15c39d46b9e/src/net/http/server.go#L1893-L1902

VojtechVitek:
Actually, I just realized that this PR reimplements Negroni middleware func signatures. Lol.

So all of these middlewares come for free:
https://github.com/urfave/negroni#third-party-middleware

pkieltyka:
Thanks @VojtechVitek for the work! I'm also interested to hear the opinion of others for this kind of PR. The pros/cons as I see it are that, pro: middleware signatures are easier to remember, con: it is never the less non-standard and breaks interoperability with stdlib. The premise of chi is to be completely interoperable with the stdlib and any other handlers/middlewares that are built with it. But maybe this middleware signature is common too, so let's have it sit and see what the community thinks.

pkieltyka:
going to close this, as its best to keep it completely standard

PR #111, opened 2016-11-17
Update not found handlers recursively fix #110

### LINKED ISSUE #110: Not found handler for subrouters
Consider this example:

```go
func Create() {
  r := chi.NewRouter()
  r.NotFound(notFound) // weird, if you set NotFound after Mount - it won't affect subrouter at all
  r.Mount("/api", protected(authService))
}

func protected() chi.Router {
  r := chi.NewRouter()
  r.Route("/group", func(r chi.Router) {
     ....
  }
  return r
}

func notFound(rw http.ResponseWriter, r *http.Request) {
  rw.Write("my not found")
}
```

Now let's try some urls:
```
/unknown - "my not found"
/api/unknown - "my not found"
/api/group/unknown - "default not found"
```

It's very unexpected behavior for me. There is workaround, I need to set my not-found handler for protected router explicitly too. Maybe on mount it's better to go through all subrouters and set not-found router there? If that's alright, I can make pull request.

smacker:
It solves 2 problems:
1. It sets not-found custom handler for all subtrees if they don't have custom handler already
2. You can define custom not-found handler before or after any mounts.

### DIFF
--- a/mux_test.go
+++ b/mux_test.go
@@ -359,6 +359,58 @@ func TestMuxNestedNotFound(t *testing.T) {
 	}
 }
 
+func TestMuxComplicatedNotFound(t *testing.T) {
+	// sub router with groups
+	sub := NewRouter()
+	sub.Route("/resource", func(r Router) {
+		r.Get("/", func(w http.ResponseWriter, r *http.Request) {
+			w.Write([]byte("private get"))
+		})
+	})
+
+	// Root router with groups
+	r := NewRouter()
+	r.Get("/auth", func(w http.ResponseWriter, r *http.Request) {
+		w.Write([]byte("auth get"))
+	})
+	r.Route("/public", func(r Router) {
+		r.Get("/", func(w http.ResponseWriter, r *http.Request) {
+			w.Write([]byte("public get"))
+		})
+	})
+	r.Mount("/private", sub)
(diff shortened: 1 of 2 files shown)


mux_test.go
```diff
@@ -359,6 +359,58 @@ func TestMuxNestedNotFound(t *testing.T) {
...
+	r.Get("/auth", func(w http.ResponseWriter, r *http.Request) {
+		w.Write([]byte("auth get"))
+	})
+	r.Route("/public", func(r Router) {
+		r.Get("/", func(w http.ResponseWriter, r *http.Request) {
+			w.Write([]byte("public get"))
+		})
+	})
+	r.Mount("/private", sub)
+	r.NotFound(func(w http.ResponseWriter, r *http.Request) {
```
pkieltyka: yea, I do like the idea that NotFound would be used all the way down from this point in the tree. If another subrouter defined NotFound it would use that one all the way down. Nice work.

However, what would happen in this test func if r.NotFound() was defined at the very top before the Mount(), I think it might fail

smacker: I thought it might fail too. So I moved it on the top and rerun. Tests passed. But maybe my test doesn't cover all possible cases. Any suggestions for improvement?

pkieltyka:
nicely done. thanks for the PR!

PR #116, opened 2016-11-30
Fix missing MethodNotAllowed in Router interface

jiop:

### DIFF
--- a/chi.go
+++ b/chi.go
@@ -77,6 +77,10 @@ type Router interface {
 	// NotFound defines a handler to respond whenever a route could
 	// not be found.
 	NotFound(h http.HandlerFunc)
+
+	// MethodNotAllowed defines a handler to respond whenever a method is
+	// not allowed.
+	MethodNotAllowed(h http.HandlerFunc)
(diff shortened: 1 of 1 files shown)


pkieltyka:
@jiop thanks for this, but a part of me would rather keep Router small, even though the Mux has a few extra methods that are available to it. In that case, one could assert to a *Mux to use it, and in actuality, chi.NewRouter() returns a *Mux. Let me know what you think.

jiop:
@pkieltyka My main concern is about consistency. NotFound handler is available in the Router interface so I thought that MethodNotAllowed should be there as well.

pkieltyka:
@jiop yea I agree with that rationale. I felt MethodNotAllowed won't be used as commonly as NotFound, so thats why I kept it a little bit lower level. I'll leave this PR open and see what other contributors think too.

jiop:
Hi @pkieltyka, any updates on this PR?

pkieltyka:
Hey Jeremie, I'm going to hold off and close it for now.

PR #118, opened 2016-12-15
add request time in default logger

chenjie4255:
I think it will be very useful if the default logger recored the request time, or the log cannot tell us anything about time.

before this changed, the default logger output looks like:

```
"GET http://127.0.0.1:3000/ HTTP/1.1" from 127.0.0.1:53322 - 200 7B in 12.29µs
```

now it will be:
```
2016/12/15 - 11:40:12 "GET http://127.0.0.1:3000/ HTTP/1.1" from 127.0.0.1:53322 - 200 7B in 12.29µs
```

### DIFF
--- a/middleware/logger.go
+++ b/middleware/logger.go
@@ -16,10 +16,10 @@ var (
 )
 
 // Logger is a middleware that logs the start and end of each request, along
-// with some useful data about what was requested, what the response status was,
-// and how long it took to return. When standard output is a TTY, Logger will
-// print in color, otherwise it will print in black and white.
-//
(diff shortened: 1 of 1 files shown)


pkieltyka:
@chenjie4255 thanks for bringing this to my attention, there has certainly been a regression here. I will fix it in a few minutes.

pkieltyka:
fixed by 18893e620db10241dd2066014f5b8ef60b2d413e

VojtechVitek:
Just to clarify: The underlying log prints the timestamps. See https://golang.org/pkg/log/#pkg-constants

PR #124, opened 2016-12-31
Add HTTP2 fancy writer

lclarkmichalek:
This gets all middleware that use NewWrapResponseWriter compatible with http.Pusher. However, it doesn't work with middleware that don't, which is currently just Compress (`maybeCompressResponseWriter`). I wasn't sure if I should refactor things to have `maybeCompressResponseWriter` use http(2?)FancyWriter, to avoid a load of duplication, or just reimplement the `fancyWriter` logic for `maybeCompressResponseWriter`.


pkieltyka:
@lclarkmichalek thanks for the PR. So from the looks of it, you've made two fancyWriters, for go<1.8 and go>=1.8, to account for the new http.Pusher interface. As well, depending on the protocol version you remove http.Hijakcer from http2 because its unsupported/incompatible with HTTP2. Does this take care of other http2 interfaces on the ResponseWriter? 

is there anything else or any other considerations we should make?

pkieltyka:
nice work btw, this looks good

lclarkmichalek:
Yeah, that's accurate. There aren't any other http2 interfaces, to the best of my knowledge, so we should be ok (until someone adds another!)

I've added support to the Compress middleware, though there's a lot of duplication. I built out a new `genericWrapper`, with a `httpGenericWrapper` and `http2GenericWrapper`, which should allow us to wrap any kind of ResponseWriter. It checks to see if a method (i.e. `Flush`, `CloseNotify`) has been implemented by the wrapper, and if not, uses the implementation from the wrapped writer (we require the wrapper to have an `Unwrap()` method for this. The `mkGenericWrapper` function uses `Unwrap()` to get the original ResponseWriter, and then selects the `httpGenericWrapper` or `http2GenericWrapper` structs to ensure that the returned ResponseWriter implements the same interfaces as the original ResponseWriter.

However, this new interface isn't compatible with the existing `WrapResponseWriter`, so there's a lot of duplicated code, as Compress couldn't be implemented in terms of the existing fancyWriters. The functionality of WrapResponseWriter could be, however the function `NewWrapResponseWriter` would need to change. How important is keeping the existing `NewWrapResponseWriter` API? I only found one usage on github.

I've also added a couple of tests to ensure that all the expected interfaces are on the ResponseWriters after being passed through a middleware using `mkGenericWriter`. 

I'm super sorry this has grown to be a larger PR. This really isn't a high point for Go's type system..

pkieltyka:
@lclarkmichalek I appreciate the work for sure! I need to get my head into this again the properly evaluate it. Sometimes when things get too complex, we need to revisit the approach and purpose

pkieltyka:
@lclarkmichalek what do you think of this? https://github.com/pressly/chi/pull/126 

The way I've seen it, the current WrapWriter fails to support the cases:

1. HTTP2 does not support http.Hijacker
2. Go1.8 introduces new http.Pusher support for HTTP2

this PR will smooth that out, for 1.8+ it adds implementation of http.Pusher() to the http2FancyWriter

PR #128, opened 2017-01-04
Http2 support for middleware, fixes #123

### LINKED ISSUE #123: HTTP2 requests are not wrapped using fancyWriter
HTTP requests do not implement `http.Hijacker` or `io.ReaderFrom`. This causes them to not be eligible for wrapping in `middleware.fancyWriter`, instead being wrapped in `middleware.flushWriter`, and losing the `http.CloseNotifier` interface. In addition, `middleware.fancyWriter` does not support `http.Pusher` (in the go 1.8 beta), meaning that even if the writers were eligible for `middleware.fancyWriter`, http2 push would not be possible.

The following snippet, adapted from https://gist.github.com/rakyll/eec415977f85d50a493ca8472ba97b68, shows the various interfaces that the http2 response writer implements:

```go
package main

import (
	"io"
	"log"
	"net/http"
)

func main() {
	http.HandleFunc("/", func(w http.ResponseWriter, r *http.Request) {
		_, cn := w.(http.CloseNotifier)
		_, fl := w.(http.Flusher)
		_, hj := w.(http.Hijacker)
		_, rf := w.(io.ReaderFrom)
		_, ps := w.(http.Pusher)

		log.Printf("rw: %T", w)
		log.Printf("cn: %v", cn)
		log.Printf("fl: %v", fl)
		log.Printf("hj: %v", hj)
		log.Printf("rf: %v", rf)
		log.Printf("ps: %v", ps)
	})
	// Run crypto/tls/generate_cert.go to generate cert.pem and key.pem.
	// See https://golang.org/src/crypto/tls/generate_cert.go
	log.Fatal(http.ListenAndServeTLS(":7072", "cert.pem", "key.pem", nil))
}
```

Output (go 1.8beta2, accessing the server using Chrome 55.0.2883.95)

```
2016/12/31 15:06:17 rw: *http.http2responseWriter
2016/12/31 15:06:17 cn: true
2016/12/31 15:06:17 fl: true
2016/12/31 15:06:17 hj: false
2016/12/31 15:06:17 rf: false
2016/12/31 15:06:17 ps: true
```

In terms of resolution, `http.Hijacker` needs to be handled separately, as http2 will not support it: https://github.com/golang/go/issues/14797. I don't know about `io.ReaderFrom` however. 

Given that to allow interfaces to pick and choose as they like, we'd need around 25 different types, it's probably not practical to support every combination of requests. Would having a `middleware.httpFancyWriter` and a `middleware.http2FancyWriter` make sense?

pkieltyka:
cc: @lclarkmichalek


middleware/wrap_writer17.go
```diff
@@ -0,0 +1,32 @@
...
+
+// NewWrapResponseWriter wraps an http.ResponseWriter, returning a proxy that allows you to
+// hook into various parts of the response process.
+func NewWrapResponseWriter(w http.ResponseWriter, protoMajor int) WrapResponseWriter {
+	_, cn := w.(http.CloseNotifier)
+	_, fl := w.(http.Flusher)
+	_, rf := w.(io.ReaderFrom)
+
+	bw := basicWriter{ResponseWriter: w}
+	if protoMajor == 2 {
```
lclarkmichalek: This is much nicer

middleware/wrap_writer18.go
```diff
@@ -0,0 +1,39 @@
...
+// hook into various parts of the response process.
+func NewWrapResponseWriter(w http.ResponseWriter, protoMajor int) WrapResponseWriter {
+	_, cn := w.(http.CloseNotifier)
+	_, fl := w.(http.Flusher)
+	_, rf := w.(io.ReaderFrom)
+
+	bw := basicWriter{ResponseWriter: w}
+	if protoMajor == 2 {
+		_, ps := w.(http.Pusher)
+		if cn && fl && ps && rf {
```
lclarkmichalek: I don't think HTTP2 requests have ReaderFrom

middleware/middleware18_test.go
```diff
@@ -0,0 +1,71 @@
...
+		if !ps {
+			t.Fatal("request should have been a http.Pusher")
+		}
+
+		w.Write([]byte("OK"))
+	})
+
+	server := http.Server{
+		Addr:    ":7072",
+		Handler: handler,
```
lclarkmichalek: This handler should be wrapped in a middleware to have this test check that the response writer still has the interfaces expected after wrapping.

middleware/middleware18_test.go
```diff
@@ -0,0 +1,71 @@
...
+
+// NOTE: we must import `golang.org/x/net/http2` in order to explicitly enable
+// http2 transports for certain tests. The runtime pkg does not have this dependency
+// though as the transport configuration happens under the hood on go 1.7+.
+
+func TestWrapWriterHTTP2(t *testing.T) {
+	handler := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		_, cn := w.(http.CloseNotifier)
+		if !cn {
+			t.Fatal("request should have been a http.CloseNotifier")
```
lclarkmichalek: Now I think about it, these messages should probably all read `ResponseWriter should have been ...`

pkieltyka:
Yea it's still missing support for compress18

I realize it makes sense for chi to support the current and one back version going forward. So 18 and 17 now, and eventually just 19 and 18, which should reduce these nuances as http2 will be mature by then

pkieltyka:
all done. thanks for the feedback

PR #129, opened 2017-01-05
add installing section

zbindenren:
For lazy people like me who want to copy/paste the install command.


pkieltyka:
I added a one-liner for it, thanks for the PR anyways

PR #135, opened 2017-01-15
Remove the content-length header from compressed responses

lclarkmichalek:
We can't allow this header through as it should be set to the length of the compressed body. We don't know that length, so we just delete it. Sending the incorrect content-length didn't seem to cause issues when requesting the server directly, but the Amazon load balancer really did not appreciate the incorrect headers (gave me SPDY errors in Chrome)

### DIFF
--- a/middleware/compress.go
+++ b/middleware/compress.go
@@ -133,6 +133,8 @@ func (w *maybeCompressResponseWriter) WriteHeader(code int) {
 	if w.ResponseWriter.Header().Get("Content-Encoding") != "" {
 		return
 	}
+	// The content-length after compression is unknown
+	w.ResponseWriter.Header().Del("Content-Length")
 
 	// Parse the first part of the Content-Type response header.
 	contentType := ""


pkieltyka:
@lclarkmichalek maybe we should set the Content-Length by the actual compressed body?

lclarkmichalek:
The problem is, we haven't seen the body at the point where `WriteHeader` is called. We could read in the entire body, compress, set the header correctly, send the header, and then send the compressed body, but this would require us to store the whole body in memory, which I don't think is acceptable.

Content-Length is an optional header, and it isn't simple to add it to this middleware without making some major tradeoffs. If a user insists on having a Content-Length on their compressed response, then they should probably adapt the middleware for their uses.

pkieltyka:
what is setting Conte-Length in the first place btw?

lclarkmichalek:
My handler is using `http.ServeFile` to serve a binary from disk, which is setting the Content-Length header.

lclarkmichalek:
Thanks 👍

PR #139, opened 2017-01-18
Assign method not allowed handler to sub router

prettyxw:
Hi, I tried to set customized `method not found` handler with `MethodNotAllowed()`, but it's not working. Then I fixed that for you.

By the way, `chi` is really amazing, thanks!


pkieltyka:
@prettyxw thanks for the PR and kind words :) cheers

prettyxw:
❤️

PR #143, opened 2017-01-26
Added CustomRecoverer middleware.

dobegor:
Added a middleware constructor that allows to create a custom panic Recoverer.


pkieltyka:
nice, but Id like to keep chi very minimal, so best to use this middleware in your packages.

PR #154, opened 2017-02-13
Compression middleware should implement http.CloseNotifier

chuyeow:
This change allows `http.Handler`s which expect `http.ResponseWriter` to implement [`http.CloseNotifier`](https://golang.org/pkg/net/http/#CloseNotifier) to work transparently with the compression middleware.

If the underlying writer does not implement http.CloseNotifier, we return a channel that never receives a value. The semantics here is that the client never disconnects before the request is processed, and this is as close to the default behavior as we can get (i.e. when `CloseNotify()` is not even called to check whether the client has disconnected).

(I couldn't find any existing tests for `middleware/compress.go`) so I didn't add any, but please point me to the correct place if I'd missed it.)

### DIFF
--- a/middleware/compress.go
+++ b/middleware/compress.go
@@ -191,6 +191,19 @@ func (w *maybeCompressResponseWriter) Hijack() (net.Conn, *bufio.ReadWriter, err
 	return nil, nil, errors.New("chi/middleware: http.Hijacker is unavailable on the writer")
 }
 
+func (w *maybeCompressResponseWriter) CloseNotify() <-chan bool {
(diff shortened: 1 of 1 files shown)


chuyeow:
> The semantics here is that the client never disconnects before the request is processed, and this is as close to the default behavior as we can get (i.e. when CloseNotify() is not even called to check whether the client has disconnected).

Hmm this explanation is not very clear. Let me try this again:

Returning a channel that never returns is closest to the "default" behavior. `http.Handler`s that do not use the `http.CloseNotifier` interface are effectively saying that they do not care when the client disconnects early, so a channel that never returns models that behavior.

PR #157, opened 2017-02-14
Add dummy CloseNotify for backwards compatibility.

nkovacs:
This makes code that uses middleware.CloseNotify compile on 1.8.

Also clarify in the documentation that the middleware
only does nothing on Go 1.8+.

### DIFF
--- a/middleware/closenotify17.go
+++ b/middleware/closenotify17.go
@@ -11,8 +11,8 @@ import (
 // connection has gone away. It can be used to cancel long operations
 // on the server when the client disconnects before the response is ready.
 //
-// Note: this behaviour is standard in Go 1.8+ so the middleware is no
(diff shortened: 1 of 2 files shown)


pkieltyka:
Why not let the compiler complain and have people remove it? its a noop middleware

nkovacs:
If you don't care about backwards compatibility, you could remove middleware.CloseNotify completely, and automatically patch context on go 1.7 somewhere (e.g. [here](https://github.com/pressly/chi/blob/45e2c82f3a180a3c94dc36207f115621a1b48731/mux.go#L78)?).

But if you want to keep backwards compatibility, you can't remove middleware.CloseNotify.

Having it compile on go 1.7 and fail to compile on go 1.8 is even worse than breaking backwards compatibility, because not even vendoring helps there.

PR #161, opened 2017-02-19
update func names for both Slashes middleware in README

dannyvankooten:
Since all other middleware listed in Chi's README is a direct match with the actual middleware name I figured it would make sense to do the same for the two slashes related middleware. 

Backstory is that I recalled reading about a slashes related middleware in Chi, skimmed through the README and then tried using `middleware.Slashes`, which of course didn't exist. This might save someone else a minute going through the middleware package to find the actual middleware name. 

Great work on Chi by the way. Thank you!


pkieltyka:
Thanks!

PR #173, opened 2017-03-24
Render subpkg with better payload control

pkieltyka:
Rewrite of chi/render pkg, introducing Binder and Renderer interfaces for handling well-defined request/response payloads. See _examples/rest/main.go in this branch for a feel of the API.

See #84

Feedback is welcome!


render/render.go
```diff
@@ -1,203 +1,121 @@
...
-// Content-Type as application/json.
-func JSON(w http.ResponseWriter, r *http.Request, v interface{}) {
-	w.Header().Set("Content-Type", "application/json; charset=utf-8")
-	if status, ok := r.Context().Value(statusCtxKey).(int); ok {
-		w.WriteHeader(status)
+// Executed top-down
+func renderer(w http.ResponseWriter, r *http.Request, v Renderer) error {
+	rv := reflect.ValueOf(v)
+	if rv.Kind() == reflect.Ptr {
+		rv = rv.Elem()
```
c2h5oh: My reflect knowledge is somewhat lacking: why not simply `v = reflect.Indirect(reflect.ValueOf(v))` ?

render/decoder.go
```diff
@@ -0,0 +1,44 @@
...
+	"net/http"
+)
+
+// Decode is a package-level variable set to our default Decoder. We do this
+// because it allows you to set render.Decode to another function with the
+// same function signature, while also utilizing the render.Decoder() function
+// itself. Effectively, allowing you to easily add your own logic to the package
+// defaults. For example, maybe you want to impose a limit on the number of
+// bytes allowed to be read from the request body.
+var Decode = DefaultDecoder
```
c2h5oh: Being able to set a Decoder per Binder (with fallback to DefaultDecoder) would be super useful for handling forms with something like http://www.gorillatoolkit.org/pkg/schema

render/render.go
```diff
@@ -1,203 +1,121 @@
...
-	switch getResponseContentType(r) {
-	case ContentTypeJSON:
-		JSON(w, r, v)
-	case ContentTypeXML:
-		XML(w, r, v)
-	default:
-		JSON(w, r, v)
-	}
+type Binder interface {
+	Bind(r *http.Request) error
```
c2h5oh: See decoder comment

PR #182, opened 2017-04-19
GetContentType ignore charset

stefanprodan:
Splits the Content-Type after `;` to correctly parse headers like `Content-Type: application/json; charset=utf-8`

### DIFF
--- a/render/content_type.go
+++ b/render/content_type.go
@@ -25,7 +25,8 @@ const (
 )
 
 func GetContentType(s string) ContentType {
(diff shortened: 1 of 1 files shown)


pkieltyka:
@stefanprodan thanks - did this work though? I believe you need `parts := ...`

stefanprodan:
I've tested it now, sorry!

stefanprodan:
Closing since it will covered by #185

PR #183, opened 2017-04-23
Return interface instead of struct implementation

djui:
`Mux` satisfies the `Router` interface and returning interfaces is preferred whenever possible to comply to S.O.L.I.D. principles.

### DIFF
--- a/chi.go
+++ b/chi.go
@@ -32,7 +32,7 @@ package chi
 import "net/http"
 
 // NewRouter returns a new Mux object that implements the Router interface.
-func NewRouter() *Mux {
+func NewRouter() Router {
 	return NewMux()
 }
 


djui:
I am aware of the proverb. Here however to me it is a misused, specifically reading the API offered:

```go
NewRouter() *Mux // router.go
NewMux() *Mux // mux.go
```

Firstly, there is no need for the function in the first place, as it is just a proxy. Secondly, it has a confusing name as: an interface with the same name as the file name, but also it does not return what its name suggests: a router.

Maybe this function is kept for backwards-compatibility as https://github.com/pressly/chi/commit/caaa5eebe3be91a41f2dbdd90402cd6374b2043f indicates? You seem to be conscious about the choice given and it would break the current API, so no need to for further arguing.

Post scriptum: Ironically I found `func New() *Mux { // return Router?` in the logs https://github.com/pressly/chi/blame/1123f01030bebd4b81f82c2ea96c9cf4233ad9be/chi.go...

PR #192, opened 2017-05-12
Add an authorization middleware that supports ACL, RBAC access control models based on Casbin.

hsluoyz:
This PR is to solve: https://github.com/pressly/chi/issues/191

### DIFF
--- a/middleware/authz.go
+++ b/middleware/authz.go
@@ -1,0 +1,41 @@
+package middleware
+
+// This plugin is based on Casbin: an authorization library that supports ACL, RBAC, ABAC
+// View source at:
+// https://github.com/hsluoyz/casbin
+
+import (
+	"net/http"
+
+	"github.com/hsluoyz/casbin"
+)
+
+// Authz is a middleware that controls the access to the HTTP service, it is based
+// on Casbin, which supports access control models like ACL, RBAC, ABAC.
+// The plugin determines whether to allow a request based on (user, path, method).
+// user: the authenticated user name.
+// path: the URL for the requested resource.
+// method: one of HTTP methods like GET, POST, PUT, DELETE.
+//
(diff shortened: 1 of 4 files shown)


middleware/authz.go
```diff
@@ -0,0 +1,41 @@
+package middleware
+
+// This plugin is based on Casbin: an authorization library that supports ACL, RBAC, ABAC
+// View source at:
+// https://github.com/hsluoyz/casbin
+
+import (
+	"net/http"
+
+	"github.com/hsluoyz/casbin"
```
VojtechVitek: Since this package imports 3rd party pkg, it can't be part of chi/middleware pkg. We could move it to a separate pkg, though. Like `github.com/pressly/chi/middleware/casbin` or to a different repo `github.com/pressly/middleware/casl`. @pkieltyka thoughts?

pkieltyka: @VojtechVitek is right, we're not able to include 3rd party imports into chi. 

@VojtechVitek cool idea, we can make pressly/middleware for 3rd party middlewares. Question: if we make pressly/middleware/casabin and say pressly/middleware/<someotherthirdparty> can govendor or other things manage import paths only for the packages actually used by a project and not bloating child deps

VojtechVitek: @pkieltyka I believe that each directory in `github.com/pressly/middleware` could have its own `vendor/` deps.

hsluoyz: It seems that we are going to make it a separate package. I made one here: https://github.com/casbin/chi-authz

Maybe you want to put it in your ``github.com/pressly/middleware``, but it seems not been created yet.

pkieltyka: thanks @hsluoyz - I've added your pkg to the list in 898d761524954def9191dc8bea69191442581385 cheers!

PR #196, opened 2017-05-13
Fix http: multiple response.WriteHeader calls in render.JSON

### LINKED ISSUE #193: http: multiple response.WriteHeader calls if called render.Status and json encode failed
Cause the issue in:
https://github.com/pressly/chi/blob/master/render/responder.go#L102

wutaizeng:
Fixes #193


pkieltyka:
thanks for the PR but I've fixed it with: f262230f5155fe761ddca49f01f78d8383bcb145

PR #201, opened 2017-05-31
Add trace handler to the profiler

kowalczykp:
This MR adds a handler to profiler, that enables downloading execution traces f.e. with:
`wget http://host:PORT/debug/pprof/trace?seconds=5`

Since chi relies on std context (>=go1.7), this addition should be safe (trace handler was added somewhere around 1.5 I think, but again this is not an issue here)


pkieltyka:
thanks @kowalczykp 
and thanks @VojtechVitek for reviewing

PR #206, opened 2017-06-15
middleware/logger: use time.Since instead of Sub

dqminh:
Noticed this while skimming through the source code. I did a search and look like this is the only usage.


pkieltyka:
@dqminh thanks for the PR, just wondering why favour Since over Sub?

dqminh:
I think Since is a bit cleaner since you dont need to declare an explicit variable to use it. Under the hood its the same thing though since it calls `Now().Sub(t)`

PR #213, opened 2017-06-26
Graceful example for golang >= 1.8

kanocz:
In golang >= 1.8 we don't need to use `github.com/tylerb/graceful` package but built-in `Shutdown` function. This is a modified example


pkieltyka:
thanks @kanocz - this is awesome! I will try it out and merge shortly

PR #216, opened 2017-06-28
Avoid matching parameters across path segments

### LINKED ISSUE #215: Suffix to variable match
Hello,

When upgrading to chi 3 we noticed a suffix to a variable caused unexpected match.

Having routes for

- /articles/{id}.json
- /articles/{id}/data.json

will route /articles/123/data.json to the earlier handler with id as 123/data.

I've added a test for it, but am not sure where to approach tree.go for a solution.

[suffix-to-variable-test.txt](https://github.com/go-chi/chi/files/1107786/suffix-to-variable-test.txt)

I've also found there is little documentation how the url pattern should work.

niklasj:
Fixes #215

### DIFF
--- a/tree.go
+++ b/tree.go
@@ -388,6 +388,9 @@ func (n *node) findRoute(rctx *Context, path string) *node {
 					if xn.rex.Match([]byte(xsearch[:p])) == false {
 						continue
 					}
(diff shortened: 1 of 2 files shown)


pkieltyka:
thanks for this! LGTM

PR #218, opened 2017-06-29
fix panic when url param is missing in called route

dvic:


dvic:
Force pushed same commit to trigger travis build

pkieltyka:
thanks for this, its solved alternatively fixing another bug with PR #224

PR #223, opened 2017-07-08
Recursive not found fix for trailing slash routes

beornf:
There was an issue using custom not found handlers in auto-generated trailing slash routes. The tests show the expected behavior for recursive mounts.

The not found handler is wrapped in a inline handler for run-time evaluation of the mux not found routine.


pkieltyka:
Cool find. Thanks @beornf for the PR, LGTM

PR #224, opened 2017-07-08
short-circuit and return no matching route for empty param values

pkieltyka:


pkieltyka:
hey @dvic - thanks for the PR on #218. I reviewed it and realized another bug was looming that caused the panic. Because, there should never be a case where a parameter key is present in URLParams, but a matching param value wouldn't be available, so I knew something was wrong there.

I debugged it and found and decided that the router should return a 404 instead of matching the handler with an empty param value, because in the case, it does not correctly match the regexp `\d+` pattern and even if it was just a normal param, it still doesnt make sense to match it. https://github.com/go-chi/chi/blob/7eca4e6882ef07a411ec24e58894a14e4e2ea754/mux_test.go#L1290

dvic:
Nice catch!

PR #229, opened 2017-07-14
Unbreaking the 'FileServer' example

HaraldNordgren:
Not sure what this is supposed to be, but after moving from Chi 2.1 => 3.1, I had problems with the FileServer.

The old one (https://github.com/go-chi/chi/blob/e6033ea75479391a4bce3918fc119cad31e1cb30/mux.go#L305) was matching

```
":*"
```

The new example is matching this, but this is giving me problems:

```
"{}*"
```

Just trying out different thing, in my project at least this works:

```
"*"
```


pkieltyka:
Can you elaborate on "but this is giving me problems"

PR #235, opened 2017-07-24
Fix WrapWriter.ReadFrom() with Tee

VojtechVitek:
@mrbanzai this is a follow-up for #234. Would you mind taking a look?

I believe we should propagate the number of bytes when using Tee as well, if I'm not wrong.


mrbanzai:
Ah, good catch @VojtechVitek !  I completely overlooked that section as I was correcting it for my use case.

PR #237, opened 2017-08-01
add go vet and golint to travis checks

zemirco:
`go vet` and `golint` show several errors.

I could start fixing them but would leave them open if you'd like to keep your comment style. In addition you know best what you code does so I'm not sure if I'm the right person for adding comments.

What do you think?


zemirco:
Here is the go report card https://goreportcard.com/report/github.com/go-chi/chi.

Let's make chi even stronger :muscle:

PR #249, opened 2017-08-23
Introduce chi.Config for very minor config options of the Mux

pkieltyka:
And yet.. another alternative for how to solve https://github.com/go-chi/chi/issues/238 

This introduces `chi.Config` which offers `AutoHeadGet bool` to vary the routing execution. The `chi.Config` can be passed as an optional argument to `chi.NewRouter()`.

I like the simplicity of this solution, which also has no executional overhead versus PR #248, but I'm wary to open a can of worms for having more nuances to support via a Config.

### DIFF
--- a/chi.go
+++ b/chi.go
@@ -32,8 +32,8 @@ package chi
 import "net/http"
 
 // NewRouter returns a new Mux object that implements the Router interface.
-func NewRouter() *Mux {
-	return NewMux()
+func NewRouter(config ...Config) *Mux {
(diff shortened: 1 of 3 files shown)


alehano:
I like simplicity and straightforward way of enabling HEAD requests. It's little bit strange to have config with just one parameter, but you might want to add others in the future.

pkieltyka:
decidedly this is not the right approach.

PR #254, opened 2017-08-28
Add generic middleware logger interface

beornf:
As I prefer to use a custom logger making the logger an interface means I can inject my own for request logging.

This also has the property of being fully backwards compatible with the stdlib logger.


pkieltyka:
please show example usage

beornf:
Overriding the default logger as below works:
`
middleware.DefaultLogger = middleware.RequestLogger(&middleware.DefaultLogFormatter{
    Logger: logrus.New(),
}
`

PR #255, opened 2017-08-30
Context.URLParamInt64 helper

cristaloleg:
I saw 1 unchecked item in #179 regarding `URLParamInt64` and also I needed it in my project.

I did a basic implementation, but there are at least 2 other versions:
1. add an additional parameter as a default value
2. return an error 
when the value doesn't exist or parsing isn't possible.

### DIFF
--- a/context.go
+++ b/context.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"net"
 	"net/http"
+	"strconv"
 	"strings"
 )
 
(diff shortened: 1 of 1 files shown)


pkieltyka:
thanks for the PR but chi's focus is to be as minimal as possible. Please add these helper functions outside of chi. You can use chi.URLParam() and cast from string to whatever value you'd like as you've done

PR #259, opened 2017-09-12
path detection missmatch

hovhannest:
to detect both http://example.com and http://example.com/ as the same path

### DIFF
--- a/tree.go
+++ b/tree.go
@@ -441,7 +441,7 @@ func (n *node) findRoute(rctx *Context, method methodTyp, path string) *node {
 		}
 
 		// did we find it yet?
-		if len(xsearch) == 0 {
+		if len(xsearch) == 0 || xsearch == "/" {
 			if xn.isLeaf() {
 				h, _ := xn.endpoints[method]
 				if h != nil && h.handler != nil {


hovhannest:
Thanks for reply.

Sorry if I have misunderstand.. In Mount you really have smt like
if pattern == "" || pattern[len(pattern)-1] != '/' 
and it understands that r.Mount("/path", ...) and r.Mount("/path/", ...) are the same, but for example in this kind of code

r.Mount("/path", MyRouter())
...

func MyRouter() http.Handler {
	r := chi.NewRouter()

	r.Get("/*", func (w http.ResponseWriter, r *http.Request){
		w.Write([]byte("smt is here"))
	})

	return  r
}

I have 404 "Page not found" for localhost/path/
and for 

r.Mount("/path/", MyRouter())
...

func MyRouter() http.Handler {
	r := chi.NewRouter()

	r.Get("/*", func (w http.ResponseWriter, r *http.Request){
		w.Write([]byte("smt is here"))
	})

	return  r
}

I have 404 "Page not found" for localhost/path

I would be appreciated if you'll tell me what is my mistake if there is everything ok with chi.

Thanks,
Hovhannes Tsakanyan

VojtechVitek:
```go
package main

import (
	"net/http"

	"github.com/go-chi/chi"
)

func main() {
	http.ListenAndServe(":9876", MyRouter())
}

func MyRouter() http.Handler {
	r := chi.NewRouter()

	r.Get("/*", func(w http.ResponseWriter, r *http.Request) {
		w.Write([]byte("smt is here"))
	})

	return r
}

```

```
$ curl localhost:9876/path
smt is here
$ curl localhost:9876/path/
smt is here
$ curl localhost:9876/path/hey
smt is here
```

VojtechVitek:
@hovhannest actually, I think I found your use-case with the sub-router:
https://github.com/go-chi/chi/issues/260

Let's follow up in this issue. Thanks!

Issue #264, opened 2017-09-25
Accept LOCK and UNLOCK methods?


tboerger:
I would like to build a terraform remote store based on chi, but for that I need 2 api endpoints with the above methods. Should I create a pr for that?

tboerger:
For reference, I want to implement this api: https://www.terraform.io/docs/backends/types/http.html

PR #265, opened 2017-09-25
Implemented LOCK and UNLOCK handlers

tboerger:
Fixes https://github.com/go-chi/chi/issues/264

### DIFF
--- a/chi.go
+++ b/chi.go
@@ -78,6 +78,8 @@ type Router interface {
 	Post(pattern string, h http.HandlerFunc)
 	Put(pattern string, h http.HandlerFunc)
 	Trace(pattern string, h http.HandlerFunc)
+	Lock(pattern string, h http.HandlerFunc)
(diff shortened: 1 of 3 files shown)


tboerger:
I have already started to use these changes within my tool and it's working good so far.

pkieltyka:
Hi, I’m on vacation for 2 weeks and won’t be able to help until I return. Just FYI, I have a different solution in mind - specifically in the core I do not want to define unstandard methods like this, however I have a solution that will allow you to add your own methods at runtime. For now just use your fork

tboerger:
Yeah currently I'm using my vendored fork for that. These methods are part of a standard, it's part of the Webdav specification, for reference: http://www.webdav.org/specs/rfc2518.html

pkieltyka:
solved in https://github.com/go-chi/chi/commit/d806569952e5c14126a155fe66cf102139d11979

PR #266, opened 2017-10-04
[WIP] Coverage

zet4:
Was bored so decided to try and increase coverage of the tests to 100%...
Work in progress, opening this to get comments.

### DIFF
--- a/context.go
+++ b/context.go
@@ -99,7 +99,10 @@ func (x *Context) RoutePattern() string {
 // RouteContext returns chi's routing Context object from a
 // http.Request Context.
 func RouteContext(ctx context.Context) *Context {
-	return ctx.Value(RouteCtxKey).(*Context)
+	if ctx, ok := ctx.Value(RouteCtxKey).(*Context); ok {
+		return ctx
+	}
+	return nil
 }
 
 // URLParam returns the url parameter from a http.Request object.
--- a/context_test.go
+++ b/context_test.go
@@ -1,0 +1,79 @@
+package chi
+
+import (
+	"context"
+	"net/http"
+	"testing"
+)
+
+func TestContextURLParam(t *testing.T) {
+	rctx := NewRouteContext()
+	rctx.URLParams.Add("id", "123")
+
+	t.Run("ID Exists", func(t *testing.T) {
+		if rctx.URLParam("id") != "123" {
+			t.Fail()
+		}
+	})
+	t.Run("NotID Doesn't Exist", func(t *testing.T) {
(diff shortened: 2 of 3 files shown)


context.go
```diff
@@ -99,7 +99,10 @@ func (x *Context) RoutePattern() string {
 // RouteContext returns chi's routing Context object from a
 // http.Request Context.
 func RouteContext(ctx context.Context) *Context {
-	return ctx.Value(RouteCtxKey).(*Context)
+	if ctx, ok := ctx.Value(RouteCtxKey).(*Context); ok {
+		return ctx
+	}
+	return nil
```
VojtechVitek: How is this extra condition helping the codebase? The one liner does the very same thing, if I'm not wrong.

zet4: Nope, without the explicit check the type cast will cause a panic, found out about this while writing test for it.

VojtechVitek: got it .. you can do this:
```go
ctx, _ := ctx.Value(RouteCtxKey).(*Context)
return ctx
```

jawnsy: (Random drive-by Internet commenter here)

While the two options are equivalent, I prefer the explicit `ok` check. It makes it easier to understand the two different possible cases, as well as to add logging or other handling later. The cost is two extra lines of code. Presumably these get compiled the same, so there's no runtime performance impact.

VojtechVitek: I don't think it gets compiled into the same machine code, since there is an extra condition when dealing with `ok`, and thus an extra jump in assembly, if I'm not wrong.

No big difference, though. I'm just used to omitting `ok` condition, if I'm fine with the zero value (both when type casting or when accessing non-existing map key).

jawnsy: Huh, interesting, you are correct, TIL. There's an extra JNE, MOVQ, and RET instruction: https://gist.github.com/jawnsy/613405bd42eb7f8ffaa6bdde4c372048 - I'm surprised that the Go compiler doesn't optimize it thanks to escape analysis. In practice I don't think it matters (just a style preference), but good to know nonetheless.

For reference, I built the files using:

`go tool compile -S assert-check.go >assert-check.s`

And my `go version` output is (using golang package from Debian unstable): `go version go1.9.1 linux/amd64`

VojtechVitek: I don't think escape analysis can eliminate the jump. No one knows the type of `ctx.Value(RouteCtxKey)` until the condition is executed in the runtime.

VojtechVitek:
@zet4 ping, this is still marked as [WIP]. Are you planning to finish this PR, or should we close?

pkieltyka:
outdated, closing

PR #272, opened 2017-10-12
New example

ffest:
Routes walk example added

### DIFF
--- a/_examples/router-walk/main.go
+++ b/_examples/router-walk/main.go
@@ -24,18 +24,13 @@ func main() {
 
 	r.Put("/ping", Ping)
 
-	methodCount := make(map[string]int)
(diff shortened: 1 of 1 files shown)


_examples/router-walk/main.go
```diff
@@ -0,0 +1,45 @@
...
+			w.Write([]byte("right road"))
+		})
+	})
+
+	r.Put("/ping", Ping)
+
+	methodCount := make(map[string]int)
+	walkFunc := func(method string, route string, handler http.Handler, middlewares ...func(http.Handler) http.Handler) error {
+		methodCount[method]++
+		return nil
```
VojtechVitek: You might wanna print the routes in the following format instead?

```
GET /
GET /ping
```

ffest: Yes, looks good!

PR #275, opened 2017-10-25
Update README to mention phi

fate-lovely:
Hi, first, thank you guys for such elegant work :D. I really enjoy using chi to build http service, 
but I prefer fasthttp to net/http, so I port chi to fasthttp, which
named [phi](https://github.com/fate-lovely/phi).

### DIFF
--- a/README.md
+++ b/README.md
@@ -109,7 +109,7 @@ func main() {
 
     // Regexp url parameters:
     r.Get("/{articleSlug:[a-z-]+}", getArticleBySlug)                // GET /articles/home-is-toronto
-    
+
     // Subrouters:
     r.Route("/{articleID}", func(r chi.Router) {
       r.Use(ArticleCtx)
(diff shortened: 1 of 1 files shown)


jawnsy:
@fate-lovely Would it be possible to add fasthttp support to chi somehow? Otherwise, it might be difficult keeping your fork up-to-date with some of the latest changes in chi. This seems, more generally, to be one of the problems with fasthttp: most libraries and frameworks are written to coexist with the standard net/http, thus severely limiting the ecosystem compatible with fasthttp.

fate-lovely:
@jawnsy There seems to be no easy way to add fasthttp support to chi except replace every `http.Handler ` interface to fasthttp request handler, cause go doesn't have generics. I will keep watching chi and update phi, it seems the only way to go😄. I really enjoy fasthttp, not for its performance, but its elegant design. A single `ctx *fasthttp.RequestCtx` is much better and flexible than `(w http.ResponseWriter, r *http.Request)`.

pkieltyka:
thanks. Can you list the project in the table of https://github.com/go-chi/chi#auxiliary-middlewares--packages

pkieltyka:
btw, its not possible to add fasthttp to chi directly, it would create a dependency and major change which goes against the design of the project

fate-lovely:
@pkieltyka It's been moved to the table, please check.

PR #279, opened 2017-11-14
Fix missing profile subpages

### LINKED ISSUE #278: pprof endpoints should not be hardcoded
I'm using go 1.9, and `/debug/pprof/` has a link to `/debug/pprof/mutex`, which doesn't work because the profiler middleware only registers `block`, `heap`, `goroutine` and `threadcreate`.
Instead of hardcoding these individually, you could just do this:

```
r.HandleFunc("/pprof/*", pprof.Index)
```

just like how http/pprof only registers pprof.Index to handle all of the subpages.

nkovacs:
Let pprof.Index handle routing for the different profiles,
just like it does with net/http's router.

Fixes  #278

### DIFF
--- a/middleware/profiler.go
+++ b/middleware/profiler.go
@@ -29,15 +29,11 @@ func Profiler() http.Handler {
 		http.Redirect(w, r, r.RequestURI+"/", 301)
 	})
 
-	r.HandleFunc("/pprof/", pprof.Index)
+	r.HandleFunc("/pprof/*", pprof.Index)
 	r.HandleFunc("/pprof/cmdline", pprof.Cmdline)
 	r.HandleFunc("/pprof/profile", pprof.Profile)
 	r.HandleFunc("/pprof/symbol", pprof.Symbol)
 	r.HandleFunc("/pprof/trace", pprof.Trace)
-	r.Handle("/pprof/block", pprof.Handler("block"))
-	r.Handle("/pprof/heap", pprof.Handler("heap"))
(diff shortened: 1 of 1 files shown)


middleware/profiler.go
```diff
@@ -29,15 +29,11 @@ func Profiler() http.Handler {
...
-	r.HandleFunc("/pprof/", pprof.Index)
+	r.HandleFunc("/pprof/*", pprof.Index)
 	r.HandleFunc("/pprof/cmdline", pprof.Cmdline)
 	r.HandleFunc("/pprof/profile", pprof.Profile)
 	r.HandleFunc("/pprof/symbol", pprof.Symbol)
 	r.HandleFunc("/pprof/trace", pprof.Trace)
-	r.Handle("/pprof/block", pprof.Handler("block"))
-	r.Handle("/pprof/heap", pprof.Handler("heap"))
-	r.Handle("/pprof/goroutine", pprof.Handler("goroutine"))
-	r.Handle("/pprof/threadcreate", pprof.Handler("threadcreate"))
```
VojtechVitek: This change would hide all these endpoints from [docgen](https://github.com/go-chi/docgen). Can we instead be more explicit and add the missing handlers?

nkovacs: The problem is that these are dynamic, you can add a new one using [pprof.NewProfile](https://golang.org/pkg/runtime/pprof/#NewProfile).

Issue #260, opened 2017-09-13
Mount() doesn't match root path of the sub-router


VojtechVitek:
Given
```go
package main

import (
	"net/http"

	"github.com/go-chi/chi"
)

func main() {
	r := chi.NewRouter()
	r.Mount("/path", MyRouter())

	http.ListenAndServe(":9876", r)
}

func MyRouter() http.Handler {
	r := chi.NewRouter()

	r.Get("/*", func(w http.ResponseWriter, r *http.Request) {
		w.Write([]byte("works..\n"))
	})

	return r
}
```

How come `localhost:9876/path/` is not matched and returns 404?
```
$ curl localhost:9876/path
works..

$ curl localhost:9876/path/
404 page not found

$ curl localhost:9876/path/hey
works..
```

Brought up in #259 by @hovhannest.

olihey:
Any fix for this?
I just went into the exact same issue and the only way I could fix it was to use
```r.Mount("/path/", MyRouter())```
but then `localhost:9876/path` fails, of course 😢

hovhannest:
I've changed the "thee.go" file line 444 from
if len(xsearch) == 0 {
to
if len(xsearch) == 0  || xsearch == "/" {

and I'm using r.Mount("/path", MyRouter())

I've suggested this change, but my pull request was denied. May be I'm doing smt. wrong, but it's working for me.

inancgumus:
Why is that your request has been denied?

pkieltyka:
I’m away on vacation but I will revisit this issue when I return

VojtechVitek:
@inancgumus you can see yourself in #259. Check Travis to see why.

pkieltyka:
Fixed in https://github.com/go-chi/chi/pull/281

lrstanley:
Any idea on user pushback for merging #281 into master? I've had to use the **fix-260** branch for over 5 projects at this time. Haven't ran into any issues, and I personally find that the implementation makes sense. When mounting, one would think that it's treated the same way that the primary/first `/` is handled.

pkieltyka:
I merged the #281 PR which fixes this. Give master a try and let us know

PR #281, opened 2017-11-25
Route trailing slash on mounted sub-routers

pkieltyka:
This PR addresses issue #260 

I'm not convinced yet this is the correct behaviour, but it is convenient.

Waiting to hear feedback from the community before we merge it.

### DIFF
--- a/middleware/strip_test.go
+++ b/middleware/strip_test.go
@@ -65,6 +65,9 @@ func TestStripSlashesInRoute(t *testing.T) {
 
 	r.Route("/accounts/{accountID}", func(r chi.Router) {
 		r.Use(StripSlashes)
+		r.Get("/", func(w http.ResponseWriter, r *http.Request) {
+			w.Write([]byte("accounts index"))
+		})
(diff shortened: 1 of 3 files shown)


middleware/strip_test.go
```diff
@@ -80,6 +83,12 @@ func TestStripSlashesInRoute(t *testing.T) {
 	if _, resp := testRequest(t, ts, "GET", "/hi/", nil); resp != "nothing here" {
 		t.Fatalf(resp)
 	}
+	if _, resp := testRequest(t, ts, "GET", "/accounts/admin", nil); resp != "accounts index" {
+		t.Fatalf(resp)
+	}
+	if _, resp := testRequest(t, ts, "GET", "/accounts/admin/", nil); resp != "accounts index" {
```
VojtechVitek: I think `GET /accounts/admin` should work for sure, being handled by `GET /` in the subrouter.

Do we need to support `/accounts/admin/` too? I think it doesn't really hurt, and makes sense ... since `""` and `"/"` from the sub-router perspective should be the same.

PR #283, opened 2017-11-27
Make Content-Type lookup in AllowContentType O(1)

vgalu:


pkieltyka:
I wonder if this is actually faster? I heard a tight for loop in cases is faster than a map look up. I know it doesn't make sense, but I've never benchmarked it either

vgalu:
@pkieltyka indeed, I should have run some benchmarks first, I'll wear the pointy hat for a while. Please disregard the noise. The short array (up to 10 elements or so) iteration is 60x faster than the map lookup.

PR #285, opened 2017-12-07
Add pattern documentation

lpar:
Someone turned up on the Gophers Slack and asked what the allowed syntax was for Chi's URL parameter placeholders. I couldn't find any documentation of that, so I've looked at the code and done some testing and tried to write up what seems to be supported at the moment.

(Since there's no code in this request I didn't bother with a topic branch.)


pkieltyka:
This is awesome! Thank you very much

PR #286, opened 2017-12-11
Added middleware to check matching charsets

csucu:
Added a middleware that generates a handler which can be used to check the content charsets on a request against a list of acceptable values, a 415 Unsupported Media Type response is written if none of the charsets match.


pkieltyka:
Thanks @csucu this is great.

PR #288, opened 2017-12-21
Add new default type image/svg+xml

ehmo:
`image/svg+xml` should be considered as a default type especially because it can be nicely compressed.

### DIFF
--- a/middleware/compress.go
+++ b/middleware/compress.go
@@ -29,7 +29,7 @@ var defaultContentTypes = map[string]struct{}{
 	"application/json":         struct{}{},
 	"application/atom+xml":     struct{}{},
 	"application/rss+xml":      struct{}{},
-        "image/svg+xml":            struct{}{},
+	"image/svg+xml":            struct{}{},
 }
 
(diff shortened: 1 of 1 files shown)


middleware/compress.go
```diff
@@ -29,6 +29,7 @@ var defaultContentTypes = map[string]struct{}{
 	"application/json":         struct{}{},
 	"application/atom+xml":     struct{}{},
 	"application/rss+xml":      struct{}{},
+        "image/svg+xml":            struct{}{},
```
pkieltyka: thx. can you fix the indentation here? its breaking the CI due to formatting

ehmo: It's very surprising that there is any kind of issue as it ran through `go fmt` before pushing. Will check it out.

VojtechVitek:
@ehmo needs rebase.

ehmo:
Considering that this was committed almost a year ago, I would hope that you guys would merge it by now? Feel free to close this and add it directly to the master. It's a single line, don't care where it came from.

pkieltyka:
roger that

PR #289, opened 2017-12-29
Added secure middleware to README

srikrsna:
Secure middleware provides common security headers and is compatible with Chi. [Godoc](https://godoc.org/github.com/srikrsna/security-headers#example-package--Chi) has a Chi integration example.


pkieltyka:
hey @srikrsna - this is neat. One suggestion though is to rename the package path from `github.com/srikrsna/security-headers` to `github.com/srikrsna/secure`. The import path should reflect the package name.

srikrsna:
I'll create a new repo and update the README accordingly @pkieltyka

PR #298, opened 2018-02-02
 mux: With new instance copy custom handlers

### LINKED ISSUE #297: Mux.With does not copy custom handlers
`Mux.Mount` would apply custom error handlers from parent to child but when there is `m.With(...).Mount()` the handlers are wiped away [1].

[1] https://github.com/go-chi/chi/blob/master/mux.go#L236

mmatczuk:
Fixes #297


mmatczuk:
Hello @pkieltyka do you have any thoughts on this PR?

pkieltyka:
thanks for the PR @mmatczuk

PR #300, opened 2018-02-02
bug fix: wroteHeader should be set when Flush is done

cyx:
WrapResponseWriter currently does `WriteHeader` when `Write` is called and assumed the operation is idempotent.

This introduces warnings if you call `Flush` first and then `Write` specifically of the form:

```
http: multiple response.WriteHeader calls
```

The go stdlib `response.Flush` implementation does a `wroteHeader=true` assignment when Flush is done, and then writes a 200. This change aligns with the behavior of the stdlib Flush / Write /  WriteHeader dance.

To see this in action, I've written this simple gist: https://gist.github.com/cyx/1e5abe141ee77e4c5b2087133360b0ea

### DIFF
--- a/middleware/wrap_writer.go
+++ b/middleware/wrap_writer.go
@@ -83,6 +83,8 @@ type flushWriter struct {
 }
 
 func (f *flushWriter) Flush() {
+	f.wroteHeader = true
+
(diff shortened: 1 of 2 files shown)


VojtechVitek:
Good catch. Thanks for the contribution!

PR #303, opened 2018-03-09
Fix URLFormat middleware for nested router

karupanerura:
`URLFormat` middleware is broken in nested router.

e.g.)
```go
r := chi.NewRouter()
r.Get("/", http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
	w.Header().Set("Content-Type", "text/plain")
	w.WriteHeader(http.StatusOK)
	w.Write([]byte("root"))
}))
r.Route("/bar", func(r chi.Router) {
	r.Use(URLFormat)
	r.Get("/baz", http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
		ext, ok := r.Context().Value(URLFormatCtxKey).(string)
		w.Header().Set("Content-Type", "text/plain")
		w.WriteHeader(http.StatusOK)
		if ok {
			w.Write([]byte("/bar/baz." + ext))
		} else {
			w.Write([]byte("/bar/baz"))
		}
	}))
})
```

It's not works because `(*chi.Context).RoutePath` should be a nested child path, but the original implementation makes it a full path.

And the original implementation sets `URLFormatCtxKey` value always, but it's not good for debugging.
So I think better that should not set `URLFormatCtxKey` value if it failed to get format.

Please consider.
Thank you.

### DIFF
--- a/middleware/url_format.go
+++ b/middleware/url_format.go
@@ -48,23 +48,22 @@ func URLFormat(next http.Handler) http.Handler {
 	fn := func(w http.ResponseWriter, r *http.Request) {
 		ctx := r.Context()
 
-		var format string
-		path := r.URL.Path
-
-		if strings.Index(path, ".") > 0 {
-			base := strings.LastIndex(path, "/")
-			idx := strings.Index(path[base:], ".")
-
-			if idx > 0 {
-				idx += base
-				format = path[idx+1:]
-
-				rctx := chi.RouteContext(r.Context())
-				rctx.RoutePath = path[:idx]
+		rctx := chi.RouteContext(ctx)
+		routePath := rctx.RoutePath
+		if routePath == "" {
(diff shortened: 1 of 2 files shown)


karupanerura:
@pkieltyka how do you think?

pkieltyka:
@karupanerura thanks for the PR. I'm tight for a deadline at the moment and I'll review the PR when I can. If others in the community can check it out that'd be appreciated too.

karupanerura:
OK. I'm waiting for you. Thank you.

karupanerura:
hey

karupanerura:
@pkieltyka how is the status? I want to stop using the patched version.

karupanerura:
@pkieltyka ping

karupanerura:
give up

PR #318, opened 2018-05-02
Make heartbeat payload configurable

dahankzter:
We would like a way to configure the returned payload so we can return perhaps "up", "down" or even "warn".

The suggested solution simply allows for a closure to be passed in alongside the endpoint string.
This is quite powerful and allows for all kinds of checks to be performed without having a lot of infrastructure wiring in place.


middleware/heartbeat.go
```diff
@@ -10,12 +10,20 @@ import (
...
 // to place this above ACL middlewares as well.
 func Heartbeat(endpoint string) func(http.Handler) http.Handler {
+	return HeartbeatFunc(endpoint, func() []byte {
+		return []byte(".")
+	})
+}
+
+// HeartbeatFunc is similar to Heartbeat but dynamically allows for a dynamic
+// payload to be returned
+func HeartbeatFunc(endpoint string, payload func() []byte) func(http.Handler) http.Handler {
```
VojtechVitek: why `func() []byte`?

dahankzter: The caller can create any closure that simply returns the payload that they want. This can be dynamic or static as the user sees fit.

We could modify a local copy for our need but if possible evolving the shared repository of middlewares is good imho.

pkieltyka:
yes, I think the answer is to copy/paste the middleware and build it to suit your needs

dahankzter:
Ok, fine. I find it very strange however. It is a very small patch that covers a lot of use cases.

PR #326, opened 2018-06-26
Add support for Brotli compression

polyfloyd:
This patch adds support for Google's Brotli compression format currently [supported](https://caniuse.com/#search=brotli) by Chrome, Firefox and IE edge.

### DIFF
--- a/middleware/compress.go
+++ b/middleware/compress.go
@@ -9,6 +9,8 @@ import (
 	"net"
 	"net/http"
 	"strings"
+
+	brotli_enc "gopkg.in/kothar/brotli-go.v0/enc"
 )
 
 type encoding int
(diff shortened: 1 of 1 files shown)


polyfloyd:
Ah yes, Runtime initialization seems like a good middle way to implement support Brotli until we have a native Go encoder.

polyfloyd:
Alright, I've added a new patch to support dynamic addition of encoders. Let me know what you think :)

pkieltyka:
nice work @polyfloyd — could use a review / small iteration but a nice addition. Do you have a separate subpkg somewhere (in another repo) with the Brotli encoder that uses this interface too? if not, we should make one once this interface is settled / merged

VojtechVitek:
Hi, I'm back from vacation. I'll try to check out & test this PR in the upcoming week or two.

illenial:
When this is going to be mergedm

nicksrandall:
Is this PR waiting on anything before it can be merged? I would like to use this feature. Thank you!

nicksrandall:
One question I have is if we plan to delete the "compress18.go" file or not with this change

polyfloyd:
I do not believe it is waiting for me! @VojtechVitek?

pkieltyka:
LGTM

VojtechVitek:
Sorry for delaying this. This is very good work! Thanks @polyfloyd

🎉 LGTM too

VojtechVitek:
I will test in production next week and report back.

PR #337, opened 2018-08-07
Don't let Mux.Route(...) stomp on already handled routes

### LINKED ISSUE #336: Cannot combine handlers with a matching Route
```go
r := NewRouter()

// A handler that does not use middleware
r.Post("/parent", postHandler)

r.Group(func(r Router) {
	r.Use(mw)

	r.Route("/parent", func(r Router) {
		r.Get("/", getHandler)
		// ... other routes at or under `/parent` that use the middleware
	})
})
```

Requests to `GET /parent` hit the middleware and handler, as expected.

Requests to `POST /parent` hit the middleware, and not the handler (`405 Method Not Allowed` is returned). Neither of those things are expected.

Simplifying to this produces the same result (`GET` handler works, `POST` handler does not):

```go
r := NewRouter()
r.Post("/parent", postHandler)

r.Route("/parent", func(r Router) {
	r.Get("/", getHandler)
	// ...
})
```

Here is a test demonstrating the issue:

```go

func TestRouteAndStandalone(t *testing.T) {
	handlerCalled := false
	handler := func(w http.ResponseWriter, r *http.Request) {
		// t.Logf("Called %s handler", r.Method)
		handlerCalled = true
	}

	r := NewRouter()
	r.Post("/parent", handler)

	r.Route("/parent", func(r Router) {
		r.Get("/", handler)
	})

	ts := httptest.NewServer(r)
	defer ts.Close()

	resp, _ := testRequest(t, ts, http.MethodPost, "/parent", nil)
	if resp.StatusCode != http.StatusOK {
		t.Errorf("Failed to reach POST handler: %s", resp.Status)
	}

	if !handlerCalled {
		t.Error("Expected to reach POST handler, but did not")
	}

	handlerCalled = false

	resp, _ = testRequest(t, ts, http.MethodGet, "/parent", nil)
	if resp.StatusCode != http.StatusOK {
		t.Errorf("Failed to reach GET handler: %s", resp.Status)
	}

	if !handlerCalled {
		t.Error("Expected to reach GET handler, but did not")
	}
}
```

The output is:

```text
$ go test -v -run TestRouteAndStandalone
=== RUN   TestRouteAndStandalone
--- FAIL: TestRouteAndStandalone (0.00s)
	mux_test.go:1311: Failed to reach POST handler: 405 Method Not Allowed
	mux_test.go:1315: Expected to reach POST handler, but did not
FAIL
exit status 1
FAIL	github.com/go-chi/chi	0.015s
```

This is true of v3.3.2, and master @ 0c5e7ab.

dcormier:
Fixes #336

### DIFF
--- a/tree.go
+++ b/tree.go
@@ -345,7 +345,13 @@ func (n *node) setEndpoint(method methodTyp, handler http.Handler, pattern strin
 		h.pattern = pattern
 		h.paramKeys = paramKeys
 		for _, m := range methodMap {
-			h := n.endpoints.Value(m)
+			h, ok := n.endpoints[m]
+			if ok {
+				// Don't stomp on already mapped endpoints
+				continue
+			}
+
+			h = n.endpoints.Value(m)
 			h.handler = handler
 			h.pattern = pattern
 			h.paramKeys = paramKeys
--- a/mux_test.go
+++ b/mux_test.go
@@ -1289,6 +1289,44 @@ func TestNestedGroups(t *testing.T) {
 	}
 }
 
+func TestRouteAndStandalone(t *testing.T) {
+	handlerCalled := false
+	handler := func(w http.ResponseWriter, r *http.Request) {
+		// t.Logf("Called %s handler", r.Method)
+		handlerCalled = true
+	}
+
+	r := NewRouter()
+	r.Post("/parent", handler)
+
+	r.Route("/parent", func(r Router) {
+		r.Get("/", handler)
+	})
+
+	ts := httptest.NewServer(r)
+	defer ts.Close()
+
+	resp, _ := testRequest(t, ts, http.MethodPost, "/parent", nil)
+	if resp.StatusCode != http.StatusOK {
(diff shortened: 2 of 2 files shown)


mcastilho:
This is good...

tree.go
```diff
@@ -345,7 +345,13 @@ func (n *node) setEndpoint(method methodTyp, handler http.Handler, pattern strin
 		h.pattern = pattern
 		h.paramKeys = paramKeys
 		for _, m := range methodMap {
-			h := n.endpoints.Value(m)
+			h, ok := n.endpoints[m]
+			if ok {
+				// Don't stomp on already mapped endpoints
+				continue
```
thedevsaddam: It should panic, instead of silently skipping

thedevsaddam: Isn't is better to panic, instead of silently skipping?

thedevsaddam: Isn't it better to panic, instead of silently skipping?

pkieltyka:
thanks for the PR, however, the intended design of chi is to allow overriding routes

PR #348, opened 2018-09-06
Typo fix in the README.md

BoRuDar:


AWulkan:
Trie is correct though, and a lot of people call it that.

> In computer science, a radix tree (also **radix trie** or compact prefix tree) is a data structure that represents a space-optimized trie (prefix tree) in which each node that is the only child is merged with its parent.

https://en.wikipedia.org/wiki/Radix_tree

PR #351, opened 2018-10-14
Fix leftover Goji reference

pdedmon:
I found a leftover Goji reference in the real IP middleware docs.


pdedmon:
I also fixed the golint installation.

PR #353, opened 2018-10-19
Exported HTTP method names to avoid manual entry

rodney-b:
Added exported http method names for use when calling functions that require manually entering the method name such as Method and MethodFunc to avoid syntax errors plus it defines which methods are supported. Also avoid having devs create their own method constants.

Used all caps for the names to avoid conflicts with the existing exported functions with the same names.

### DIFF
--- a/tree.go
+++ b/tree.go
@@ -36,14 +36,14 @@ var mALL = mCONNECT | mDELETE | mGET | mHEAD |
 // that require a litteral method name such as Method and MethodFunc
 const (
 	CONNECT = "CONNECT"
-	DELETE = "DELETE"
-	GET = "GET"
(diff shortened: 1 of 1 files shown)


rodney-b:
Agreed. Good to know about those.

pkieltyka:
just wondering, what's new or improved that this PR provides..?

pkieltyka:
hey guys, small oversight.. we've broken support for Go 1.7 and Go 1.8 with this PR -- see https://travis-ci.org/go-chi/chi/builds/445197603

pkieltyka:
false alarm, the issue was actually with golint pkg on older versions. all good now

PR #358, opened 2018-11-14
url params keys and values lenght should be equal

speier:
checking length of both keys and values to avoid `runtime error: index out of range` when there is no value for a key.

example to reproduce the issue:
```
r.Get("/foo-{suffix:[a-z]{2,3}}.json", func(w http.ResponseWriter, r *http.Request) {
  fmt.Println(chi.URLParam(r, "suffix"))
})
```

```
GET /foo-.json # runtime error: index out of range
```


pkieltyka:

PR #363, opened 2018-11-26
#362 NotFound instead of panic

### LINKED ISSUE #362: Chi should not panic when no routes
Hello!

We use a generic service skeleton based on chi in our company. Sometimes we needn't handle http-request and our skeleton creates chi.Mux with no roures.

If someone try to ping our service, it fails with panic because of [this if section](https://github.com/go-chi/chi/blob/master/mux.go#L62-L64). 

I think Chi should return 404 error instead of panic.

makhov:
Fixes #362


makhov:
I have no idea why Travis has failed the build. Previous PR also failed with the same error.

pkieltyka:
fixed in #378

PR #364, opened 2018-11-26
Bump module version to v3 (again) now that support is backported

SamWhited:
Hello,

Discussion in #302 concluded by removing support for Go Modules (without the "incompatible" workaround) because the `/v3` suffix was only supported by users of vgo. However, since then limited support for importing using the version suffix has been backported to older versions of Go (including 1.9.7+ and 1.10.3+). For more information see here: https://github.com/golang/go/wiki/Modules#releasing-modules-v2-or-higher

I would like to request that the `/v3` suffix be added back to the module and the import path.
This does require that support for Go 1.7 and 1.8 be dropped, but they haven't been supported for several versions anyways.

This will also (EDIT: this should have read **NOT break**) break existing import paths, which was another reason it was reverted previously, however, since it's an easy update I am hoping you will reconsider now that it is supported,

Thank you for your consideration.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,14 +1,15 @@
 language: go
 
 go:
-  - 1.7.x
-  - 1.8.x
   - 1.9.x
   - 1.10.x
   - 1.11.x
 
 install:
-  - go get -u golang.org/x/tools/cmd/goimports
+  - GO111MODULE=off go get -u golang.org/x/tools/cmd/goimports
+
+env:
+  - GO111MODULE=on
 
 script:
   - go get -d -t ./...
--- a/README.md
+++ b/README.md
@@ -52,7 +52,7 @@ package main
 
 import (
 	"net/http"
-	"github.com/go-chi/chi"
+	"github.com/go-chi/chi/v3"
 )
 
 func main() {
@@ -77,8 +77,8 @@ of chi and serve as a good form of documentation.
 import (
   //...
   "context"
-  "github.com/go-chi/chi"
-  "github.com/go-chi/chi/middleware"
+  "github.com/go-chi/chi/v3"
+  "github.com/go-chi/chi/v3/middleware"
 )
 
 func main() {
--- a/_examples/custom-handler/main.go
+++ b/_examples/custom-handler/main.go
@@ -4,7 +4,7 @@ import (
 	"errors"
 	"net/http"
 
-	"github.com/go-chi/chi"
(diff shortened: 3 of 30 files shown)


VojtechVitek:
@SamWhited You are assuming that chi will support Go 1.9.7+ and Go 1.10.3+ only from now on .. and that we'll silently drop support for 1.7, 1.8, 1.9.0, 1.10.0. We need community consensus on this -- please create an issue for this. This PR is indeed breaking backward compatibility.

RE: Import paths .. thanks for the explanation. So even if we don't add the `/v3` suffix, the import paths will work just fine? So why add the suffix in the first place, should this PR avoid adding it then?

SamWhited:
> You are assuming that chi will support Go 1.9.7+ and Go 1.10.3+ only from now on .. and that we'll silently drop support for 1.7, 1.8, 1.9.0, 1.10.0. We need community consensus on this -- please create an issue for this. This PR is indeed breaking backward compatibility.

I had assumed those had just been left in the test matrix but weren't actually supported given that the Go team only supports the last two versions of the package. I'll open an issue.

> RE: Import paths .. thanks for the explanation. So even if we don't add the /v3 suffix, the import paths will work just fine? So why add the suffix in the first place, should this PR avoid adding it then?

This is for backwards compatibility, it works this way when using versions that don't have full module support. In go 1.9.7 we know where the library is (because we're looking it up in the GOPATH), and we can't import multiple versions of it without renaming the imports anyways, so it effectively just ignores the version suffix (if it's not an actual package that was named v3 or whatever).

With Go 1.11+ however, it doesn't use GOPATH so it uses the version suffix to decide which package to download.

SamWhited:
> We don't want to ever break import paths of chi v3.x (by adding /v3 suffix) and break our users.
Thus, we'll stay on +incompatible flag, until v4.x is released (not planned at this point).

Hello,

Chi v4 was released, but it still doesn't have module support. Does this mean modules and the proper import path won't be supported until v5 for compatibility?

(also asked on #378; sorry, I wasn't sure where the appropriate place to ask was)

pkieltyka:
@SamWhited yea, it would be chi v5 when we support go modules. I don't see this happening for another 6+ months until go 1.13 is available, in hope the go modules is standard and robust for packages. We had hopes to use go modules now, but ran into a bunch of complexity that didn't make sense to put onto chi which literally has zero dependencies to run, lol.

PR #367, opened 2018-11-28
Middleware method RequestID() modified to support X-Request-ID from HTTP header

harikb:
Following middleware RequestIDFromHeader uses the value of "X-Request-ID" header. I am looking at ways to make it a bit more generic and make the header value configurable. Wikipedia mentions X-Request-Id as well X-Correlation-ID

### DIFF
--- a/middleware/request_id.go
+++ b/middleware/request_id.go
@@ -62,9 +62,13 @@ func init() {
 // counter.
 func RequestID(next http.Handler) http.Handler {
 	fn := func(w http.ResponseWriter, r *http.Request) {
-		myid := atomic.AddUint64(&reqid, 1)
 		ctx := r.Context()
-		ctx = context.WithValue(ctx, RequestIDKey, fmt.Sprintf("%s-%06d", prefix, myid))
+		requestID := r.Header.Get("X-Request-Id")
+		if requestID == "" {
+			myid := atomic.AddUint64(&reqid, 1)
+			requestID = fmt.Sprintf("%s-%06d", prefix, myid)
+		}
+		ctx = context.WithValue(ctx, RequestIDKey, requestID)
 		next.ServeHTTP(w, r.WithContext(ctx))
 	}
 	return http.HandlerFunc(fn)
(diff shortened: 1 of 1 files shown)


harikb:
travis build is failing with an error https://travis-ci.org/go-chi/chi/jobs/460622264#L479 on Go 1.8 only but this seems unrelated to the changes above

middleware/request_id.go
```diff
@@ -86,3 +86,22 @@ func GetReqID(ctx context.Context) string {
 func NextRequestID() uint64 {
 	return atomic.AddUint64(&reqid, 1)
 }
+
+// RequestIDFromHeader is a middleware similar to RequestID
+// except it uses the value from request header X-Request-Id
+// if one is available. Else revert to default behavior.
+func RequestIDFromHeader(next http.Handler) http.Handler {
```
VojtechVitek: Can we provide a `SetReqID()` function, instead of hardcoding the X-Request-Id header name?

VojtechVitek: Would be nice to provide and example for #368 as part of this PR.

harikb: I thought about this first. However I thought it was more intrusive change. The RequestID is maintained in a context variable keyed by a non-exported variable. Exporting that variable doesn't sound good either. So it would have a SetReqID() method inside the middleware which also takes the context as an explicit parameter, and set the given string as the request-id. Is that what you are looking for?  As for #368 , isn't there already a `GetReqID()` method available in the middleware?

VojtechVitek: You know what.. I think it's not a bad idea to use the `X-Request-ID` header directly after all. This middleware is opinionated anyway.

Imho, the is the desired lifecycle of "tracing" the request-id across multiple services is as follows:

1. browser request

 ->

2. optionally, an external service, creates reqID and passes it as X-Request-ID header

 ->

3. our program reads the header and sets the context value, ie.
```go
r.Use(middleware.RequestID)

// which behind the scenes does this:
reqID := r.Header.Get("X-Request-ID")
if reqID == "" {
  reqID = generateReqId()
}
setReqId(ctx, reqID)
```

 ->

4. middlewares/handlers use the reqID value (from ctx) and pass it to logger or to external HTTP calls as X-Request-ID header

    we just provide a getter for this:
```go
reqID := middleware.GetReqId(ctx)

// log.With().String("requestId", reqID) // zerolog structured log example
// req.Header.Set("X-Request-ID", reqID) // HTTP calls to other services that should be aware of this request ID
```

harikb: Item 4 already exists. Please confirm if that is what you meant.
So the remaining change is to merge RequestIDFromHeader in to RequestID middleware. I wasn't sure if it would break any existing usage.

PR #379, opened 2019-01-08
chi v4 + Go modules

VojtechVitek:


VojtechVitek:
This PR was rejected by the main chi author. He doesn't like the idea of changing the import paths because of Go modules support (ie. `github.com/go-chi/chi/v4`) and become a tech support for the confused chi users. So we're keeping it simple and chi stays on `+incompatible` flag for now.

PR #386, opened 2019-01-19
add a middleware to change post method to "_method" in HTML form

tomocy:
This middleware enables us to request with other methods than GET and POST from HTML form.
In order to change to these methods, 
add (hidden) input named "_method" in POST form and set desired method to the value.
eg)

```html
<form action="/articles/1" method="POST">
    <input type="hidden" name="_method" value="PUT">
</form>
```

```go
r := chi.NewRouter()
r.Use(ChangePostToHiddenMethod)
r.Put("/articles/{id}", func(w http.ResponseWriter, r *http.Request) {
    // do something
})
```

Now, the changeable methods are PUT and DELETE.

### DIFF
--- a/middleware/change_to_hidden_method.go
+++ b/middleware/change_to_hidden_method.go
@@ -1,0 +1,24 @@
+package middleware
+
+import "net/http"
+
+func ChangePostToHiddenMethod(next http.Handler) http.Handler {
+	var changeableMethods = map[string]bool{
+		http.MethodPut:    true,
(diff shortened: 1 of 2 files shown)


tomocy:
Thanks for commenting

I understand, so I close and will move that middleware in my repo

VojtechVitek:
@tomocy sounds good. You can also reference that repo from our main README.md file if you want :) Cheers

PR #391, opened 2019-01-28
Added middleware for AppEngine Context

kazukousen:
Using Libraries for AppEngine/Go (https://github.com/golang/appengine) requires the context derived from an AppEngine Context.
this PR is added an middleware that http.Handler wrapper for it.

### DIFF
--- a/middleware/appengine.go
+++ b/middleware/appengine.go
@@ -1,0 +1,29 @@
+package middleware
+
+import (
+	"context"
+	"net/http"
+
+	"google.golang.org/appengine"
+)
+
+var (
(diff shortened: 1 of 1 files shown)


pkieltyka:
@kazukousen thanks, this looks pretty good -- but, I wonder if this method would work too, https://github.com/go-chi/chi/blob/master/context.go#L134 setting the base context with appengine context

kazukousen:
thank you for reply.
sorry for my late reply.
`BaseContext` looks nice.
I'll use it.

Thanks.

Issue #277, opened 2017-11-11
Panic on empty path segments in request


cskr:
`chi.URLParam` panics when path segments matching path variables is empty.

chi version: 94e6bc0
Go version: go1.9.2 linux/amd64

Sample code for reproduction:

```
package main

import (
	"fmt"
	"net/http"

	"github.com/go-chi/chi"
)

func main() {
	mux := chi.NewMux()
	mux.Get("/prefix/{x}/{y}/suffix", func(w http.ResponseWriter, r *http.Request) {
		fmt.Fprintln(w, chi.URLParam(r, "x"), chi.URLParam(r, "y"))
	})

	http.ListenAndServe(":8080", mux)
}
```

curl command:

```
curl 'http://localhost:8080/prefix///suffix'
```

Panic:

```
2017/11/11 19:31:39 http: panic serving 127.0.0.1:52118: runtime error: index out of range
goroutine 5 [running]:
net/http.(*conn).serve.func1(0xc42007ad20)
	$GOROOT/src/net/http/server.go:1697 +0xd0
panic(0x675140, 0x82f160)
	$GOROOT/src/runtime/panic.go:491 +0x283
github.com/go-chi/chi.(*Context).URLParam(0xc4200fc000, 0x6c5b1c, 0x1, 0x0, 0xc420010ac0)
	$HOME/go/src/github.com/go-chi/chi/context.go:74 +0x102
github.com/go-chi/chi.URLParam(0xc4200f6100, 0x6c5b1c, 0x1, 0x8, 0xc4200ee048)
	$HOME/go/src/github.com/go-chi/chi/context.go:108 +0x6e
main.main.func1(0x805f20, 0xc4200f8000, 0xc4200f6100)
	$HOME/test.go:13 +0x55
net/http.HandlerFunc.ServeHTTP(0x6d8940, 0x805f20, 0xc4200f8000, 0xc4200f6100)
	$GOROOT/src/net/http/server.go:1918 +0x44
github.com/go-chi/chi.(*Mux).routeHTTP(0xc420060120, 0x805f20, 0xc4200f8000, 0xc4200f6100)
	$HOME/go/src/github.com/go-chi/chi/mux.go:428 +0x289
github.com/go-chi/chi.(*Mux).(github.com/go-chi/chi.routeHTTP)-fm(0x805f20, 0xc4200f8000, 0xc4200f6100)
	$HOME/go/src/github.com/go-chi/chi/mux.go:372 +0x48
[...]

PR #392, opened 2019-02-15
Solve context panic caused by index out of range issue

PumpkinSeed:
I found out the #277 issue, and we also run into this panic. Preferably it would be nice to have a check for index out of range, however it won't cause any other malfunctioning, but at least filter out the panic caused of empty param.


pkieltyka:
yea, if you provide a test case, I can review+fix it. this sort of bug should never happen though if the parser is doing its job correctly

PumpkinSeed:
I will look into where is the best to insert this test and push into the PR.

PR #395, opened 2019-02-22
fix SetEncoder for adding to encodingPrecedence

pdedmon:
I think we want to add new encodings to the precedence list, rather than empty string.


pdedmon:
Or maybe the condition is supposed to be `if e != ""`?

pkieltyka:
indeed, I will change it to `if e != ""` .. thank you

PR #398, opened 2019-02-25
Resolve Possible Panic for Out of Range URLParam

husainaloos:
This is related to #277. Added tests to confirm the fix is correct.

This panic should not be possible if this was written with good encapsulation. Unfortunately users might be able to manipulate the values directly (like in the test case) and that usage could cause a panic.


husainaloos:
This is also related to PR #392

PR #401, opened 2019-02-26
Re-wrote the middleware compress library.

### LINKED ISSUE #397: rewrite Compress middleware
the Compress middleware is a bit of a mess and needs a rewrite. As this is not core to the chi router, it's not something I have cycles to work on. However, I'd be happy to see a community contribution that deletes and replaces our current middleware.Compress handler.

A few suggestions/ideas for the new Compress middleware:

- [ ] Compress handler should be implemented as its own type/struct where additional encoders can be set. As opposed to currently the implementation is a single handler, and a lot of logic is outside of it in package-variables. ie. `middleware.SetEncoder()` pollutes the namespace and should be moved.

proposed syntax..

```go
r := chi.NewRouter()
compress := middleware.Compress(5, "text/other")
compress.SetEncoder("br", func(..) { ...})
r.Use(compress.Handler)
...
```

- [ ] Reduce GC overhead by using sync.Pool -- see https://github.com/go-chi/chi/issues/306

awbraunstein:
This re-write introduces the concept of a Compressor which stores all the settings for how compression/encoding should happen. The old API was changed to use the new Compressor struct which should prevent this change from breaking current consumers.

Additionally, this uses a sync.Pool for encoders that have a Reset(io.Writer) method to reduce memory overhead.

The new API is as follows:

```Go
r := chi.NewRouter()
compressor := NewCompressor(5, "text/html", "text/css")
// Add in a no-op encoder.
compressor.SetEncoder("nop", func(w io.Writer, _ int) io.Writer {
	return w
})
r.Use(compressor.Handler())
```

This fixes #397.

### DIFF
--- a/middleware/compress.go
+++ b/middleware/compress.go
@@ -6,24 +6,75 @@ import (
 	"compress/gzip"
 	"errors"
 	"io"
+	"io/ioutil"
 	"net"
 	"net/http"
 	"strings"
+	"sync"
 )
 
-var encoders = map[string]EncoderFunc{}
+var defaultCompressibleContentTypes = []string{
+	"text/html",
+	"text/css",
+	"text/plain",
+	"text/javascript",
+	"application/javascript",
+	"application/x-javascript",
+	"application/json",
+	"application/atom+xml",
+	"application/rss+xml",
+	"image/svg+xml",
+}
 
-var encodingPrecedence = []string{"br", "gzip", "deflate"}
+// A default compressor that allows for the old API to use the new code.
+// DEPRECATED
+var defaultCompressor *Compressor
 
-func init() {
+// Compressor represents a set of encoding configurations.
+type Compressor struct {
+	level int // The compression level.
+	// The mapping of encoder names to encoder functions.
+	encoders map[string]EncoderFunc
+	// The mapping of pooled encoders to pools.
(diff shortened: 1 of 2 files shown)


pkieltyka:
wow @awbraunstein well done. the code looks very sharp and complete. I will review it in detail this evening. thanks for the quick turnaround

pkieltyka:
did a review, tested various edge cases manually, and works great. thanks again @awbraunstein

abenz1267:
I just noticed it breaks images, if the extension is not in allowedTypes.

```go
r := chi.NewRouter()

	c := middleware.Compress(5)
	r.Use(c)

	wDir, _ := os.Getwd()
	path := filepath.Join(wDir, "/public")
	fs.Serve(r, "/public", http.Dir(path))

	http.ListenAndServeTLS(":3000", os.Getenv("CERT"), os.Getenv("KEY"), r)
```
https://localhost:3000/public/hi1.jpg -> "The image ... can not be displayed because it contains errors". Same issue in Firefox and Chrome.

Adding the extensions to the allowedTypes "solves" this, but i wouldn't call this solved nor expected behaviour. If i don't want a certain image type to be compressed, it shouldn't break.

Furthermore: adding single content-types seems very cumbersome at the moment, as you have to then also manually add the default content-types.

Regards

Issue #409, opened 2019-03-09
Middleware.Compress breaks images (and possibly more?)


abenz1267:
I just noticed it breaks content, if the extension is not in allowedTypes.

```go
r := chi.NewRouter()

c := middleware.Compress(5)
r.Use(c)

wDir, _ := os.Getwd()
path := filepath.Join(wDir, "/public")
fs.Serve(r, "/public", http.Dir(path))

http.ListenAndServeTLS(":3000", os.Getenv("CERT"), os.Getenv("KEY"), r)
```
"The image ... can not be displayed because it contains errors". Same issue in Firefox and Chrome.

Adding the extensions to the allowedTypes "solves" this, but i wouldn't call this solved nor expected behaviour. If i don't want a certain image type to be compressed, it shouldn't break.

Furthermore: adding single content-types seems very cumbersome at the moment, as you have to then also manually add the default content-types.

Regards

**EDIT: after adding the content-types manually it broke everything else. So: the middleware breaks everything that's not in the allowedTypes.**

Edit 2: it kinda makes sense, since the response will always be written through the encoder:
```go
Line 341: return cw.w.Write(p)
```

where `cw.w` = encoder

abenz1267:
Possible fix:

```go
// Is the content type compressable?
if _, ok := cw.contentTypes[contentType]; !ok {
cw.w = cw.ResponseWriter
return
}
```

Although it would be nice to determine the content-type earlier and see if you can avoid every further computation, if it's not supposed to be compressed and instead of next.ServeHTTP(w,r) instead

chenjie4255:
I meet this same issue and it took me a whole day to debug what was going wrong with my code. However, it finally turned out to be a ridiculous issue with the framework of chi, I suggest fixing this issue ASAP or at least adding some warn document about this broked middleware. It should stop people using it anymore if this bug must be fixed until the next major version.

pkieltyka:
hey gents, sorry for the hiccups with this middleware.

This PR fixes the issues described, and doesn't require a major version update. https://github.com/go-chi/chi/pull/422

lmk feedback and I'll merge. I've tested it manually in a helper example something to the effect..

```golang
package main

import (
	"io/ioutil"
	"log"
	"net/http"

	"github.com/go-chi/chi"
	"github.com/go-chi/chi/middleware"
)

func main() {
	r := chi.NewRouter()
	r.Use(middleware.RequestID)
	r.Use(middleware.Logger)
[...]

PR #410, opened 2019-03-09
Compress Middleware, fixed ResponseWriter breaking content

abenz1267:
see https://github.com/go-chi/chi/issues/409

I've removed all the deprecated code, since I actually don't know what it's good for. The old compress middleware was borked and this one should be used anyways.

I've also added a bool in order to actively set the defaultTypes, instead of passively enforcing them in case you don't input any content-types. This makes it much more convient to add content-types on top of the default ones.

Regards

### DIFF
--- a/middleware/compress.go
+++ b/middleware/compress.go
@@ -26,10 +26,6 @@ var defaultCompressibleContentTypes = []string{
 	"image/svg+xml",
 }
 
-// A default compressor that allows for the old API to use the new code.
-// DEPRECATED
-var defaultCompressor *Compressor
-
 // Compressor represents a set of encoding configurations.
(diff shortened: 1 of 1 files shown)


abenz1267:
Yes, this change would be indeed breaking, due to the change of how default types are added. However it would be good to still add line 278 to the current version, as it fixes the broken middleware. I'm not sure if i should make a new PR for this or if you just wanna add that 1 line.

VojtechVitek:
Please, submit a new issue (bug report) with description of what's broken. Any reproducers are much appreciated. We'll accept a new PR with the fix for this. Thanks!

To suggest a breaking API change for next major release, please submit a new issue with `v4` label.

PR #417, opened 2019-04-13
middleware: fixes #409 to encode only content-types in allowedTypes

### LINKED ISSUE #409: Middleware.Compress breaks images (and possibly more?)

aaharu:
#409 #410

I want to fix this problem before next major release...

### DIFF
--- a/middleware/compress.go
+++ b/middleware/compress.go
@@ -322,6 +322,7 @@ func (cw *compressResponseWriter) WriteHeader(code int) {
 
 	// Is the content type compressable?
 	if _, ok := cw.contentTypes[contentType]; !ok {
+		cw.w = cw.ResponseWriter
 		return
 	}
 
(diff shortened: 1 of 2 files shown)


pkieltyka:
see https://github.com/go-chi/chi/pull/422 instead which covers a few more cases/edges

PR #425, opened 2019-05-22
adding a configurable request id middleware

jhjaggars:
For our project we want to use a different header to track our request id.
This change adds a new method that wraps the current one to allow you to
configure the header like so:

r.Use(middleware.ConfigurableRequestID("my-header"))


jhjaggars:
Looks like the failure is unrelated to my change.  Would you like me to fix that error in this PR?

pkieltyka:
hey @jhjaggars thanks for the PR. How about just copying and supporting your own local version of requestID middleware that suits your needs? I prefer to keep surface area of code as small as possible in the core

Issue #432, opened 2019-07-13
Accept content-type wildcards for compress middleware


flimzy:
I would find it useful to be able to pass, say `text/*` as a content type to the compress middleware.
Would a PR to add this functionality be considered useful?

I'd be happy to write such a PR, if so.

I think there are two possible approaches. I'm curious which people would prefer:

1. Accept a wildcard just for the subtype.  i.e. `text/*`
2. Accept a wildcard anywhere in the type, to allow for things like `audio/*wav`, to match `audio/wav`, `audio/x-wav`, etc,  or `text/json*` to match `text/json` and `text/json+suffix`.

For my purposes, 1 is sufficient, but while I'm at it, I could add support for 2 as well, if it would be of value to others.

pkieltyka:
hey @flimzy thats a cool idea, I think it makes sense. I'll have to review your PR in detail, but feel free to package a custom version of the middleware in your own projects, as would be best way to move forward without having to wait for upstream middlewares which have a longer review process since its only me and sometimes Vojtech

flimzy:
Thanks for the response. I'm already using my fork of chi, with the changes in the PR, for my own project.

flimzy:
I've been using my own modifications for a while now without any problems. Would it be appropriate to review the PR at this stage?

pkieltyka:
@flimzy sure, a PR would be great. Can you make sure its off of the latest version though? the compress middleware has changed a few times over the last few versions

flimzy:
I'll work on rebasing the PR, although it looks like it still merges cleanly.  There's one question in particular that would be good to answer, even before a rebase. And maybe you can comment on the general applicability of the approach?

pkieltyka:
@flimzy i've reviewed your PR and it looks great -- as you mentioned you've only implemented option 1 above, which I also think is sufficient for this middleware. However, what if someone attempts to use `audio/*wav`, and would think it matches the wildcard (as in option 2), but would silently fail to work? Perhaps you could add a simple error checking to ensure that in case of a wildcard content-type def, that `*` is the last character? otherwise it could `panic("middleware/compress: content-type patterns with a wildcard must define the '*' at the end")`

flimzy:
Updated with the suggested panic, and I added some tests for the new functionality.

pkieltyka:
merged :) well done @flimzy -- thank you for the contribution

flimzy:
Thanks for taking the time to review and merge!

PR #433, opened 2019-07-13
Allow wildcards for subtypes in compressor

flimzy:
Fix for #432 and RFC

### DIFF
--- a/middleware/compress.go
+++ b/middleware/compress.go
@@ -5,6 +5,7 @@ import (
 	"compress/flate"
 	"compress/gzip"
 	"errors"
+	"fmt"
 	"io"
 	"io/ioutil"
 	"net"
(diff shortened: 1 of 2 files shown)


middleware/compress.go
```diff
@@ -38,7 +38,8 @@ type Compressor struct {
 	// The mapping of pooled encoders to pools.
 	pooledEncoders map[string]*sync.Pool
 	// The set of content types allowed to be compressed.
-	allowedTypes map[string]bool
+	allowedTypes     map[string]struct{}
+	allowedWildcards map[string]struct{}
```
flimzy: This might be a contentious change.

I always prefer `struct{}` when a map is used to check only for existence, because then I'm never left wondering "does the bool value mean something?"

It also uses less memory (one byte less memory per map entry), which is, at best, a minor concern here, where maps probably contain tens of items, or less, on average.

I've met a few fellow developers who hate this convention, though.  So if this bothers anyone, I can revert to `bool` :)

pkieltyka: since its an internal type and `struct{}` takes less memory in the map, better to go with the `struct{}` if its just intended to work as a marker

flimzy:
This is now rebased against latest master.

PR #440, opened 2019-08-01
documented Handlers map key

tkivisik:
As it's not immediately clear what is used as key in Handlers map, I added a line to document it.

### DIFF
--- a/tree.go
+++ b/tree.go
@@ -795,6 +795,7 @@ func (ns nodes) findEdge(label byte) *node {
 }
 
 // Route describes the details of a routing handler.
+// Handlers map key is an HTTP method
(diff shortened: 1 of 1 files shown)


tree.go
```diff
@@ -795,6 +795,7 @@ func (ns nodes) findEdge(label byte) *node {
 }
 
 // Route describes the details of a routing handler.
+// Handlers map key is an HTTP method
```
gayanhewa: The comment doesn't necessarily make it clear though

tkivisik: Would `// Handlers map key is an HTTP method, e.g. "get"` be clearer?

PR #461, opened 2019-10-16
Make fileserver example a little more elaborate

dschmidt:
While it's trivial to remove the basePath from the code, it wasn't quite obvious to me which paths I had to strip where to make the fileserver work if not mounted to /.


dschmidt:
ping? :)

dschmidt:
Sweet, thanks!

PR #465, opened 2019-10-31
Fix cast panic and just return nil if invalid

coryory:
https://play.golang.com/p/yXfFGBIltXY


pkieltyka:
indeed, this wouldn't happen from chi itself, but it could happen to a caller if they pass the wrong ctx arg. thx for the PR

PR #470, opened 2019-11-14
AllowContentType middleware refactor and tests

bernardopericacho:
- adds tests for `AllowContentType` middleware
- use map for better lookup performance

### DIFF
--- a/middleware/content_type.go
+++ b/middleware/content_type.go
@@ -19,9 +19,9 @@ func SetHeader(key, value string) func(next http.Handler) http.Handler {
 // AllowContentType enforces a whitelist of request Content-Types otherwise responds
 // with a 415 Unsupported Media Type status.
 func AllowContentType(contentTypes ...string) func(next http.Handler) http.Handler {
-	allowedContentTypes := make(map[string]bool, len(contentTypes))
+	allowedContentTypes := make(map[string]struct{}, len(contentTypes))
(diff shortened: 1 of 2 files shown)


middleware/content_type.go
```diff
@@ -37,11 +37,9 @@ func AllowContentType(contentTypes ...string) func(next http.Handler) http.Handl
 				s = s[0:i]
 			}
 
-			for _, t := range cT {
-				if t == s {
-					next.ServeHTTP(w, r)
-					return
-				}
+			if allowedContentTypes[s] {
```
mfridman: If all you're doing is checking existence _within_ the map, might as well use `map[string]struct{}` and then comma okay. We don't actually care about the value, since they are all true.

```
if _, ok := allowedContentTypes[s]; ok {
	next.ServeHTTP(w, r)
	return
}
```

bernardopericacho: Done. Tested with a benchmark test and it is a few nanoseconds better

pkieltyka:
With latest versions of go 1.3+, what is the consensus that performs faster? a map lookup or a tight string+for loop?

thanks for the PR

bernardopericacho:
I did a benchmark test after @mfridman comment and it was way better in Go 1.13. 
Let me know any next steps needed in order to merge this PR

PR #471, opened 2019-11-19
Include Allow header in 405 Method Not Allowed response. Closes #446

### LINKED ISSUE #446: Include Allow header in 405 Method Not Allowed response
At the moment the default [methodNotAllowedHandler](https://github.com/go-chi/chi/blob/43097498be03c1a12a01dd82d36f4465cde9bcb4/mux.go#L457) doesn't include an `Allow` header containing the resource's supported methods.

I've made a minimal runnable example here: [https://play.golang.org/p/CSUjlf8e30S](https://play.golang.org/p/CSUjlf8e30S)

According to the [HTTP RFC](https://tools.ietf.org/html/rfc7231#section-6.5.5) the `Allow` header must be included in 405 responses, so it would be good if there is a way that chi could support this.

bernardopericacho:
- Include `Allow` Header for `405 MethodNotAllowed` responses. Context [RFC](https://tools.ietf.org/html/rfc7231#section-6.5.5)
- Include tests for default `methodNotAllowed` handler
- improve performance in the [node routes](https://github.com/go-chi/chi/compare/master...bernardopericacho:allow_header?expand=1#diff-37cf0fd7a164303538a2d22d965646dfR625) checking existence on a map instead of doing a loop calling [`methodTypString`](https://github.com/go-chi/chi/compare/master...bernardopericacho:allow_header?expand=1#diff-37cf0fd7a164303538a2d22d965646dfL750)

### DIFF
--- a/mux.go
+++ b/mux.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"fmt"
 	"net/http"
-	"sort"
 	"strings"
 	"sync"
 )
@@ -355,11 +354,11 @@ func (mx *Mux) NotFoundHandler() http.HandlerFunc {
 
 // MethodNotAllowedHandler returns the default Mux 405 responder whenever
 // a method cannot be resolved for a route.
-func (mx *Mux) MethodNotAllowedHandler(allowedMethods ...string) http.HandlerFunc {
+func (mx *Mux) MethodNotAllowedHandler() http.HandlerFunc {
 	if mx.methodNotAllowedHandler != nil {
 		return mx.methodNotAllowedHandler
 	}
-	return methodNotAllowedHandler(allowedMethods...)
+	return methodNotAllowedHandler()
 }
 
 // buildRouteHandler builds the single mux handler that is a chain of the middleware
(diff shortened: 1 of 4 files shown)


bernardopericacho:
Not sure why is failing for `go 1.12` only
https://travis-ci.org/go-chi/chi/builds/614093612?utm_medium=notification&utm_source=github_status

mux.go
```diff
@@ -454,7 +455,13 @@ func (mx *Mux) updateSubRoutes(fn func(subMux *Mux)) {
 
 // methodNotAllowedHandler is a helper function to respond with a 405,
 // method not allowed.
-func methodNotAllowedHandler(w http.ResponseWriter, r *http.Request) {
-	w.WriteHeader(405)
-	w.Write(nil)
+func methodNotAllowedHandler(allowedMethods ...string) http.HandlerFunc {
+	fn := func(w http.ResponseWriter, r *http.Request) {
+		// sort them so they always appear in the same order
+		sort.Strings(allowedMethods)
```
pkieltyka: how about instead of sorting we use the natural order as defined by `methodMap` ? would produce better performance for methods that return 405, without requiring a string sort on the invocation

mux.go
```diff
@@ -354,11 +355,11 @@ func (mx *Mux) NotFoundHandler() http.HandlerFunc {
 
 // MethodNotAllowedHandler returns the default Mux 405 responder whenever
 // a method cannot be resolved for a route.
-func (mx *Mux) MethodNotAllowedHandler() http.HandlerFunc {
+func (mx *Mux) MethodNotAllowedHandler(allowedMethods ...string) http.HandlerFunc {
```
pkieltyka: for this exported method, what is the benefit of the `allowedMethods` argument? the for internal method you're using it to pass the request content methodNotAllowed string array of methods that are allowed, but why for the exported method?

mux.go
bernardopericacho: done, thanks for the review!

mux.go
bernardopericacho: I did that initially so other parts on the code could eventually use it. Instead, I follow your recommendation and addressed this way https://github.com/go-chi/chi/pull/471/files#diff-b303fad270323289aeb17e2f0930ffa7R428. Let me know what you think

Issue #462, opened 2019-10-17
What's up with go module support?


MaikuMori:
Go `1.13` is out, `GOPATH` is gone. Most of the packages have go module support.

What's holding Chi back? How can I help?

VojtechVitek:
> From https://github.com/golang/go/wiki/Modules#releasing-modules-v2-or-higher
> 
> Note that if you are adopting modules for the first time for a pre-existing repository or set of packages that have already been tagged v2.0.0 or higher before adopting modules, then the recommended best practice is to increment the major version when first adopting modules.

As per the above official recommendation, chi's not going to adopt Go modules until we plan to release v5.0.0.

Otherwise, we'd have to introduce a breaking API change, as we'd have to change the import paths to `"github.com/go-chi/chi/v4"`. See https://github.com/go-chi/chi/pull/379/files how the changes would look like. We opted out for now.

The `+incompatible` flag works OK for now. You can install latest version of chi with `GO111MODULE=on go get -u github.com/go-chi/chi@v4.0.2`.

pkieltyka:
I've been thinking about this for a while and I believe I have the approach that will make me the happiest. Chi needs to get proper go.mod compatibility, and with our current versioning from v1.0.0 to v4.0.0, it requires as @VojtechVitek suggests to add the suffix `/v4` at the end of all import paths. If I could go back in time, I would have just released chi versions v1.1, v1.2, v1.3, and v1.4 -- instead of v2.0.0, v3.0.0, v4.0.0. In keeping with v1.x versioning, we can keep the import path as it is with `github.com/go-chi/chi`, and we get module compatibility.

Therefore, here's my idea. We can still do that! and re-version v2.0.0 as v1.2, v3.0.0 as v1.3, and v4.0.0 as v1.4.

VojtechVitek:
That is an interesting idea!

But we'd have to make it clear that v1.4 is not really backward compatible with all v1.x versions, as it should be per semver specs. Example of breaking changes:
- v1.0 had pre-context signatures (`func(ctx context.Context, w http.ResponseWriter, r *http.Request)`)
- v2.0 had old URL parameter syntax (`/:id` instead of `/{id}`)
- probably more stuff, I'm not sure

But thinking about it.. when did we switch from https://github.com/pressly/chi to https://github.com/go-chi/chi? Maybe go-chi/chi v1.0.0 wasn't actually referenced by anyone and we're good to go? (and maybe we could even go for v1.0.0 ?! :D)

👍

[...]

PR #473, opened 2019-11-26
feat: add go mod to remove +incompatible suffix in dependent projects

akkumar:
Add go mod (and .sum) files .

This makes the list of dependencies clearer (even if none exists outside stdlib in this case). 

It also prevents the ugly +incompatible suffix that is added in downstream projects dependent on this.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,0 +1,5 @@
+module github.com/go-chi/chi
+
+go 1.13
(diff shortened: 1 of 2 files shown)


cristaloleg:
see https://github.com/go-chi/chi/issues/462

PR #474, opened 2019-12-05
Retry-After HTTP header in Throttle middleware

vasayxtx:
I think it would be good to be able to set a "Retry-After" HTTP header for response with 503 status code in `Throttle` middleware when the limit is exceeded. Thus, the HTTP client will know after how long it can repeat the request.

I need this in a several projects, probably it will be useful for someone else.

### DIFF
--- a/middleware/throttle.go
+++ b/middleware/throttle.go
@@ -2,6 +2,7 @@ package middleware
 
 import (
 	"net/http"
+	"strconv"
 	"time"
 )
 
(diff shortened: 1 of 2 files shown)


pkieltyka:
Nice PR, thanks for the addition. I'll give this an indepth review once I have a chance. Other reviewers welcome too :)

vasayxtx:
@pkieltyka thanks! I'm waiting for comments and suggestions:)

Issue #426, opened 2019-05-28
URLParam need to check is nil or not


powxiao:
the route like below,
```
func (h *PostSessionOperate) URL() string {
	return "/api/v1/cps/sessions/{session_id}/operate"
}
```
and post request as localhost:8080/api/v1/cps/sessions//operate, then in function
```
func (x *Context) URLParam(key string) string {
	for k := len(x.URLParams.Keys) - 1; k >= 0; k-- {
		if x.URLParams.Keys[k] == key {
			return x.URLParams.Values[k]
		}
	}
	return ""
}
```
x.URLParams.Values is nil , but k ==0 , use x.URLParams.Values[0] panic occurs.

![image](https://user-images.githubusercontent.com/1414815/58455061-06ed6280-8153-11e9-8797-c9a8935aa329.png)

DCRcoder:
meet same problem 
if server get /api/v1/cps/sessions//operate request like this example. there will be panic.
is this expect？

rof20004:
@DCRcoder @powxiao your implementation got panic?

DCRcoder:
yes URLParams.Keys existed but URLParams.values not
there a route like this
```
r.Get('/url/{param:(\\w+)}/test', xxx)
```
if web get request '/url//test' that missing url param, URLParams.values will panic because no value in there.
> @DCRcoder @powxiao your implementation got panic?
> 
> I think URLParams.Keys will not be nil, because is not a pointer.

moorereason:
I was only able to reproduce this panic when using a regexp param.  A regular param (like the OP) doesn't trigger a panic for me.

I've pushed a PR that adds a failing test.

pkieltyka:
thanks for the report + suggestions -- fixed: https://github.com/go-chi/chi/commit/3658d98ee4dc2e8e25a4b7cf759ad0dd42f4c075

PR #476, opened 2019-12-21
Add test for #426

moorereason:


nu50218:
`(\w+)` does not match `""`?

pkieltyka:
fixed: https://github.com/go-chi/chi/commit/3658d98ee4dc2e8e25a4b7cf759ad0dd42f4c075

PR #477, opened 2019-12-22
tree.go: check if ntParam, ntRegexp was found

### LINKED ISSUE #426: URLParam need to check is nil or not

nu50218:
maybe this will fix #426 .

Since the current `(*node).findRoute` does not check, If there is a route like `r.Get('/url/{param:(\\w+)}/test', f)` and get request `/url//test` (don't match), Router calls `f` and causes the bug.

Forked from #476 and fixed it, so if this PR will be merged, please also close #476 .

### DIFF
--- a/mux_test.go
+++ b/mux_test.go
@@ -93,6 +93,11 @@ func TestMuxBasic(t *testing.T) {
 		w.Write([]byte("woop." + URLParam(r, "iidd")))
 	}
 
+	pingPoop := func(w http.ResponseWriter, r *http.Request) {
+		w.WriteHeader(200)
+		w.Write([]byte("poop." + URLParam(r, "id")))
+	}
+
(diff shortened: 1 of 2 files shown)


pkieltyka:

PR #483, opened 2020-01-14
Flush compressed streams

remko:
When using stream compression in combination with (SSE) event streams, you need to flush every time you sent an event (to ensure data still in the compression buffer is effectively sent).

The compression middleware's Flush doesn't flush compressed data, because the underlying writer (e.g. gzip.Writer)'s Flush doesn't have the http.Flusher signature (because it has an error return argument). 

This patch flushes the underlying stream if it implements gzip's flush signature. 

After flushing the GZip stream, the underlying response writer will have pending data as well, and so needs to be flushed too.

### DIFF
--- a/middleware/compress.go
+++ b/middleware/compress.go
@@ -382,9 +382,23 @@ func (cw *compressResponseWriter) writer() io.Writer {
 	}
 }
 
+type compressFlusher interface {
+	Flush() error
+}
+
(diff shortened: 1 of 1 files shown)


ChadTaljaardt:
I am getting the same issue as listed here, @pkieltyka Could you take a look at this when you get a second?

PR #488, opened 2020-02-05
Fix typo

jeffwidman:


pkieltyka:
thx but I like "pattern"

jeffwidman:
Hmm... how is that consistent with:
* the very next sentence of that docstring?
 https://github.com/go-chi/chi/pull/488/files#diff-05a3aa0ab9e5687ff167e0ef3ae4c5a6R28
* The name of the variable? 
https://github.com/go-chi/chi/pull/488/files#diff-05a3aa0ab9e5687ff167e0ef3ae4c5a6L29

In general, it feels a little weird to me to use the singular form of an object to refer to multiple objects... typically you'd use the plural form.

PR #495, opened 2020-03-12
Use http status constants instead of numbers

kkirsche:
Use http status constants instead of numbers to improve readability of the code


kkirsche:
To each their own, personally I disagree but I can understand that point of view

pkieltyka:
I as well prefer to read the codes as numbers. thank you though for the work and PR

PR #496, opened 2020-03-24
middleware.Recoverer, now with pretty stack traces

pkieltyka:
The stack traces have been a bit unwieldy so I've added a built-in decorator when printing stack traces caught by the middleware.Recoverer

**Before**:
image

**Now**:
image


pkieltyka:
there is also a version by uncommenting two lines that could produce this result:

image

if folks prefer this lmk

PR #505, opened 2020-04-15
Fix RedirectSlashes, Closes #504

### LINKED ISSUE #504: Redirect error in RedirectSlashes
RedirectSlashes middleware can redirect users to external URL.
Example:  http://myserver.com//evil.com/ will redirect to evil.com.

sintah:


middleware/strip.go
```diff
@@ -47,6 +49,9 @@ func RedirectSlashes(next http.Handler) http.Handler {
 			} else {
 				path = path[:len(path)-1]
 			}
+			if u, err := url.Parse(path); err == nil && u.Host != "" && u.Host != r.Host {
+				path = fmt.Sprintf("//%s/%s", r.Host, strings.TrimLeft(strings.TrimRight(path, "/"), "/"))
+			}
```
dustinmoris, line 54: Doesn't the `strings.TrimLeft` and `strings.TrimRight` trim potentially too many `/` and therefore change the actual request? I'm just thinking that there could be a reason for someone calling a route like `https://example.org/foo//bar/`. At least if someone did support such a route then this would be a breaking change.

Instead of the new if statement here, wouldn't this be sufficient:

```go
url := fmt.Sprintf("//%s%s", r.Host, path)
http.Redirect(w, r, url, 301)
```

PR #511, opened 2020-05-18
middleware/logger: add X-Forwarded-For support

ddevault:
This just prefers to show the X-Forwarded-For header rather than
http/Request.RemoteAddr, if present.


ddevault:
Haha, thanks. Not sure how I missed that.

PR #515, opened 2020-05-29
Replace wildcards correctly in RoutePattern

unmultimedio:
In case there are many `/*/` next to each other, this PR
Replaces all occurrences leaving the RoutePattern w/out
any wildcard.


unmultimedio:
Thanks, @pkieltyka for taking a look at this. I forgot to explain in the description, this is a backwards-incompatible change since it changes something in the response of a public interface.

On another topic, when can we expect a new release with this change?

pkieltyka:
@unmultimedio I'll prepare a release for you in a few minutes. I'd consider this a fix, so I'm inclined to make this a tiny release at 4.1.2

PR #516, opened 2020-06-10
[Proposal] link to middleware documentation

solderjs:
I've visted the README.md several times with the intention to get to the middleware documentation, but ending up one step removed. I propose to include the links of the middleware, in this style.

Thoughts?

PR #517, opened 2020-06-11
update complete middleware list w/ godoc descs

solderjs:
an alternate take on #516, which updates the table with the full list of middleware and the godoc description

### DIFF
--- a/README.md
+++ b/README.md
@@ -314,28 +314,97 @@ with `net/http` can be used with chi's mux.
 
 ### Core middlewares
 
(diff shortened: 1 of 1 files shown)


arp242:
This makes `less README.md` a lot harder to read; the concise descriptions were, IMHO, much better.

solderjs:
I don't disagree.

Just experimenting here to see about making it easy to automatically update the table in the README.

I think using a description up to the first `.` or newline might work.

However, that's a completely different discussion from the original issue  #516

PR #518, opened 2020-06-12
Add port number to RealIP middleware

### LINKED ISSUE #453: RealIP sets r.RemoteAddr without port
The Go documentation states the following on `http.Request.RemoteAddr`: 
```
The HTTP server in this package sets RemoteAddr to an "IP:port" address before invoking a handler.
```

Therefore, you could expect that a code like this is correct:

```go
package main

import (
	"fmt"
	"net"
	"net/http"

	"github.com/go-chi/chi"
	"github.com/go-chi/chi/middleware"
)

func main() {
	r := chi.NewMux()
	r.Use(middleware.RealIP)
	r.Get("/test", func(rw http.ResponseWriter, r *http.Request) {
		host, port, err := net.SplitHostPort(r.RemoteAddr)
		if err != nil {
			fmt.Fprintf(rw, "Error: %v\n", err)
		} else {
			fmt.Fprintf(rw, "Host: %s\nPort: %s\n", host, port)
		}
	})
	http.ListenAndServe(":8000", r)
}
```

However, the RealIP middleware is just copying the `X-Forwarded-For` value into `r.RemoteAddr`, which usually does not contain a port, making the code fail:

```
[dirbaio@mars]$ curl localhost:8000/test
Host: ::1
Port: 40188
[dirbaio@mars]$ curl localhost:8000/test -H 'X-Forwarded-For: 1.2.3.4'
Error: address 1.2.3.4: missing port in address
```

Perhaps RealIP should try to parse `X-Forwarded-For` for a `host:port`, and if it isn't, add a port? Maybe `0` to denote the port is unknown, like `1.2.3.4` -> `1.2.3.4:0`?

This is particularly frustrating with IPv6 addresses because they have colons but they're not the `host:port` colon, making it hard to parse the RemoteAddr in user code.

arp242:
Fixes #453

Downside: this is not entirely backwards compatible; e.g. my app will
actually break because it *assumes* you're running it behind a proxy and
won't remove the port. Then again, it already breaks for people not
using a proxy, but in certain scenarios (e.g. some closed source app)
that will probably never really happen.

So ... maybe something for v5? Just thought I'd open a PR in any case.

### DIFF
--- a/middleware/realip.go
+++ b/middleware/realip.go
@@ -4,6 +4,7 @@ package middleware
 // https://github.com/zenazn/goji/tree/master/web/middleware
 
 import (
+	"net"
 	"net/http"
 	"strings"
 )
@@ -29,7 +30,7 @@ var xRealIP = http.CanonicalHeaderKey("X-Real-IP")
 func RealIP(h http.Handler) http.Handler {
 	fn := func(w http.ResponseWriter, r *http.Request) {
 		if rip := realIP(r); rip != "" {
-			r.RemoteAddr = rip
+			r.RemoteAddr = net.JoinHostPort(rip, "0")
 		}
 		h.ServeHTTP(w, r)
 	}
--- a/middleware/realip_test.go
+++ b/middleware/realip_test.go
@@ -8,50 +8,69 @@ import (
 	"github.com/go-chi/chi"
 )
 
-func TestXRealIP(t *testing.T) {
-	req, _ := http.NewRequest("GET", "/", nil)
-	req.Header.Add("X-Real-IP", "100.100.100.100")
-	w := httptest.NewRecorder()
-
-	r := chi.NewRouter()
-	r.Use(RealIP)
-
-	realIP := ""
-	r.Get("/", func(w http.ResponseWriter, r *http.Request) {
-		realIP = r.RemoteAddr
-		w.Write([]byte("Hello World"))
-	})
-	r.ServeHTTP(w, req)
-
-	if w.Code != 200 {
-		t.Fatal("Response Code should be 200")
+func TestRealIP(t *testing.T) {
+	tests := []struct {
+		name       string
+		remoteAddr string
+		header     http.Header
+		want       string
+	}{
+		{
+			"remote-addr",
+			"1.1.1.1:42",
+			nil,
+			"1.1.1.1:42",
+		},
+		{
+			"x-real-ip",
+			"1.1.1.1",
+			http.Header{"X-Real-Ip": {"100.100.100.100"}},
+			"100.100.100.100:0",
+		},
+		{
+			"x-real-ip-6",
+			"2001:beef::0",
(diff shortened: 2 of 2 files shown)


middleware/realip.go
```diff
@@ -29,7 +30,7 @@ var xRealIP = http.CanonicalHeaderKey("X-Real-IP")
 func RealIP(h http.Handler) http.Handler {
 	fn := func(w http.ResponseWriter, r *http.Request) {
 		if rip := realIP(r); rip != "" {
-			r.RemoteAddr = rip
+			r.RemoteAddr = net.JoinHostPort(rip, "0")
```
VojtechVitek, line 33: Why would we add `:0` port to the address? This doesn't make sense to me. Can you clarify?

arp242: Because that's the format Go uses for RemoteAddr; see  #453

pavelpascari: I am wondering if the use of port `0` is actually the best way to go? 
Since `chi` is an HTTP router - wouldn't `1.2.3.4` be equal to `1.2.3.4:80` (the default HTTP port)?
Or am I missing something?

arp242: It's normally set to the source port, not destination port. Any value is "wrong", and at least 0 is "clearly wrong".

pkieltyka:
I think the best solution is to stop overwriting r.RemoteAddr from RealIP middleware and instead setting a value on the request context

jawnsy:
> I think the best solution is to stop overwriting r.RemoteAddr from RealIP middleware and instead setting a value on the request context

@pkieltyka I think the problem with that approach is that it requires downstream consumers (either other middlewares or handlers) to be aware that the realip middleware is higher in the chain. It also means that removing the middleware (e.g. commenting out the `r.Use` line) would either require changes to downstream code (to stop checking for that context), or, I suppose compatibility logic in the middleware to just return the unmodified `RemoteAddr` in case the context does not exist. That behavior, in turn, seems like it could make debugging difficult in that case.

There are also some other disadvantages of the current approach (such as being vulnerable to spoofing attacks and having limited support for different header types) for which I'd like to propose some changes (I'll open a separate issue for that). Thanks for this awesome library!

VojtechVitek:
Thanks for this PR — and sorry it took so long to get back to it.

Closing as superseded: `middleware.RealIP` is now deprecated in favor of the new `ClientIPFromHeader` / `ClientIPFromXFF` / `ClientIPFromRemoteAddr` middlewares (#1160). These never mutate r.RemoteAddr — it keeps the original ip:port from the TCP connection, so `net.SplitHostPort` always works and the root cause of #453 is gone. The client IP is read explicitly with `middleware.GetClientIP(r.Context())`.

I'm going to leave the deprecated `RealIP` behavior as-is, since changing RemoteAddr to ip:0 would break existing users who parse it as a bare IP.

PR #519, opened 2020-06-21
Add a hint about Logger middleware ordering

adw1n:
I've seen people make the mistake of using `Recoverer` and `Logger` middlewares in the wrong order which resulted in Logger logging `0` status code when panic happened. I think we could use a hint about middleware ordering.

I changed the table style because having to align `|` with spaces and `-` with more `----` characters seems to me like a waste of time.

### DIFF
--- a/README.md
+++ b/README.md
@@ -321,7 +321,7 @@ BasicAuth        | Basic HTTP authentication
 Compress         | Gzip compression for clients that accept compressed responses
 GetHead          | Automatically route undefined HEAD requests to GET handlers
 Heartbeat        | Monitoring endpoint to check the servers pulse
-Logger           | Logs the start and end of each request with the elapsed processing time. Logger should go before any middleware that may change the response like Recoverer.
+Logger           | HTTP request logger. Logger should go before any other middleware that may change the response like Recoverer.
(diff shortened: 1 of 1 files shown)


pkieltyka:
Nice, thanks for the suggestion.. can you update the text to:

```
HTTP request logger. Logger should go before any other middleware that may change the response like Recoverer.
```

we could also add some detection in Recoverer to check the if Logger is mounted after it, in which case panic with a warning

adw1n:
> can you update the text to:

done

> detection in Recoverer

Are you sure it belongs in Recoverer? My first thought was to add it to https://github.com/go-chi/chi/blob/86f9a6e7ce9bf453eaa339b51f88f586edbccbc1/mux.go#L96 and https://github.com/go-chi/chi/blob/86f9a6e7ce9bf453eaa339b51f88f586edbccbc1/mux.go#L226

Detection in Recoverer would have a runtime overhead rather than doing it once during the setup. And tbh it's not immediately obvious to me how to get a list of router middlewares from the inside of one of the middlewares.

pkieltyka:
Fixed in https://github.com/go-chi/chi/commit/042ce0e7de20d7ae20a8be26cff70a1be05a76a6

PR #528, opened 2020-07-03
Respond with 429 when throttled

davidc604:
Responding with `HTTP 429 Too Many Requests` status code seems more inline with what actually happened instead of the service unavailable response status code, because _technically_, the service is still available, it's just the sender has send too many requests in a given amount of time. (Typically a 400s error is client fault, whereas a 500s error is server).


pkieltyka:
@VojtechVitek let us know what you think

PR #534, opened 2020-08-11
Improve middleware.NoCache() performance

VojtechVitek:


arp242:
Do you have any benchmarks for this? It's generally a good idea to add some benchmarks so it's clear that it improves performance and by how much.

PR #537, opened 2020-09-03
Add permissions2

xyproto:
Add a link to [permissions2](https://github.com/xyproto/permissions2), a package that has a `Midddleware` function that is written specifically for `chi`.


pkieltyka:
thanks for the PR but I prefer to keep the list down to just go-chi packages

xyproto:
I don't get it. Why does the README say this then?

> please submit a PR if you'd like to include a link to a chi-compatible middleware

PR #541, opened 2020-09-28
Be explicit about the need for the @v4 tag

coolaj86:
Be explicit about using v4.x tags. Addresses https://github.com/go-chi/chi/issues/462.


6543 (approved):

VojtechVitek (approved):
LGTM. Thanks!

This is exactly what has been suggested a while ago at https://github.com/go-chi/chi/issues/462#issuecomment-585724509.

PR #542, opened 2020-10-01
Add request logger only for 4xx and 5xx

cikupin:
## What I do?

- Add one more request logger to only log request for  http status 4xx/5xx response `Logger4xxAnd5xx`

# Reason why I do this

Currently middleware `Logger` will log every request whether it's 2xx, 3xx, 4xx, and 5xx.  Sometimes we only need to log for unsuccessful request. That's why I propose this middleware.

In my project, logs are filled by http 200 request which most of them are health check (I run my project on kubernetes, so I need to provide health check endpoint). It's so painful to see my log full of unnecessary log like this. It makes me hard to find/see if error happened.

Screen Shot 2020-10-01 at 21 44 50

### DIFF
--- a/README.md
+++ b/README.md
@@ -331,28 +331,29 @@ with `net/http` can be used with chi's mux.
 
 ### Core middlewares
 
-----------------------------------------------------------------------------------------------------
-| chi/middleware Handler | description                                                             |
-| :--------------------- | :---------------------------------------------------------------------- |
-| [AllowContentType]     | Explicit whitelist of accepted request Content-Types                    |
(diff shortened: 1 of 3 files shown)


vtolstov:
I think you need to skip logging of kubernetes endpoints like health, live , metrics...

pkieltyka:
thanks for the PR, but I think this logger belongs in a subpackage on your repo, and not part of the chi core. If you're getting a lot of noise from healthchecks, write a middleware which skips those from the logger.

The one PR I would potentially accept is one that passes a filter func from a request and then you can skip the logger, at which point you could skip non 4xx/5xx, or skip the health checks

PR #549, opened 2020-11-12
use ConstantTimeCompare for password comparison

JamesClonk:
Using constant time comparison for the password to prevent timing attacks.

### DIFF
--- a/middleware/basic_auth.go
+++ b/middleware/basic_auth.go
@@ -1,6 +1,7 @@
 package middleware
 
 import (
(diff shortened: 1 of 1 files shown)


JamesClonk:
Ideally we should compare MD5 or SHA hashes of the password instead of comparing the raw password itself, to also avoid timing attacks on the length of the password, which just using ConstantTimeCompare alone does not prevent. But it's certainly better than nothing already :sweat_smile:

PR #553, opened 2020-11-24
Cleanup wildcard URLParam which connects the subrouter

pkieltyka:


mux.go
```diff
@@ -300,6 +300,13 @@ func (mx *Mux) Mount(pattern string, handler http.Handler) {
 	mountHandler := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		rctx := RouteContext(r.Context())
 		rctx.RoutePath = mx.nextRoutePath(rctx)
+
+		// reset the wildcard URLParam which connects the subrouter
+		n := len(rctx.URLParams.Keys) - 1
+		if n >= 0 && rctx.URLParams.Keys[n] == "*" && len(rctx.URLParams.Values) > n {
+			rctx.URLParams.Values[n] = ""
+		}
```
NonLogicalDev, line 308: Would it potentially be cleaner for this to happen in the Mux.ServeHTTP implementation? So that Mux will always override/augment the current parameters with their own?

Currently as I see : (on line 68, in `(mx *Mux) ServeHTTP`)
```
	// Check if a routing context already exists from a parent router.
	rctx, _ := r.Context().Value(RouteCtxKey).(*Context)
	if rctx != nil {
		mx.handler.ServeHTTP(w, r)
		return
	}
```

The context is currently not modified if it already exists, but in theory that site could be another place for this logic.

pkieltyka: Naw, it’s less performant and not in the right context. My PR implementation has little to no difference in my benchmarks.

PR #559, opened 2020-12-03
logger: unwrap to return original responseWriter

bernhardreisenberger:
currently the logger middleware returns a WrapResponseWriter which does not support the http.Hijacker interface.

Calling `Unwrap()` on the WrapResponseWriter, when passed to the handler will work.

This is necessary, because i cannot upgrade to websocket without the hijacker interface. I am using https://github.com/gorilla/websocket to upgrade.

### DIFF
--- a/middleware/logger.go
+++ b/middleware/logger.go
@@ -43,7 +43,7 @@ func RequestLogger(f LogFormatter) func(next http.Handler) http.Handler {
 				entry.Write(ww.Status(), ww.BytesWritten(), ww.Header(), time.Since(t1), nil)
 			}()
 
-			next.ServeHTTP(ww, WithLogEntry(r, entry))
+			next.ServeHTTP(ww.Unwrap(), WithLogEntry(r, entry))
 		}
 		return http.HandlerFunc(fn)
 	}
--- a/middleware/logger_test.go
+++ b/middleware/logger_test.go
@@ -1,0 +1,34 @@
+package middleware
+
+import (
+	"bufio"
+	"net"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+)
+
+type testLoggerWriter struct {
+	*httptest.ResponseRecorder
+}
+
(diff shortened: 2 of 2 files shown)


middleware/logger.go
```diff
@@ -43,7 +43,7 @@ func RequestLogger(f LogFormatter) func(next http.Handler) http.Handler {
 				entry.Write(ww.Status(), ww.BytesWritten(), ww.Header(), time.Since(t1), nil)
 			}()
 
-			next.ServeHTTP(ww, WithLogEntry(r, entry))
+			next.ServeHTTP(ww.Unwrap(), WithLogEntry(r, entry))
```
VojtechVitek, line 46: Please, correct me if I'm wrong, but this will pass the original `w ResponseWriter` down to the Middleware chain and to the final Handler. It's like passing `w` variable directly.

Thus, the `ww.Status()` and `ww.BytesWritten()` above won't work anymore.

VojtechVitek (changes requested):

middleware/logger.go
bernhardreisenberger: yes, you are right, i missed that. to fix my issue with the hijacker interface i have another solution: 

i saw, that the `NewWrapResponseWriter` function can return a httpFancyWriter. the fancywriter is only chosen, if the original response writer has a flusher, hijacker and readerFrom.
in my specific case, a flushWriter will be returned, which does not support hijack.

there are multiple options here. 
* i could add the hijack method to the flushwriter.
* i could make a new writer that supports hijack and flush
* i could remove the flushwriter and use the httpFancyWriter if one of the interfaces (flusher, hijacker, readerFrom) is supported. in wrap_writer.go line 28. `if fl && hj && rf `-> `if fl || hj || rf`

i would go for the third option, but i do not know if there is a specific intention for the flushWriter.

VojtechVitek: 3. sounds good to me too

bernhardreisenberger:
i will make a new pull request with the upcoming changes.

PR #562, opened 2020-12-11
logger: support websocket upgrade

bernhardreisenberger:
currently the logger middleware returns a WrapResponseWriter which does not support the http.Hijacker interface.

These changes are necessary, because i cannot upgrade to websocket without the hijacker interface. I am using https://github.com/gorilla/websocket to upgrade.

see #559 for previous pull request

### DIFF
--- a/middleware/logger_test.go
+++ b/middleware/logger_test.go
@@ -1,0 +1,34 @@
+package middleware
+
+import (
+	"bufio"
+	"net"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+)
+
+type testLoggerWriter struct {
+	*httptest.ResponseRecorder
+}
+
+func (cw testLoggerWriter) Hijack() (net.Conn, *bufio.ReadWriter, error) {
+	return nil, nil, nil
+}
+
+func TestRequestLogger(t *testing.T) {
+	testHandler := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		_, ok := w.(http.Hijacker)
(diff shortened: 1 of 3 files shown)


pkieltyka:
Thanks but why did you remove the flush writer? I believe it’s needed for chunked transfer. As well, I swore I used chi logger with gorilla websocket in the past — what is the end goal for the logger in your website you’re looking to achieve? To log from the websocket connection the different payloads?

bernhardreisenberger:
i want to upgrade a http connection to a websocket connection. using the `upgrader.Upgrade(w, r, nil)` method.
If i do not use the logger middleware, it works fine. But if i use the logger middleware, the upgrade method of the websocket.Upgrader returns an error with "websocket: response does not implement http.Hijacker"
see https://github.com/gorilla/websocket/blob/c3dd95aea9779669bb3daafbd84ee0530c8ce1c1/server.go#L173 for details.

while the original http.ResponseWriter supports the http.Hijacker interface, the wrapper used in the logger middleware did not, because the basicWriter does not support the http.Hijacker interface. i changed the NewWrapResponseWriter function to return an httpFancyWriter (this type supports all common interfaces).

in #559 i wrote with @VojtechVitek and mentioned to remove the flushWriter and replace it with the httpFancyWriter. He acknowledged and that is why i removed it.

pkieltyka:
ah, I forgot httpFancyWriter also implemented the Flush() method. This PR looks good then and should work -- thanks for the iteration

PR #567, opened 2020-12-24
Fix: validate empty path param

### LINKED ISSUE #551: Empty path param always passes regexp validation
Empty path param passes regexp validation regardless the regexp used.

Chi version `v4.1.2`

**Reproducer**
```
package main

import (
	"net/http"

	"github.com/go-chi/chi"
)

func main() {
	mux := chi.NewMux()
	mux.Get("/foo/{x:.+}/bar", func(w http.ResponseWriter, r *http.Request) {
		if _, err := w.Write([]byte("passed\n")); err != nil {
			panic(err)
		}
	})
	http.ListenAndServe(":8080", mux)
}
```
Execute `curl -v 'http://localhost:8080/foo//bar'`. 

**Expected**

- the 404 status code be returned without  executing handler fnc

**Actual**

- handler executed and `passed` string returned

**Notes**

- From my point of view the non-empty path param should be validated by default, even without specifying regexp. The regexp `.*` would have to be used to allow empty path param. This is also how [Gorilla mux](https://github.com/gorilla/mux) works and it's intuitive in my opinion.

matsuyoshi30:
Fix #551

### DIFF
--- a/mux_test.go
+++ b/mux_test.go
@@ -1452,7 +1452,7 @@ func TestMuxWildcardRouteCheckTwo(t *testing.T) {
 
 func TestMuxRegexp(t *testing.T) {
 	r := NewRouter()
-	r.Route("/{param:[0-9]+}/test", func(r Router) {
+	r.Route("/{param:[0-9]*}/test", func(r Router) {
 		r.Get("/", func(w http.ResponseWriter, r *http.Request) {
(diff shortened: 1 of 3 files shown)


pkieltyka:
Thanks, however a few test cases are failing

matsuyoshi30:
Sorry, I'll reconsider it soon.

PR #571, opened 2021-01-10
Ability to specify response HTTP status code for Throttle middleware

vasayxtx:
Some time ago the PR [#528](https://github.com/go-chi/chi/pull/528) was merged but seems like in some cases 503 HTTP code is more suitable. Actually, using 503 or 429 response code for rate/conn limiting is a controversial topic. As for me, 429 HTTP code should be used when a **specific** client is breaking a contract by making too many requests over a certain period of time. It will be a bit strange if the client receives 429 on its first request.

P.S.:
- [Good post](http://blog.cellfish.se/2013/12/429-vs-503.html) about it.
- At Nginx 503 is used by default for [rate](http://nginx.org/en/docs/http/ngx_http_limit_req_module.html#limit_req_status) and [conn](http://nginx.org/en/docs/http/ngx_http_limit_conn_module.html#limit_conn_status) limiting.

This PR provides the ability to specify the needed status code. Could you please consider it?

### DIFF
--- a/middleware/throttle.go
+++ b/middleware/throttle.go
@@ -22,6 +22,7 @@ type ThrottleOpts struct {
 	Limit          int
 	BacklogLimit   int
 	BacklogTimeout time.Duration
+	StatusCode     int
 }
 
 // Throttle is a middleware that limits number of currently processed requests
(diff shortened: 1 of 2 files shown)


pkieltyka:
thanks for the PR. I'll need to find some head space to think about this

senekis:
Any news on this PR?

Agree with @vasayxtx, allowing to custom the status code (eg. 503) is better than always returning 429. It is misleading for the clients of the service, if they receive a 429 error will think are reaching their limits and that modifying the pace will make the error go away, but no, they will continue getting the errors, so is more suitable in this case 503.

vasayxtx:
I've just resolved the conflict, so the PR is ready for merging:)
@pkieltyka JFYI

VojtechVitek (approved):
LGTM

PR #573, opened 2021-01-15
Fixing RegisterMethod method value generation

esachser:
- When registering the first New Method, de value generated was the same as mTRACE. So, some handler created to accept the "My new registered method" was accepting TRACE calls too.
- The "n" value must be lesser than IntSize - 2 to generate a fresh new value, otherwise returns 0.
- Changed code to use bit operations to generate the value

### DIFF
--- a/tree.go
+++ b/tree.go
@@ -6,7 +6,6 @@ package chi
 
 import (
 	"fmt"
-	"math"
 	"net/http"
 	"regexp"
 	"sort"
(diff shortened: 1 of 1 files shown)


VictorWiedemann:
The merge breaks all builds using go version 1.12 and older:

github.com/go-chi/chi/tree.go:60:20: invalid operation: 2 << n (shift count type int, must be unsigned integer)

pkieltyka:
@vasayxtx thanks for the note, out of curiousity, why are you using go 1.12 still..?

VictorWiedemann:
I am using a version of yocto (embedded linux) that is controlled at version go v1.12.6. I am planning on changing my end to grab a specific tag of go-chi/chi right now to fix my issue. I wasn't sure if you had any requirements for backward compatibility and wanted to raise this up.

pkieltyka:
cool thanks again for the feedback. I will think about some more on the best place to draw the line

VictorWiedemann:
Yeah no problem. I didn't realize by build was pulling the latest git commit and not a specific tag, so this actually pointed out an issue that could have been way worse if not caught. I will say for my case, we are using an older revision of golang because we don't want to upgrade all of our software in an already released product.

PR #579, opened 2021-02-06
use http.Status const

chemidy:
- Use http.Status const


pkieltyka:
Thank you for this, however I find the numbers to be more readable

Issue #576, opened 2021-01-27
TestMuxContextIsThreadSafe occasional crash


rantala:
`TestMuxContextIsThreadSafe` test is occasionally crashing:

```
$ go test
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x20 pc=0x6bddd0]

goroutine 173608 [running]:
github.com/go-chi/chi.(*directContext).Done(0xc00036c680, 0xad696809bbda0a01)
        /home/tomranta/git/chi/context.go:161 +0x30
context.parentCancelCtx(0x812580, 0xc00036c680, 0x0, 0x0)
        /usr/lib/golang/src/context/context.go:295 +0x35
context.removeChild(0x812580, 0xc00036c680, 0x80dee0, 0xc00059e300)
        /usr/lib/golang/src/context/context.go:314 +0x39
context.(*timerCtx).cancel(0xc00059e300, 0xc0002e1f01, 0x80c700, 0xa6a0c0)
        /usr/lib/golang/src/context/context.go:476 +0x141
context.WithDeadline.func2()
        /usr/lib/golang/src/context/context.go:446 +0x47
created by time.goFunc
        /usr/lib/golang/src/time/sleep.go:168 +0x44
exit status 2
FAIL    github.com/go-chi/chi   0.159s
```

Running with the race detector produces warnings:
```
$ go test -race
==================
WARNING: DATA RACE
Write at 0x00c00069eb50 by goroutine 93:
  github.com/go-chi/chi.(*Context).Reset()
      /home/tomranta/git/chi/context.go:95 +0x3a7
  github.com/go-chi/chi.(*Mux).ServeHTTP()
      /home/tomranta/git/chi/mux.go:78 +0x18a
  github.com/go-chi/chi.TestMuxContextIsThreadSafe.func2()
      /home/tomranta/git/chi/mux_test.go:1586 +0xdf

Previous read at 0x00c00069eb50 by goroutine 87:
  github.com/go-chi/chi.(*directContext).Value()
      /home/tomranta/git/chi/context.go:172 +0x85
  context.parentCancelCtx()
      /usr/lib/golang/src/context/context.go:299 +0xb8
  context.removeChild()
      /usr/lib/golang/src/context/context.go:314 +0x46
  context.(*timerCtx).cancel()
      /usr/lib/golang/src/context/context.go:476 +0x176
  context.WithDeadline.func2()
[...]

PR #555, opened 2020-11-26
Remove one allocation by foregoing context.WithValue

bouk:
Instead of we can make Context implement the context.Context interface directly through a private type, which just implements the Value method.

If you want to remove the last allocation you could also do `*r = *r.WithContext((*directContext)(rctx))`, although that would be modifying something you don't own.

### Before

```
~/s/g/g/chi (master) go test -bench B
goos: darwin
goarch: amd64
BenchmarkMux/route:/-12                  4089687               291 ns/op             448 B/op          3 allocs/op
BenchmarkMux/route:/hi-12                3795570               306 ns/op             448 B/op          3 allocs/op
BenchmarkMux/route:/sup/123/and/this-12                  2908492               412 ns/op             448 B/op          3 allocs/op
BenchmarkMux/route:/sup/123/foo/this-12                  2328763               519 ns/op             448 B/op          3 allocs/op
BenchmarkMux/route:/sharing/z/aBc-12                     2355368               509 ns/op             448 B/op          3 allocs/op
BenchmarkMux/route:/sharing/z/aBc/twitter-12             1925193               618 ns/op             456 B/op          4 allocs/op
BenchmarkMux/route:/sharing/z/aBc/direct-12              1654678               726 ns/op             456 B/op          4 allocs/op
BenchmarkMux/route:/sharing/z/aBc/direct/download-12             1445162               827 ns/op             480 B/op          5 allocs/op
BenchmarkTreeGet-12                                              8847862               133 ns/op               0 B/op          0 allocs/op
PASS
ok      _/Users/bouke/src/github.com/go-chi/chi 16.505s
```

### After

```
~/s/g/g/chi (master) go test -bench B
goos: darwin
goarch: amd64
BenchmarkMux/route:/-12                  4975675               238 ns/op             400 B/op          2 allocs/op
BenchmarkMux/route:/hi-12                4772744               255 ns/op             400 B/op          2 allocs/op
BenchmarkMux/route:/sup/123/and/this-12                  3353338               360 ns/op             400 B/op          2 allocs/op
BenchmarkMux/route:/sup/123/foo/this-12                  2503300               500 ns/op             400 B/op          2 allocs/op
BenchmarkMux/route:/sharing/z/aBc-12                     2226513               477 ns/op             400 B/op          2 allocs/op
BenchmarkMux/route:/sharing/z/aBc/twitter-12             2153860               553 ns/op             408 B/op          3 allocs/op
[...]

PR #581, opened 2021-02-10
revert directContext

pkieltyka:
I've done some testing, and I cannot produce any race condition "in practice", however, under `go test -race`, a race is flagged and so I've removed this community submission which offered a 1 allocation performance optimization, but also introduced this warning.

related to: https://github.com/go-chi/chi/issues/576 and https://github.com/go-chi/chi/pull/555


bouk:
Huh, interesting. I wonder why this happens!

VojtechVitek (approved):
LGTM

Issue #583, opened 2021-02-15
httptest.ResponseRecorder incompatible with Logger in chi v1.5.2+ (including v5.0.2)


zlogic:
In my app, I'm using the Logger middleware:

```
r := chi.NewRouter()

r.Use(middleware.RequestID)
r.Use(middleware.RealIP)
r.Use(middleware.Logger) // commenting this line resolves the issue
r.Use(middleware.Recoverer)

r.HandleFunc("/favicon.ico", func(w http.ResponseWriter, r *http.Request) {
	http.ServeFile(w, r, path.Join("static", "favicon.ico"))
})
```

and `httptest.ResponseRecorder` to test the Router:

```
req, _ := http.NewRequest("GET", "/favicon.ico", nil)
res := httptest.NewRecorder()

router.ServeHTTP(res, req)
```

It works without issues when using:
* chi v1.5.1, with the Logger enabled (and disabled)
* chi v1.5.2, with the Logger disabled

However when using chi v1.5.2 with the Logger enabled, I'm getting a panic when trying to run unit tests: 

```
 panic: interface conversion: *httptest.ResponseRecorder is not io.ReaderFrom: missing method ReadFrom
 
 -> github.com/go-chi/chi/middleware.(*httpFancyWriter).ReadFrom
 ->   /Users/zlogic/go/pkg/mod/github.com/go-chi/chi@v1.5.2/middleware/wrap_writer.go:135

    io.copyBuffer
      /usr/local/go/src/io/io.go:395
    io.Copy
      /usr/local/go/src/io/io.go:368
    io.CopyN
      /usr/local/go/src/io/io.go:344
    net/http.serveContent
      /usr/local/go/src/net/http/fs.go:300
    net/http.serveFile
      /usr/local/go/src/net/http/fs.go:627
    net/http.ServeFile
      /usr/local/go/src/net/http/fs.go:688
    github.com/zlogic/nanorss-go/server.FaviconHandler
      /Users/zlogic/Documents/Development/nanoRSS/Code/server/routehandlers.go:70
    net/http.HandlerFunc.ServeHTTP
      /usr/local/go/src/net/http/server.go:2042
    github.com/go-chi/chi.(*Mux).routeHTTP
      /Users/zlogic/go/pkg/mod/github.com/go-chi/chi@v1.5.2/mux.go:436
    net/http.HandlerFunc.ServeHTTP
      /usr/local/go/src/net/http/server.go:2042
[...]

PR #584, opened 2021-02-17
Fix support for httptest.ResponseRecorder

zlogic:
If the `http.ResponseWriter` doesn't implement `io.ReaderFrom`, use a fallback implementation.

This fixes https://github.com/go-chi/chi/issues/583 without breaking the changes from https://github.com/go-chi/chi/pull/562.

### DIFF
--- a/middleware/logger_test.go
+++ b/middleware/logger_test.go
@@ -2,10 +2,12 @@ package middleware
 
 import (
 	"bufio"
+	"bytes"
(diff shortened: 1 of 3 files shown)


pkieltyka:
Looks good, if anyone else is any to help test + review this, we can merge it

pkieltyka:
Actually, it looks good.. merging.

PR #586, opened 2021-02-19
update for go 1.16

pkieltyka:


go.mod
```diff
@@ -1,3 +1,32 @@
...
+	v3.1.4+incompatible
+	v3.1.3+incompatible
+	v3.1.2+incompatible
+	v3.1.1+incompatible
+	v3.1.0+incompatible
+	v3.0.0+incompatible
+	v2.1.0+incompatible
+	v2.0.0+incompatible
+	v1.0.0+incompatible
+	v0.9.0+incompatible
```
zikaeroh, line 31: These two don't need `+incompatible`, as they are before v2.

zikaeroh:
You may want to verify that you're retracting all of the versions you want gone from:

- https://proxy.golang.org/github.com/go-chi/chi/@v/list
- https://pkg.go.dev/github.com/go-chi/chi?tab=versions

As there appear to be a couple (the `-rc` one at least) that are missing.

PR #589, opened 2021-02-27
Memory alignment on structs

chemidy:
Memory alignment  on structs, can optimize memory. 

```
Using  vscode config : 
  "gopls": {
    "analyses": { "fieldalignment": true }
  },
```


pkieltyka:
Brilliant! thank you. I will merge this soon

Issue #561, opened 2020-12-07
chi@v1.5.x mod issues


umputun:
I have tried to switch to v1.5.0 from v4.1.2+incompatible but because the version number decreased it caused a downgrade of all other dependencies using chi. For example, I have started with this:

```
git.example.com/commons/pkg/rest v1.3.2
github.com/go-chi/chi v4.1.2+incompatible
github.com/go-chi/render v1.0.1
```

After running `go get github.com/go-chi/chi@v1.5.1` I got:

```
git.example.com/commons/pkg/rest v1.1.0
github.com/go-chi/chi v1.5.1
github.com/go-chi/render v1.0.1
```
As you can see my `pkg/rest` downgraded to some old version and such side effects can be very unexpected and unpleasant. 

Is there a reason why you decided to go with v1.5.x and not with a more traditional (and less dangerous) approach with v5 (also changing the package name to "/v5")? I'm not sure how to deal with this. In order to avoid those unexpected downgrades I have to have all of my dependencies migrated to v1.5, but unfortunately some of them not under my control.

pkieltyka:
Sorry for the trouble -- I realize this is an usual change, but unfortunately the versioning nuance pre-date go.mod's existence. For dependencies which themselves depend on chi, they will have to be updated to chi@v1.5.1 as well, in order to make use of the new go.mod version -- which is confusing by "numbers", but is the same in either case if a new v5 was released just the same.

As for the reasoning to switch back to v1, please see https://github.com/go-chi/chi/pull/560. 

Note, Go 1.16 will include a new feature for go.mod to mark certain versions as "retracted", which should smooth this out and allow you to upgrade then. TBH, there are not many urgent features in chi from v4+incompatible to v1 with go.mod. However, I've had no issues switching to chi v1.5.1 (go.mod) for all of my projects internally.

umputun:
thx, I just read the discussion in #462 and @zikaeroh summarized it nicely - "v1.5.1 is still semantically lower than any of the +incompatible versions. That's why I'm saying this is going to turn out to not be so fun to handle...". It is indeed not fun and even worse, because in my case an innocent-looking and minor update of the chi module triggered a chain reaction downgrading other dependencies, some of them significantly. 

I understand you made the call and it is what it is. However, I'd suggest documenting possible risks or at least mention in README how potentially destructive such a `go get -u github.com/go-chi/chi@v1.5.0` could be in case if it has a dependency which themselves depend on chi.

[...]

PR #593, opened 2021-02-28
Fix last set of missing v5 paths

zikaeroh:
This fixes the paths that were missed in https://github.com/go-chi/chi/issues/561#issuecomment-787403243.

These `_` prefixed folders are actually ignored by the Go tool when ran from an upper directory, so there's no imminent need. My comment was wrong, this is strictly for finishing the move.

Might be a good idea to add to the builder something that explicitly tries to build the files in `_examples`, to catch any errors.

### DIFF
--- a/_examples/versions/main.go
+++ b/_examples/versions/main.go
@@ -14,11 +14,11 @@ import (
 	"net/http"
 	"time"
 
-	"github.com/go-chi/chi/_examples/versions/data"
-	v1 "github.com/go-chi/chi/_examples/versions/presenter/v1"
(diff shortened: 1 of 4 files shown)


icholy:
@zikaeroh I've added a `path` subcommand to [gomajor](https://github.com/icholy/gomajor) to help with upgrades like this in the future.

``` sh
# increment the major version
$ gomajor path -next

# set the major version
$ gomajor path -version=v7

# change the whole module path (if you decide to move the repo to avoid SIV)
$ gomajor path chisite.io/chi
```

PR #596, opened 2021-03-06
optimize sructs with memory alignment / padding

chemidy:
Another small optimizations on structs.

- middleware/compress.go :
  - struct with 48 pointer bytes could be 40
  - struct with 64 pointer bytes could be 56
- middleware/route_headers.go : struct with 72 pointer bytes could be 56
- mux.go : struct with 88 pointer bytes could be 64
- tree.go : struct with 136 pointer bytes could be 128

### DIFF
--- a/middleware/compress.go
+++ b/middleware/compress.go
@@ -45,7 +45,6 @@ func Compress(level int, types ...string) func(next http.Handler) http.Handler {
 
 // Compressor represents a set of encoding configurations.
 type Compressor struct {
-	level int // The compression level.
 	// The mapping of encoder names to encoder functions.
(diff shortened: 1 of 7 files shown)


vtolstov:
do yo have some go binary that check alignment and display needed changes? or tool is only says about misaligning and numbers?

chemidy:
Sure : i'm using gopls experimental Field alignment analyzer
:  https://github.com/golang/tools/releases/tag/gopls/v0.6.0#Experimental

and enabled it on vscode

```
  "gopls": {
    "analyses": { "fieldalignment": true }
  }
```

Issue #536, opened 2020-08-29
RealIP allows any value, including non-IP addresses and private IP addresses


arp242:
The `RealIP` accept any value. Since clients can set this themselves this is perhaps not a good idea; for example:

    X-Forwarded-For: 192.168.1.1
    X-Forwarded-For: -- `some injected malicious value?`

It's perhaps better to 1) lock this down to just IP addresses, and 2) not allow non-public IP addresses?

This rolled out of a security evaluation of my product. For what it's worth, here's the code I'm using now: https://github.com/zgoat/zhttp/blob/master/realip.go#L18

cilindrox:
from [the docs](https://pkg.go.dev/github.com/go-chi/chi/middleware#RealIP): 

> You should only use this middleware if you can trust the headers passed to you (in particular, the two headers this middleware uses), for example because you have placed a reverse proxy like HAProxy or nginx in front of Goji. If your reverse proxies are configured to pass along arbitrary header values from the client, or if you use this middleware without a reverse proxy, malicious clients will be able to make you very sad (or, depending on how you're using RemoteAddr, vulnerable to an attack of some sort).

Seems like this is expecting a reverse proxy sitting in front of the app, and that's where the actual filtering should occur

arp242:
You may not be running a proxy but clients can still set this header.

cilindrox:
Precisely, in those scenarios (no proxy, untrusted network) then I’d argue you shouldn’t be trusting arbitrary user input, which is what the docs mention

arp242:
It's worth remembering that everything sent over the network is "arbitrary user input"; you can fake the RemoteAddr value too if you want, it's just a little bit more effort. There is never any guarantee that you're getting the "real" IP address.

It's not a problem if I fake this to 1.2.3.4, since that's just a random IP address. But 127.0.0.1 or 10.0.0.5 is a local address or address in the local network and may, in some situations, lead to some sort of abuse, so it's probably best to filter those out while we're processing these headers anyway. Come to think of it, I should probably change my implementation to account for a faked RemoteAddr.

At any rate, the practical problem is that I want to ship an app people can run with or without a proxy, depending on what works best for them, which seems a common use case.

pkieltyka:
[...]

PR #614, opened 2021-04-21
Correct docs regarding X-Real-IP/X-Forwarded-For precedence

JohnMaguire:
It's important to be clear about the order - as an example, AWS ALBs set X-Forwarded-For. If you as a developer assume the docs are correct, it may be possible for a user to spoof their IP address by setting an X-Real-IP header.

Added a test case demonstrating the precedence.

Some discussion here: https://github.com/go-chi/chi/issues/536#issuecomment-811615534


JohnMaguire:
Test failure appears unrelated.

PR #624, opened 2021-05-29
recoverer: don't recover http.ErrAbortHandler

drakkan:
This error is generally used to abort a request while streaming a response
so it should not be recovered otherwise the request is not aborted and
the client does not detect the error.

In my use case I'm streaming a zip file directly to the client (without writing it to the disk). If an error happen while streaming the request body I return `http.ErrAbortHandler` to abort the request without logging a panic. This is the only way to report an error to the client/browser. So this error should not be recovered otherwise the request is not aborted and the client does not detect the error.

My use case is quite similar to the one described [here](https://github.com/golang/go/issues/23643)

### DIFF
--- a/middleware/recoverer.go
+++ b/middleware/recoverer.go
@@ -22,7 +22,12 @@ import (
 func Recoverer(next http.Handler) http.Handler {
 	fn := func(w http.ResponseWriter, r *http.Request) {
 		defer func() {
-			if rvr := recover(); rvr != nil && rvr != http.ErrAbortHandler {
+			if rvr := recover(); rvr != nil {
+				if rvr == http.ErrAbortHandler {
+					// we don't recover http.ErrAbortHandler so the response
+					// to the client is aborted, this should not be logged
+					panic(rvr)
+				}
 
 				logEntry := GetLogEntry(r)
 				if logEntry != nil {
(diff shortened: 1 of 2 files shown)


middleware/recoverer.go
```diff
@@ -21,7 +21,12 @@ import (
 func Recoverer(next http.Handler) http.Handler {
 	fn := func(w http.ResponseWriter, r *http.Request) {
 		defer func() {
-			if rvr := recover(); rvr != nil && rvr != http.ErrAbortHandler {
+			if rvr := recover(); rvr != nil {
+				if rvr == http.ErrAbortHandler {
```
pkieltyka, line 25: what difference does this provide to move the `rvr` condition testing `http.ErrAbortHandler` below..? I'm wondering how this solves your issue? thanks for explaining

drakkan: It solves the issue because before this change the panic is recovered and so the http connection is not aborted (the recoverer sends a 500 status code). 

After this change we have:

```
if rvr == http.ErrAbortHandler {
	// we don't recover http.ErrAbortHandler so the response
	// to the client is aborted
	panic(rvr)
}
```

and so if the error is `http.ErrAbortHandler` the panic is not recovered anymore.

PR #625, opened 2021-05-30
fix some warnings found using staticcheck

drakkan:
Fixed these warnings:

unnecessary guard around call to delete (S1033)
when ok is true, encoder can't be nil (S1020)

There are also unused variables like nBlack, nBlue etc. I suppose we
want to keep them

### DIFF
--- a/middleware/strip.go
+++ b/middleware/strip.go
@@ -53,7 +53,7 @@ func RedirectSlashes(next http.Handler) http.Handler {
 				path = path[:len(path)-1]
 			}
 			redirectURL := fmt.Sprintf("//%s%s", r.Host, path)
-			http.Redirect(w, r, redirectURL, 301)
+			http.Redirect(w, r, redirectURL, http.StatusMovedPermanently)
 			return
 		}
 		next.ServeHTTP(w, r)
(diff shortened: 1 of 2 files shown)


middleware/strip.go
```diff
@@ -53,7 +53,7 @@ func RedirectSlashes(next http.Handler) http.Handler {
 				path = path[:len(path)-1]
 			}
 			redirectURL := fmt.Sprintf("//%s%s", r.Host, path)
-			http.Redirect(w, r, redirectURL, 301)
+			http.Redirect(w, r, redirectURL, http.StatusMovedPermanently)
```
mfridman, line 56: Outside of being more verbose and descriptive, I don't see the benefit of using this over the value directly.

tbh I find the value 301 more readable and its _very_ unlikely these status codes will change.

drakkan: staticcheck suggests changing as you see above, I understand this might be moot, it depends if we want to trust this tool

dominikh: FWIW, the check has an [option](https://staticcheck.io/docs/options#http_status_code_whitelist) of which numerics to allow directly, and it defaults to what I perceived as codes everybody knows (200, 400, 404, 500). For all other status codes, it recommends using the constants instead to aid readability, assuming that people are more likely to understand "moved permanently" than "301".

You can either [add numerics to this option](https://staticcheck.io/docs/configuration#configuration-files), or disable the entirety of ST1013 if you disagree with the check as a whole.

mfridman: ye that's fair, please ignore my comment. I interjected where I shouldn't have, and this comes down to the maintainer's preference.

pkieltyka:
I prefer the code as is, thank you though

PR #632, opened 2021-06-27
Allow regex url paths ending which includes slashes

pcdummy:
This PR allows paths like (it makes the second one possible):

```go
r.Get("/v0/{path:[a-zA-Z0-9-/]+}.{format:(json|xml|html)+}", a.getTreeObject)
r.Get("/v0/{path:[a-zA-Z0-9-/]+}", a.getTreeObject)
```

witch matche URL's like:
```yaml
urls:
  -  /v0/lists/com/github/go-chi/chi/v5
  -  /v0/lists/com/github/go-chi/chi/v5.json
```

While this works for me (tm) and i ran the test cases, idk it's the right way.

Kind Regards,
René

Signed-off-by: René Jochum <rene@jochum.dev>

### DIFF
--- a/tree.go
+++ b/tree.go
@@ -423,6 +423,14 @@ func (n *node) findRoute(rctx *Context, method methodTyp, path string) *node {
 				// label for param nodes is the delimiter byte
 				p := strings.IndexByte(xsearch, xn.tail)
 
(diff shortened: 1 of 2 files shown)


pcdummy:
@pkieltyka wanna have a look?

PR #640, opened 2021-07-16
middleware: fix set Vary header

mytheta:
There are cases where you want to define a value(ex. origin, accept ...) other than accept-encoding in the Vary header.
So, if the Vary header is already set, why not give priority to the existing settings?


davidspiess:
Any chance to have this merged? We ran in the same issue.

PR #645, opened 2021-08-19
Docsify Based Markdown Documentation

Shubhaankar-Sharma:
### Preview :- [Docs](https://shubhaankar-sharma.github.io/chi)
Let me know what to change...
This is supposed to be served with github pages, where /docs is the directory where index.html is placed


VojtechVitek:
Lovely!

@pkieltyka Do we want to create a separate repo for the docs? Or perhaps we could use a dedicated branch for the docs?

Shubhaankar-Sharma:
Yea I think we can do a orphan branch for docs as well...

pkieltyka:
@VojtechVitek @Shubhaankar-Sharma what if we merged to master under “docs/“ folder? btw I also registered go-chi.io for us to put the docs site on :)

Shubhaankar-Sharma:
@pkieltyka yes thats what i originally planned... so now the import path will also be go-chi.io/chi/v5 ? awesome

pkieltyka:
I’m going to keep the import path the same for now, but in the future will consider a vanity import path. For now go-chi.io will just be for the docs site

Shubhaankar-Sharma:
ok cool 👍 then we should just do in /docs in my opinion...

docs/user_guide/testing.md
```diff
@@ -0,0 +1,116 @@
...
+	rr := httptest.NewRecorder()
+	s.Router.ServeHTTP(rr, req)
+
+	return rr
+}
+
+// checkResponseCode is a simple utility to check the response code
+// of the response
+func checkResponseCode(t *testing.T, expected, actual int) {
+	if expected != actual {
```
robsonpeixoto, line 89: Please, add `t.Helper()`

Shubhaankar-Sharma: Will do 👍

PR #654, opened 2021-08-30
Fix middleware.Recoverer under Go 1.17+ (again)

subtle-byte:
In Go 1.17 the format of the `debug.Stack()` was changed. In Go 1.16 we could see the following fragment:
```console
panic(0x67b480, 0x72a100)
``` 
But in Go 1.17 we have:
```console
panic({0x6d8bc0, 0x796560})
```
I.e. now there are curly brackets.

How does this show up while using `middleware.Recoverer`? With Go 1.16 I used to see something like:

image

But now with Go 1.17 I see:

image
I.e. now I see the internals.

How does this pull request fix that? By replacing the code fragment `strings.HasPrefix(stack[i], "panic(0x")` with `strings.HasPrefix(stack[i], "panic(")`.

### DIFF
--- a/middleware/recoverer.go
+++ b/middleware/recoverer.go
@@ -7,6 +7,7 @@ import (
 	"bytes"
 	"errors"
 	"fmt"
+	"io"
 	"net/http"
 	"os"
 	"runtime/debug"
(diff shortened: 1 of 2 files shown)


pkieltyka:
Can we ensure this will behave correctly on Go < 1.16 and also Go 1.17?

subtle-byte:
@pkieltyka

Obviously, any line that had `panic(0x` prefix, will have `panic(` prefix. So previously matching lines will continue to match.

On other hand, it seems like the line with the prefix `panic(` is quite specific to we can say it will continue with either `0x` (in Go <= 1.16) or `{0x` (in Go 1.17). So no new matches for Go <= 1.16.

So behavior does not change for Go <= 1.16 but at the same time adapts to the Go 1.17 changes.

middleware/recoverer.go
```diff
@@ -40,12 +41,14 @@ func Recoverer(next http.Handler) http.Handler {
 	return http.HandlerFunc(fn)
 }
 
+var recovererErrorWriter io.Writer = os.Stderr
```
pkieltyka, line 44: can you leave comment above this stating why you added this pkg-level variable? (as it appears for ability to test the PrintPrettyStack function, but its helpful for other reasons to know its rationale.)

subtle-byte: @pkieltyka Done

PR #659, opened 2021-09-15
add tests to match emails in simple patterns

butonic:
I was expecting to be able to match percent encoded URL parameters, e.g. in email adresses. The percent encoded URL is not matched however. I know I can change the pattern, but is that expected?


butonic:
The error we saw was caused by shift path routing that was only shifting the URL.Path, not the URL.RawPath before handing it over to chi. We now unset the URL.RawPath and this works as expected. Sorry for the noise.

PR #675, opened 2021-11-08
Correction in comment

### LINKED ISSUE #662: Small typo in comment
I think there is a small typo in a comment [here](https://github.com/go-chi/chi/blob/master/middleware/compress.go#L117-L118).

I think it should be "That's why we prefer gzip over deflate. It's just more reliable and not significantly slower than **deflate**" instead of "That's why we prefer gzip over deflate. It's just more reliable and not significantly slower than **gzip**".

Swastyy:
Fixes #662


VojtechVitek (approved):
LGTM

Swastyy:
> LGTM

Thanks @VojtechVitek

PR #677, opened 2021-11-12
Fix comment location

lord63:
The comment for `parent` should be for `inline`.

Checked the original comment from commit: c54b122dfe60b902d8f27b5c75ed2fcae97439af

### DIFF
--- a/mux.go
+++ b/mux.go
@@ -29,8 +29,8 @@ type Mux struct {
 	// Custom method not allowed handler
 	methodNotAllowedHandler http.HandlerFunc
 
-	// Controls the behaviour of middleware chain generation when a mux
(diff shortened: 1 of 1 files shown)


lord63:
By the way, feel confused when see:

```
var _ Router = &Mux{}
```

in https://github.com/go-chi/chi/blob/master/mux.go#L11

To make lint happy?

lord63:
ping @pkieltyka

VojtechVitek:
`var _ Router = &Mux{}` ensures that Mux implements Router interface in the build time. This is idiomatic Go code.

@lord63 This PR is nothing urgent. Please, be patient ;)

PR #684, opened 2021-12-02
support ip chain without spaces in xForwarded header

evgenyKharitonov:
This change covers a use case in `RealIP` middleware, when the chained IPs in `X-Forwarded-For` header are without space after comma.
This change is backwards compatible.
Test updated to cover this use case as well.


pkieltyka:
lol great catch, I dont know how we didnt catch it sooner. Should we also add a strings.Trim() anywhere..?

evgenyKharitonov:
I don't think so, since it's getting the first IP in the chain.

PR #700, opened 2022-01-17
test(middleware): Improves test coverage

nszilard:
This PR just aims to add some more code coverage for the middleware package. 

* Before

    `ok  	github.com/go-chi/chi/v5/middleware	25.464s	coverage: 61.0% of statements`

* After

    `ok  	github.com/go-chi/chi/v5/middleware	25.455s	coverage: 67.2% of statements`

### DIFF
--- a/middleware/basic_auth_test.go
+++ b/middleware/basic_auth_test.go
@@ -1,0 +1,71 @@
+package middleware
+
+import (
+	"net/http"
+	"net/http/httptest"
+	"testing"
+
+	chi "github.com/go-chi/chi/v5"
+)
+
+var testAuthCreds = map[string]string{
+	"testUser": "testPassword",
+}
+
+func TestBasicAuth(t *testing.T) {
+	r := chi.NewRouter()
+	r.Use(BasicAuth("localhost", testAuthCreds))
+	r.Get("/secure", http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		w.Write([]byte("authentication accepted"))
+	}))
+
+	cases := []struct {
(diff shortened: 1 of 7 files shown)


VojtechVitek (changes requested):
Thanks for adding these — the middleware package genuinely needed this coverage, and most of the tests here are solid. `TestBasicAuth`, `TestCleanPath`, `TestPathRewrite`, and `TestWithValue` all verify real behavior, and the branch still applies cleanly on today's master. 👍

Two tests pass even with the `r.Use(...)` line deleted (I verified by mutation-testing them), so they don't actually exercise the middleware yet:

1. `TestHeartbeat` — the test registers its own `/ping` handler, which returns 200 on its own. Drop that handler and assert the body is `"."` and `Content-Type: text/plain` instead.
2. `TestNew` — the `/` handler returns 200 anyway. Add a second request without `X-Custom-Header` and assert it gets 401.

Minor, take or leave:
- `terminal_test.go`: `for name, c := range cases` — `name` is the slice index; the struct's `name` field goes unused.
- The BasicAuth error messages say "bearer token" — these are basic auth credentials.

With those two tests fixed this is good to merge. Happy to push a fixup commit on top if that's easier.

VojtechVitek (approved):
Thanks for the quick turnaround, especially on a PR this old! 🙌

Verified the update: both `TestHeartbeat` and `TestNew` now fail when the middleware is removed, which is exactly what we wanted — and the `http.NotFoundHandler()` fallback in the heartbeat test is a clean way to do it. Nits are addressed too. LGTM 👍

PR #705, opened 2022-02-14
Remove unsed mux test structs

0daryo:
Hi team.
I just removed some structs which is not used in anywhere.

best regards.


pkieltyka:
thank you :D

PR #718, opened 2022-04-04
Use RoutePath in URLFormat middleware

Kazetsukai:
Currently the URLFormat middleware is resetting the RoutePath to the full URL when used in a sub-router. This change makes it look at the RoutePath, if it exists, and uses that instead of the full URL.

### DIFF
--- a/middleware/url_format.go
+++ b/middleware/url_format.go
@@ -51,6 +51,11 @@ func URLFormat(next http.Handler) http.Handler {
 		var format string
 		path := r.URL.Path
 
+		rctx := chi.RouteContext(r.Context())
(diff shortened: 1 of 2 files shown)


Kazetsukai:
Without this, sub routers break when adding URLFormat middleware. A possible workaround would be to just always use URLFormat at the top level router before any RoutePath is set, but being able to apply this middleware only to a subset of routes is nice.

Kazetsukai:
Without this, sub routers break when adding URLFormat middleware. A possible workaround would be to just always use URLFormat at the top level router before any RoutePath is set, but being able to apply this middleware only to a subset of routes is nice.

PR #720, opened 2022-04-11
Add a check to prevent modifying outer scope routers

polyfloyd:
This patch adds a panic that triggers when the build function passed to
(*Mux).Route() or (*Mux).Group() attempts to set routes or middleware on a
router is a parent of its argument router.

Example:

```
root := chi.NewRouter()
root.Route("/api", func(api chi.Router) {
    root.Use(DatabaseTransaction)
    // ...
})
root.Get("/health-check", HealthCheck)
```

This reads as if the DatabaseTransaction was intended to be used as middleware
on the api sub-router but it is actually used on the top-level router, making
it run in the hot path of the /health-check handler.

Adding this check will catch such mistakes.

### DIFF
--- a/mux.go
+++ b/mux.go
@@ -45,6 +45,11 @@ type Mux struct {
 	// Controls the behaviour of middleware chain generation when a mux
 	// is registered as an inline group inside another mux.
 	inline bool
+
+	// This bit is set while Route() or Group() is descending into their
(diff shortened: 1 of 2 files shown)


pkieltyka:
thanks for the submission, but not thinking to add such validation as the example you posted could actually be intended in some cases

PR #723, opened 2022-04-23
Added Bearer Token Authentication

dschemp:
Hey,

I've added a simple middleware analog to BasicAuth for a Bearer Token in the Authentication header.

This can be used like this:
```go
r.Use(middleware.BearerToken("testing123"))
// or
r.Get("/{filename}", handler.GetFile)
r.Head("/{filename}", handler.HeadFile)
r.With(middleware.BearerToken(Token())).Put("/{filename}", handler.PutFile)
r.With(middleware.BearerToken(Token())).Delete("/{filename}", handler.DeleteFile)
```

Currently only a single token can be passed. If it seems necessary or useful the method can be changed to accept multiple tokens.

### DIFF
--- a/middleware/bearer_token.go
+++ b/middleware/bearer_token.go
@@ -1,0 +1,27 @@
+package middleware
+
+import (
+	"fmt"
+	"net/http"
+)
+
+// BearerToken implements a simple middleware handler for adding bearer token auth to a route.
+func BearerToken(token string) func(http.Handler) http.Handler {
+	return func(next http.Handler) http.Handler {
+		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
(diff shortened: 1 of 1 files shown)


VojtechVitek:
I don't think this belongs in go-chi core library. This looks like a custom middleware you want to maintain in your codebase.

Afaik, people are mostly using JWT middleware https://github.com/go-chi/jwtauth instead of a static BEARER token string. Or they validate the token/session against their database.

pkieltyka:
It's interesting, but its too simple as it is and the bearer token check should happen from a function and not as a single expected value. As Vojtech mentioned though, using jwtauth is likely the better intended use.

dschemp:
I didn't notice that. If you want, you can dismiss this PR.
JWT seems to be the better option.

PR #726, opened 2022-05-13
Check for NO_COLOR in default logging middleware

### LINKED ISSUE #724: Support NO_COLOR
Would you be open to a simple `os.LookupEnv` for `middleware.Logger` default that supports [NO_COLOR](https://no-color.org/)?
I'd be happy to send in a PR if so.

jolheiser:
Resolves #724

As noted on https://no-color.org:

> Command-line software which adds ANSI color to its output by default should check for the presence of a `NO_COLOR` environment variable that, when present (regardless of its value), prevents the addition of ANSI color.

So I've opted for `os.LookupEnv` to just check for presence.

-----

Also of note, chi isn't necessarily "command-line software" on its own, however the check is fairly simple so I figured it could still be of use.

### DIFF
--- a/middleware/logger.go
+++ b/middleware/logger.go
@@ -166,7 +166,8 @@ func (l *defaultLogEntry) Panic(v interface{}, stack []byte) {
 }
 
 func init() {
-	color := true
+	_, noColor := os.LookupEnv("NO_COLOR")
+	color := !noColor
 	if runtime.GOOS == "windows" {
 		color = false
 	}


programmer04 (approved):
🆒

vbetsun (approved):

jasonwbarnett:
Why was this closed? This seems like a cool interface for running things in containers, etc.

VojtechVitek:
I think you can still override `middleware.DefaultLogger` from your code instead of introducing a new env var.

https://github.com/jolheiser/chi/blob/master/middleware/logger.go#L173

```go
middleware.DefaultLogger = middleware.RequestLogger(&middleware.DefaultLogFormatter{Logger: log.New(os.Stdout, "", log.LstdFlags), NoColor: true})
```

PR #735, opened 2022-06-26
middleware: fix broken comment on go doc

catatsuy:
Example code is broken on go doc.
https://pkg.go.dev/github.com/go-chi/chi/middleware#Logger
image


catatsuy:
Thank you!

PR #757, opened 2022-10-27
refactor(tree): clean code

aniaan:
1. make the node of `ntParam` have the correct prefix
2. clean up the code and make it more readable

### DIFF
--- a/tree.go
+++ b/tree.go
@@ -136,7 +136,7 @@ func (n *node) InsertRoute(method methodTyp, pattern string, handler http.Handle
 
 		// We're going to be searching for a wild node next,
 		// in this case, we need to get the tail
-		var label = search[0]
+		label := search[0]
 		var segTail byte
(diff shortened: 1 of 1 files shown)


pkieltyka:
thanks for the PR -- for (1) does this fix any particular issues / cases you've identified?

aniaan:
> thanks for the PR -- for (1) does this fix any particular issues / cases you've identified?

No, only the node related to `ntRegexp` uses the `prefix` field during the whole run, the node of `ntParam` does not use the `prefix` field, so whether the `ntParam` node has the correct `prefix` field or not does not affect the correctness of the program.
But in terms of the positioning of the `prefix` field, having the `ntParam` node have the correct `prefix` field will help to understand the code better.

The above issues are all some nuisances in reading the code, and if fixed, the readability of the code would be a little better.

Issue #320, opened 2018-05-14
Feature Request: Add URL param/query helpers


lansana:
In Chi, there is a helper to get the URL param. It has this signature:

```go
func Param(r *http.Request, key string) string
```

I found I would often have to import `strconv` and convert a number param to an int. So I forked and created a method with a signature like so:

```go
func ParamInt(r *http.Request, key string) (int, error)
```

Then I came across an issue with query parameters. I want to get a query parameter by name without having to parse it manually on each endpoint, and they also might be numbers that should be converted to ints.

So I created these two signatures as well:

```go
func QueryString(r *http.Request, key string) string
func QueryStringInt(r *http.Request, key string) (int, error)
```

Here is the logic for everything:

context.go:
```go
// Param returns the url parameter from a http.Request object.
func Param(r *http.Request, key string) string {
	if rctx := RouteContext(r.Context()); rctx != nil {
		return rctx.Param(key)
	}
	return ""
}

// ParamInt will get a param from the request URL and attempt to parse it as an int.
func ParamInt(r *http.Request, key string) (int, error) {
	val, err := strconv.Atoi(Param(r, key))
	if err != nil {
		return 0, err
	}
	return val, nil
}

// Query will get a query parameter by key.
func QueryString(r *http.Request, key string) string {
	if rctx := RouteContext(r.Context()); rctx != nil {
		return rctx.QueryString(key)
	}
	return ""
}

// QueryInt will get a query parameter by key and convert it to an int or return an error.
func QueryStringInt(r *http.Request, key string) (int, error) {
	val, err := strconv.Atoi(QueryString(r, key))
	if err != nil {
		return 0, err
	}
	return val, nil
}
```

Would you be willing to implement this so I don't need to use a forked version? :P

VojtechVitek:
This is indeed handy and we use something similar internally. I think this could be a nice standalone package, ie. https://github.com/go-chi/param.

```
i, _ := param.Int64(r, "id")

str, _ := param.String(r, "hash")
```

Thoughts?

VojtechVitek:
yeah, it can rely on chi's context, why not

lansana:
@VojtechVitek How would I go about creating the `param` package? I've created it on my end, would you or anyone else be able to create the param package for go-chi and merge in my changes in my own repository? Haven't done this sort of integration before.

Thanks!

VojtechVitek:
@lansana pls go ahead and hack on it on your Github profile

[...]

PR #760, opened 2022-11-09
add an example of getting http request parameters

ggicci:
Updates #320 

Here's an example of using [httpin](https://github.com/ggicci/httpin) with [chi](https://github.com/go-chi/chi).

Refer to httpin's integrations doc: https://ggicci.github.io/httpin/integrations/gochi.


ggicci:
Precisely speaking it's not. httpin decodes an http request into a struct, including querystring, while go-querystring encodes a struct to URL querystring.

pkieltyka:
thanks for sharing, but for the chi project and _examples in this repo, I prefer to keep it using zero external dependencies.

However, I did create a new repo, https://github.com/go-chi/community-examples which you can submit the PR over there :)

PR #772, opened 2022-12-18
Use timestamps in GMT format per RFC2616

falentio:
it seems http.TimeFormat used to format time that related with http header, and nocache middleware was not use that(time.RFC1123 instead)

### DIFF
--- a/middleware/nocache.go
+++ b/middleware/nocache.go
@@ -9,7 +9,7 @@ import (
 )
 
 // Unix epoch time
-var epoch = time.Unix(0, 0).Format(http.TimeFormat)
+var epoch = time.Unix(0, 0).UTC().Format(http.TimeFormat)
 
 // Taken from https://github.com/mytrile/nocache
 var noCacheHeaders = map[string]string{
--- a/middleware/nocache.go
+++ b/middleware/nocache.go
@@ -9,7 +9,7 @@ import (
 )
 
 // Unix epoch time
-var epoch = time.Unix(0, 0).Format(time.RFC1123)
+var epoch = time.Unix(0, 0).Format(http.TimeFormat)
 
 // Taken from https://github.com/mytrile/nocache
 var noCacheHeaders = map[string]string{


middleware/nocache.go
```diff
@@ -9,7 +9,7 @@ import (
 )
 
 // Unix epoch time
-var epoch = time.Unix(0, 0).Format(time.RFC1123)
+var epoch = time.Unix(0, 0).Format(http.TimeFormat)
```
VojtechVitek, line 12: https://pkg.go.dev/net/http#TimeFormat
"The time being formatted must be in UTC for Format to generate the correct format."

This is not correct

VojtechVitek (changes requested):
Is there any benefit to switching to a GMT (UTC) format? Can you point us at any RFC?

middleware/nocache.go
falentio: sorry my bad, fixed in recent commit

falentio:
here is the rfc that all http date must be represented in GMT
https://www.rfc-editor.org/rfc/rfc2616#section-3.3

falentio:
also [here](https://www.rfc-editor.org/rfc/rfc7231#section-7.1.1.1) , GMT in format was represent the time was UTC

VojtechVitek:
```
- All HTTP date/time stamps MUST be represented in Greenwich Mean Time
   (GMT), without exception. For the purposes of HTTP, GMT is exactly
   equal to UTC (Coordinated Universal Time). This is indicated in the
   first two formats by the inclusion of "GMT" as the three-letter
   abbreviation for time zone, and MUST be assumed when reading the
   asctime format. HTTP-date is case sensitive and MUST NOT include
   additional LWS beyond that specifically included as SP in the
   grammar.

- All expiration-related calculations MUST be done in GMT. The
        local time zone MUST NOT influence the calculation or comparison
        of an age or expiration time.

- If an HTTP header incorrectly carries a date value with a time
        zone other than GMT, it MUST be converted into GMT using the
        most conservative possible conversion.
```

VojtechVitek (approved):
LGTM

Issue #773, opened 2022-12-20
Support of pgzip


bronger:
In my current project, the response throughput is CPU-limited because gzip only works on one core. I suggest to make it possible to have pgzip do the compression work if the content type “gzip” is requested. According to https://github.com/go-chi/chi/pull/21/files, this has already been discussed.

VojtechVitek:
I think you can do this by setting your own encoder (pgzip) for "gzip" enconding:

https://github.com/go-chi/chi/blob/master/middleware/compress.go#L138-L148

bronger:
Thank you, it worked!  For anybody searching for “pgzip” in the issues, here is how I did it:

```go
import "github.com/klauspost/pgzip"

…

func pgzipCompress(level int, types ...string) func(next http.Handler) http.Handler {
        compressor := middleware.NewCompressor(level, types...)
        compressor.SetEncoder("gzip", func(w io.Writer, level int) io.Writer {
                writer, err := pgzip.NewWriterLevel(w, level)
                if err != nil {
                        panic(err)
                }
                return writer
        })
        return compressor.Handler
}

func main() {
        r := chi.NewRouter()
        r.Use(
                middleware.Logger,
                pgzipCompress(5, "text/xml"),
        )
…
```

As a side note, I think in the code excerpt you cited, `func(w http.ResponseWriter, level int)` must be replaced with `func(w io.Writer, level int)`, otherwiese, my Go 1.19 compiler complains.

VojtechVitek:
@bronger Cool! Can you please update the function signature in that comment then?

Also, it might make sense to add this as an "compress" example to https://github.com/go-chi/chi/tree/master/_examples if you're up to it.

PR #775, opened 2022-12-21
Correct doc comment

bronger:
The function must be of type middleware.EncoderFunc, and this has an io.Writer as its first argument. See https://github.com/go-chi/chi/issues/773#issuecomment-1360940386.


VojtechVitek (approved):
LGTM

PR #776, opened 2022-12-24
middleware: SetAllowHeader adds the Allow header on 405 response

### LINKED ISSUE #446: Include Allow header in 405 Method Not Allowed response

EwenQuim:
Makes chi compliant with [rfc9110](https://httpwg.org/specs/rfc9110.html#field.allow), because this behaviour is not currently supported in Chi.

It adds a small but non-negligable performance offset, that's why it is proposed as a **middleware**

Closes #446 

Don't hesitate to ask for changes. Thanks for this amazing package :) 

---

Performance details

```
BenchmarkSetAllowHeaderWhen405-8         1509776               776.7 ns/op           483 B/op          2 allocs/op
BenchmarkSetAllowHeaderWhen200-8         5560737               209.7 ns/op           336 B/op          3 allocs/op
BenchmarkWithoutSetAllowHeader-8         6710652               163.7 ns/op           332 B/op          3 allocs/op
```

- `BenchmarkWithoutSetAllowHeader` is without the middleware, to compare
- `BenchmarkSetAllowHeaderWhen200` is when the middleware is activated but not really used - it's a 200 response code
- `BenchmarkSetAllowHeaderWhen405` is  when the middleware is activated and used


EwenQuim:
@pkieltyka Any news about this ?

thexpand:
Why is this not merged? Is it still pending on a PR review or is it something else?

moutafatin1:
Any updates regarding this matter?

VojtechVitek (commented):
Hi,

I think the chi router itself should be able to set the `Allow` header along with the HTTP 405 responses. If we do this only on HTTP 405s, we wouldn't introduce any performance impact in the "happy path" unlike this middleware.

However, I think you've done a great job on this middleware. Could you share it as a standalone package with the community?

Best,
Vojtech

EwenQuim:
I really believe that it is possible to add this important feature without performance impact, and I am actively working on it. I have a working prototype that has **no (really 0ms/0byte) performance impact** on the happy path for the default `methodNotAllowed` handler, with an impact in case of 405 similar to what I got for the middleware.

I am refactoring it to try to keep the "chi" spirit (simple, light, idiomatic and maintainable) so that it is easy to review and rework later. I'll send it when I am satisfied with it. Thank you for your review and comment.

mux.go
```diff
@@ -378,11 +378,11 @@ func (mx *Mux) NotFoundHandler() http.HandlerFunc {
 
 // MethodNotAllowedHandler returns the default Mux 405 responder whenever
 // a method cannot be resolved for a route.
-func (mx *Mux) MethodNotAllowedHandler() http.HandlerFunc {
+func (mx *Mux) MethodNotAllowedHandler(methodsAllowed ...methodTyp) http.HandlerFunc {
 	if mx.methodNotAllowedHandler != nil {
 		return mx.methodNotAllowedHandler
 	}
-	return methodNotAllowedHandler
+	return methodNotAllowedHandler(methodsAllowed...)
```
EwenQuim, line 385: Unfortunately I cannot pass the Allowed Methods to the `(Mux).methodNotAllowedHandler` given by the user directly in the Mux with the `MethodNotAllowed()` method, as it would break the current API.

It only set the `Allow` header when using default handler. But I think it is enough anyway, at least for this PR.

We might want to give the possibility for users to implement more easily this feature in their own handler by making the `(Context).methodsAllowed` field public. Is it worth it? It's up to you.

VojtechVitek (commented):
This looks so much better compared to the previous middleware solution. Good job!

@pkieltyka This PR sets `Allow` header on HTTP 405 responses (which makes chi compliant with RFC 9110) and doesn't affect the performance of the "happy path". Looks pretty good to me. Mind having a look?

niss36:
Seems really useful as this is required by the RFC 😄

Remy-Luciani:
Intent looks good, code looks good, perf look good. 😁 I hope it will be merged soon!

EwenQuim:
Hey, @VojtechVitek! I was wondering if you had any updates on @pkieltyka?

In case he's tied up or not available, I was thinking we could explore alternative ways to merge this feature. Seems like it could be good for a lot of people!

VojtechVitek (approved):
LGTM from me

thexpand:
Do we have docs on how to use this?

pkieltyka (approved):

pkieltyka:
thanks for the contribution @EwenQuim and to the reviewers (@VojtechVitek + others)

merged + released https://github.com/go-chi/chi/releases/tag/v5.0.10

EwenQuim:
> Do we have docs on how to use this?

@thexpand it works "out of the box", it automatically sets the Allow header in case of a 405, so no docs are required :)

PR #778, opened 2023-01-04
Fix typo

keribend:


VojtechVitek:
It's trie. https://github.com/go-chi/chi/pulls?q=is%3Apr+is%3Aclosed+trie

keribend:
Trie does not look english. Also, referred link just beside link to "Radix tree", which make sense.

Why text should be "radix trie" while link points to "radix_tree" instead?

Also, what does it mean "trie"?

PR #786, opened 2023-01-27
Make CleanPath middleware work with other routers

romain-tadiello:
The rctx is not checked and could be nil if the function isn't used with a chi.Router. 
It ends up in a panic.

Additional point: the cleaning only happens in the rctx.RoutePath is an empty string. If a function like [GetHead](https://github.com/go-chi/chi/blob/v1.5.4/middleware/get_head.go#L10) [¶](https://pkg.go.dev/github.com/go-chi/chi/middleware#GetHead) is called before it won't clean the path.

### DIFF
--- a/middleware/clean_path.go
+++ b/middleware/clean_path.go
@@ -12,17 +12,19 @@ import (
 func CleanPath(next http.Handler) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		rctx := chi.RouteContext(r.Context())
-
-		routePath := rctx.RoutePath
+		
+		var routePath string
+		if rctx != nil {
+			routePath = rctx.RoutePath
+		}
 		if routePath == "" {
 			if r.URL.RawPath != "" {
 				routePath = r.URL.RawPath
 			} else {
 				routePath = r.URL.Path
 			}
-			rctx.RoutePath = path.Clean(routePath)
 		}
-
+		rctx.RoutePath = path.Clean(routePath)
 		next.ServeHTTP(w, r)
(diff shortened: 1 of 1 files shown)


middleware/clean_path.go
```diff
@@ -12,17 +12,19 @@ import (
...
 		if routePath == "" {
 			if r.URL.RawPath != "" {
 				routePath = r.URL.RawPath
 			} else {
 				routePath = r.URL.Path
 			}
-			rctx.RoutePath = path.Clean(routePath)
 		}
-
+		rctx.RoutePath = path.Clean(routePath)
```
pkieltyka, line 27: wont it panic on this line if rctx is nil..? and this PR changes the logic of the method as well, calling path.Clean() more often then previously

pkieltyka (changes requested):

gangelop:
in #787 @romain-tadiello says

> Additional point: the cleaning only happens in the rctx.RoutePath is an empty string. If a function like [GetHead](https://github.com/go-chi/chi/blob/v1.5.4/middleware/get_head.go#L10) [¶](https://pkg.go.dev/github.com/go-chi/chi/middleware#GetHead) is called before it won't clean the path.

I'm not sure but this might be relevant to why he moved the `path.Clean()` out of the conditional.
Apologies if that's not relevant. I literally just started using chi today.

VojtechVitek:
For reference, this is similar fix to https://github.com/go-chi/chi/pull/841 but in a different MW.

@romain-tadiello Closing due to inactivity. Feel free to reopen the PR if you plan to address the above comment.

PR #788, opened 2023-01-29
refactor: using strings.ReplaceAll instead of strings.Replace

robsonpeixoto:


pkieltyka:
what’s the benefit?

robsonpeixoto:
Only legibility, @pkieltyka. `strings.ReplaceAll` call `strings.Replace` directly:

```go
func ReplaceAll(s, old, new string) string {
	return Replace(s, old, new, -1)
}
```

pkieltyka:
ok thanks, I will skip then

PR #790, opened 2023-01-30
Don't strip trailing slash if it's escaped in StripSlashes middlware

nuuls:
This fixes an edge case where a path ending in a trailing encoded slash `%2F` is incorrectly stripped and rctx.RoutePath is set to the unescaped Path, resulting in routing errors.

The specific edge case in my project is:
- rctx is nil
- rctx.RoutePath is empty
- r.URL.RawPath contains multiple escaped slashes in various places
- rctx.RoutePath is set to the unescaped values contained in r.URL.Path
- The Mux no longer matches the route `/link_resolver/{url}` because all `%2F` are now `/`

A similar issue also exists in `RedirectSlashes`, I can open a second PR for that as well if you'd like.

### DIFF
--- a/middleware/strip.go
+++ b/middleware/strip.go
@@ -3,6 +3,7 @@ package middleware
 import (
 	"fmt"
 	"net/http"
+	"strings"
 
 	"github.com/go-chi/chi/v5"
 )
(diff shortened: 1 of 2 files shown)


VojtechVitek (commented):
LGTM

Can you please force-push to the branch to kick off CI?

brasic:
@VojtechVitek it seems like the author may have abandoned this PR, but it's needed: the current behavior inappropriately strips trailing `%2F` strings from the URL inappropriately. I've tested this code in our project and it fixed the issue. If @nuuls isn't around to kick-off CI I'm happy to open a new PR with this content to do so.

Although this might properly be considered a breaking change since in my experience it changes the routing of plenty of real-world URLs that apps might have depended on.

nuuls:
Ah sorry, completely forgot about this one 😄 Pushed it now

PR #793, opened 2023-02-05
Simplify wrap writer

matterpale:


VojtechVitek (commented):
This approach works as well as the existing code. LGTM.

https://go.dev/play/p/4yce_wKJp_g

Can we please run performance tests comparing this PR to the existing code? I'm mostly curious about the returned anonymous structs passed down as a value.

While testing this PR, I think I found a bug (present in the existing chi code too). If the underlying response writer only implements `io.ReaderFrom`, we don't handle that.

I'll write some tests for this.

vtolstov:
I'm using go:generate to create all possible combinations like via https://git.unistack.org/unistack-org/micro-wrapper-sql/src/branch/master/gen.go so code looks like https://git.unistack.org/unistack-org/micro-wrapper-sql/src/branch/master/wrap_gen.go

PR #795, opened 2023-02-15
chore: recoverer middleware skip write header on upgrade connection

### LINKED ISSUE #661: `middlewear.Recover` causes undefined behaviour if a connection has been upgraded.
The [`middlewear.Recover` method ](https://github.com/go-chi/chi/blob/df44563f0692b1e677f18220b9be165e481cf51b/middleware/recoverer.go#L34)attempts to write an HTTP 500 Internal Server Error to a response as part of the error recovery process. This will cause apparently undefined behavior if the connection has been hijacked.

In our use case, we have upgraded our connection to a WebSocket using `golang.org/x/net/websocket/websocket.Handler` to hijack the connection. A bug in our WebSocket handler, specifically a null pointer error, cause Chi to call `Recover`. The attempt to write to the upgraded connection after hijack caused apparently undefined behaviour: Internal variables were written down the WebSocket to the client, and the Go runtime reported a generic `panic`, reducing the diagnosting use of the `Recover` method.

Can we suggest the `Recover` method test if the connection has been upgraded or hijacked, and alter it's behavior appropriate - for example not trying to write to the HTTP connection, and possible providing an error over the WebSocket?

fatelei:
Fix #661 

skip write header when connection has been upgrade

### DIFF
--- a/middleware/recoverer.go
+++ b/middleware/recoverer.go
@@ -36,7 +36,9 @@ func Recoverer(next http.Handler) http.Handler {
 					PrintPrettyStack(rvr)
 				}
 
-				w.WriteHeader(http.StatusInternalServerError)
+				if r.Header.Get("Connection") != "Upgrade" {
+					w.WriteHeader(http.StatusInternalServerError)
+				}
 			}
 		}()
 


VojtechVitek (approved):
LGTM

Issue #808, opened 2023-03-31
Request size middleware


farazfazli:
I notice often times when I'm using Chi I add in a middleware to limit request body size. This is a good practice if we want something smaller or larger than the default request size (1MB).

I propose that we add a middleware which takes in a number of bytes and restricts request size accordingly.

PR #809, opened 2023-03-31
Request size middleware

farazfazli:
Adds the middleware discussed in #808.

### DIFF
--- a/middleware/request_size.go
+++ b/middleware/request_size.go
@@ -1,0 +1,18 @@
+package middleware
+
+import (
+	"net/http"
+)
+
+// RequestSize is a middleware that will limit request sizes to a specified
+// number of bytes. It uses MaxBytesReader to do so.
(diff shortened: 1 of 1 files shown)


pkieltyka:
Nice

middleware/request_size.go
```diff
@@ -0,0 +1,18 @@
...
+import (
+	"net/http"
+)
+
+// RequestSize is a middleware that will limit request sizes to a specified
+// number of bytes. It uses MaxBytesReader to do so.
+func RequestSize(bytes int64) func(http.Handler) http.Handler {
+	f := func(h http.Handler) http.Handler {
+		fn := func(w http.ResponseWriter, r *http.Request) {
+			r.Body = http.MaxBytesReader(w, r.Body, bytes)
```
VojtechVitek, line 12: @pkieltyka @farazfazli Do you think we could also error out (`HTTP 4xx`) early if `r.ContentLength > bytes` ?

PR #811, opened 2023-04-09
Support matching parameters that containing the delimiter

### LINKED ISSUE #781: Matching parameters containing the delimiter
Example:
```go
r.Get("/{name}.json", ...)

// Request: GET /1.0.json
```

This fails because of (how I think) the routing algorithm works. For `{name}.json` the first character after the placeholder is saved as "tail delimiter". The route search now resolves the placeholder by searching for the delimiter in the string. For `/test.json` this searches for `.` and the result is `test`. However if the request is `/1.0.json` this logic fails to resolve the correct value because the delimiter `.` is contained in the value. The result is `1` instead of the expected `1.0`.

A possible fix would be to search for the whole non-placeholder-suffix `.json`. That would still fail for something like `{name}.` or `{name}.{extension}`.
Another fix would be to make the search greedy and read the value until the last occurence of the delimiter. As this works now fine for my case it fails for `{name}.{other}.json` even without the delimiter inside the values. Still I think the greedy matching is more often what you would expect.

Greedy example:
```diff
			// serially loop through each node grouped by the tail delimiter
			for idx := 0; idx < len(nds); idx++ {
				xn = nds[idx]

+ 				segmentlen := strings.IndexByte(xsearch, '/')
+ 				if segmentlen == -1 {
+ 					segmentlen = len(xsearch)
+ 				} else {
+ 					segmentlen += 1 // just keep the old code working
+ 				}

				// label for param nodes is the delimiter byte
- 				p := strings.IndexByte(xsearch, xn.tail)
+ 				p := strings.LastIndexByte(xsearch[:segmentlen], xn.tail)

				if p < 0 {
					if xn.tail == '/' {
```

6543:
close  #781

### DIFF
--- a/tree_test.go
+++ b/tree_test.go
@@ -232,8 +232,8 @@ func TestTreeMoar(t *testing.T) {
 		{m: mGET, r: "/articles/456/data.json", h: hStub11, k: []string{"id"}, v: []string{"456"}},
 
 		{m: mGET, r: "/articles/files/file.zip", h: hStub12, k: []string{"file", "ext"}, v: []string{"file", "zip"}},
-		{m: mGET, r: "/articles/files/photos.tar.gz", h: hStub12, k: []string{"file", "ext"}, v: []string{"photos", "tar.gz"}},
-		{m: mGET, r: "/articles/files/photos.tar.gz", h: hStub12, k: []string{"file", "ext"}, v: []string{"photos", "tar.gz"}},
+		{m: mGET, r: "/articles/files/photos.tar.gz", h: hStub12, k: []string{"file", "ext"}, v: []string{"photos.tar", "gz"}},
+		{m: mGET, r: "/articles/files/photos.tar.gz", h: hStub12, k: []string{"file", "ext"}, v: []string{"photos.tar", "gz"}},
 
 		{m: mPUT, r: "/articles/me", h: hStub13, k: []string{}, v: []string{}},
 		{m: mGET, r: "/articles/me", h: hStub, k: []string{"id"}, v: []string{"me"}},
--- a/tree_test.go
+++ b/tree_test.go
@@ -31,6 +31,7 @@ func TestTree(t *testing.T) {
 	hHubView3 := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {})
 	hUserInfo0 := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {})
 	hUserInfo1 := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {})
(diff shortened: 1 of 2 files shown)


tree_test.go
```diff
@@ -221,8 +232,8 @@ func TestTreeMoar(t *testing.T) {
 		{m: mGET, r: "/articles/456/data.json", h: hStub11, k: []string{"id"}, v: []string{"456"}},
 
 		{m: mGET, r: "/articles/files/file.zip", h: hStub12, k: []string{"file", "ext"}, v: []string{"file", "zip"}},
-		{m: mGET, r: "/articles/files/photos.tar.gz", h: hStub12, k: []string{"file", "ext"}, v: []string{"photos", "tar.gz"}},
-		{m: mGET, r: "/articles/files/photos.tar.gz", h: hStub12, k: []string{"file", "ext"}, v: []string{"photos", "tar.gz"}},
+		{m: mGET, r: "/articles/files/photos.tar.gz", h: hStub12, k: []string{"file", "ext"}, v: []string{"photos.tar", "gz"}},
```
6543, line 235: I would call it breaking :/

6543:
@pkieltyka @VojtechVitek please have a look!

PS: this might require a major version bump but I think it's worth it

tree_test.go
silverwind: Maybe add a test case that parses this correctly? It should be called out in the changelog how to fix such cases.

VojtechVitek (commented):
Thanks @6543. I think this looks great!

---

It'd be nice to provide a high-level description of the changes. Something like:

Given this route
```go
r.Get("/files/{file}.{ext}", handleFileExt)
```
and a request to `/files/photos.tar.gz` path

chi will match the longest leftmost parameter first (same as Regex or `filepath.Ext()`):

```diff
fmt.Println(chi.URLParam(req, "file"))
fmt.Println(chi.URLParam(req, "ext"))

- photos
- tar.gz
+ photos.tar
+ gz
```

---

Yes, this can be considered a breaking change. But I'm not sure if it's worth the major version bump.

Anyway, let's see what @pkieltyka has to say

tree_test.go
6543: Sorry, I dont understand what you mean / what I have to do

silverwind: May have been a brain fart. I do see you added positive test cases above.

Not sure how to proceed here. These cases do look bad. I think a option whether parameters catch a dot may be needed, ideally on route level.

VojtechVitek:
This is a breaking change, we can consider it for chi v6.

VojtechVitek:
Closing in favor of https://github.com/go-chi/chi/pull/813

Issue #816, opened 2023-04-26
Collateral damage when upgrading to version 5.0.8 with Datadog


luca147:
Hello, we have our healtcheck over `/` for example: `http://xx.xx.xxx.xx:8080/` and also we use Datadog with the go-tracer implementation.

But after upgrade go-chi from 5.0.7 to 5.0.8, our resource dashboard found a unknown new entry.
This is cause because [this](https://github.com/DataDog/dd-trace-go/blob/main/contrib/go-chi/chi.v5/chi.go#L79-L80)

https://github.com/go-chi/chi/blob/77d709fcde559ca625d4e126478835a0e85d79a9/context.go#L126

I think one solution to keep find the `/` route valid is only trim the dash if have a route like this `/example/`

What you think?

PR #817, opened 2023-04-26
Keep "/" route in RoutePattern method

luca147:
Issue related: https://github.com/go-chi/chi/issues/816

### DIFF
--- a/context.go
+++ b/context.go
@@ -112,15 +112,18 @@ func (x *Context) URLParam(key string) string {
 //
 // For example,
 //
-//   func Instrument(next http.Handler) http.Handler {
-//     return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-//       next.ServeHTTP(w, r)
-//       routePattern := chi.RouteContext(r.Context()).RoutePattern()
-//       measure(w, r, routePattern)
-//   	 })
-//   }
+//	func Instrument(next http.Handler) http.Handler {
+//	  return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+//	    next.ServeHTTP(w, r)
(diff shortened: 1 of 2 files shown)


VojtechVitek:
Sorry I don't follow the issue. #816 looks to be in DataDog/dd-trace-go?

luca147:
Hi @VojtechVitek ! Yes the issue take place into dd-trace-go, but after upgrade go-chi to 5.0.8

```go
	routePattern := cfg.modifyResourceName(chi.RouteContext(r.Context()).RoutePattern())
	span.SetTag(ext.HTTPRoute, routePattern)
	var resourceName string
	if cfg.resourceNamer != nil {
		resourceName = cfg.resourceNamer(r)
	} else {
		resourceName = routePattern
		if resourceName == "" {
			resourceName = "unknown"
		}
		resourceName = r.Method + " " + resourceName
	}
	span.SetTag(ext.ResourceName, resourceName)
```
https://github.com/DataDog/dd-trace-go/blob/main/contrib/go-chi/chi.v5/chi.go#L77-L89

As you can see, the tracer for chi, get the resourceName from `RoutePattern`, but with the change introduced in 5.0.8 this route is removed by trim call, and we started fall into`resourceName = "unknown"`

I think the root route is valid and must be returned by `RoutePattern`

luca147:
Hello @VojtechVitek if you think this case must be contemplated in dd-trace-go I can close this PR and Issue, but to make it work is a bit hacky due I need to check if the path is '/' to prevent the tracer from not seeing it as an unresolved path.

pkieltyka:
solved in https://github.com/go-chi/chi/commit/f4ab9b1f050bdcf14bd096399e8bd57363406514

Issue #821, opened 2023-05-03
Heartbeat middleware is not invoked when added to a subrouter


mo-rieger:
Hi there,

The Heartbeat middleware is not invoked when attached to a sub router, whereas other middlewares are.

The expected behaviour would be that `/connectivity/ping` would return `200`. Instead, a `404` is returned.

```go
package main

import (
	"net/http"

	"github.com/go-chi/chi/v5"
	"github.com/go-chi/chi/v5/middleware"
)

func main() {
	r := chi.NewRouter()

	r.Use(middleware.RequestID)
	r.Use(middleware.RealIP)
	r.Use(middleware.Recoverer)
	
	r.Mount("/connectivity", NewSubrouter())
	
	http.ListenAndServe(":3333", r)
}

func NewSubrouter() chi.Router {
	r := chi.NewRouter()
	r.Use(middleware.Logger)
	r.Use(middleware.Heartbeat("/ping"))
	
	r.Get("/", func(w http.ResponseWriter, r *http.Request) {
		w.Write([]byte("."))
	})
	
	return r
}
```
Best,
Moritz

kubosuke:
I created PR for this issue, please check

VojtechVitek:
Please see https://github.com/go-chi/chi/pull/822#pullrequestreview-2313031493

PR #822, opened 2023-05-07
Enable Heartbeat Middleware for subrouter too

kubosuke:
- make enable `Heartbeat` middleware to use in subrouter
- add test for this middleware

PR for https://github.com/go-chi/chi/issues/821

### DIFF
--- a/middleware/heartbeat.go
+++ b/middleware/heartbeat.go
@@ -3,6 +3,8 @@ package middleware
 import (
 	"net/http"
 	"strings"
+
+	"github.com/go-chi/chi/v5"
 )
 
 // Heartbeat endpoint middleware useful to setting up a path like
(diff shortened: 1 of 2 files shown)


VojtechVitek (changes requested):
Hi @kubosuke,

would something like this work for you, since we just merged https://github.com/go-chi/chi/pull/875?

```go
	r := chi.NewRouter()
	middleware.Heartbeat("/ping")
	
	r.Route("/_api", func(r chi.Router) {
		r.Use(middleware.StripPrefix("/_api"))
		r.Use(middleware.Heartbeat("/ping"))

		r.Handle("/rpc/*", rpcServerHandler)
	})
```

Heartbeat mw is usually invoked on all server requests, so I'd like to avoid slowing it down with the extra `strings.EqualFold()`.

PR #831, opened 2023-07-10
Added Any route

lesichkovm:


chi.go
```diff
@@ -95,6 +95,7 @@ type Router interface {
 	MethodFunc(method, pattern string, h http.HandlerFunc)
 
 	// HTTP-method routing along `pattern`
+	Any(pattern string, h http.HandlerFunc)
```
VojtechVitek, line 98: Have you tried
```
Handle(pattern string, h http.Handler)
HandleFunc(pattern string, h http.HandlerFunc)
```
?

lesichkovm: These two do work as intended, but are not very intuitive.

The Any method is more recognizable for folks coming from other routers, like Laravel, Fiber, etc

mstgnz: Laravel:
Sometimes you may need to register a route that responds to multiple HTTP verbs. You may do so using the match method. Or, you may even register a route that responds to all HTTP verbs using the any method:
`Route::match(['get', 'post'], '/', function () {});`
`Route::any('/', function () {});`

Fiber:
All will register the route on all HTTP methods
Almost the same as app.Use but not bound to prefixes
`func (app *App) All(path string, handlers ...Handler) Router`

Issue #839, opened 2023-08-09
Panic if rctx is nil


yurineydachin:
https://github.com/go-chi/chi/blame/99bc99e0bf0edc147e82765f2848167922a5c644/middleware/url_format.go#L67

PR #841, opened 2023-08-10
Add error handling for nil RouteContext in URLFormat

Taaipi:
Fix https://github.com/go-chi/chi/issues/839
Ref https://github.com/go-chi/chi/pull/718

In the current URLFormat middleware implementation, the absence of rctx (Route Context) could lead to unexpected behavior or panic errors in subsequent processing.

To avoid this, I have prioritized not allowing unexpected behavior by implementing error handling for the cases where rctx is missing or nil. Now, instead of failing silently or causing a panic error, the system will return a proper HTTP error response.

This change ensures that the system behaves consistently and avoids potential pitfalls that might arise from missing or improperly initialized rctx.

### DIFF
--- a/middleware/url_format_test.go
+++ b/middleware/url_format_test.go
@@ -1,8 +1,10 @@
 package middleware
 
 import (
+	"context"
 	"net/http"
 	"net/http/httptest"
+	"strings"
 	"testing"
 
 	"github.com/go-chi/chi/v5"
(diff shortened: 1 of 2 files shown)


middleware/url_format_test.go
```diff
@@ -67,3 +69,30 @@ func TestURLFormatInSubRouter(t *testing.T) {
 		t.Fatalf(resp)
 	}
 }
+
+func TestURLFormatWithoutChiRouteContext(t *testing.T) {
+	r := chi.NewRouter()
+
+	r.Use(func(next http.Handler) http.Handler {
+		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			newCtx := context.WithValue(r.Context(), chi.RouteCtxKey, nil)
```
VojtechVitek, line 78: Hi @Taaipi,

Thanks for the PR!

It looks like this test cases injects a `nil` RouteContext down the chain intentionally.

Do you have a scenario how could this happen in real-world use case? I wonder how is the `middleware.URLFormat` ever called without chi route context unless we set it to `nil` explicitly.

Thanks!

VojtechVitek: I guess based on https://github.com/go-chi/chi/pull/718, it'd be better to create a test case with some sub-routers (instead of setting `nil`). Right?

VojtechVitek: Are you using middleware.URLFormat outside of chi router?

VojtechVitek (changes requested):
Hi,

Thank you for your contribution.

However, I think I prefer https://github.com/go-chi/chi/pull/878 approach. I don't think returning `http.StatusInternalServerError` is the right thing to do, if this middleware is used outside of chi.

PR #854, opened 2023-10-02
feat: make fileserver example simpler

zHenriqueGN:
Remove a lot of code to make the file server example simpler


zHenriqueGN:
Remove a lot of code to make the file server example simpler

PR #856, opened 2023-10-07
Fix condition in TestRedirectSlashes

tchssk:
I changed the conditions to make sure that both the response body and status are correct.


VojtechVitek (approved):
LGTM.

However, I think we should change `t.Fatalf(body)` too
1. body acts as "format"
2. responseCode is missing

PR #862, opened 2023-10-17
Removes ambiguity around "fresh middleware stack" in GoDoc

TheRandomCharacter:
#### For `Group()`:
`fresh` implies that a new empty middleware stack is created, while in fact the parent's stack is copied.

#### For `Route()`:
Looks like `with a fresh middleware stack` was added to emphasize that while the parent middleware stack is frozen, one can still add middlewares to new router's stack before starting to define routes on it.

In my opinion this is clear from `creates a new Mux`. Adding `with a fresh middleware stack` makes it more obscure that the parent middleware stack will be executed wrapping the new one.

### DIFF
--- a/mux.go
+++ b/mux.go
@@ -250,7 +250,7 @@ func (mx *Mux) With(middlewares ...func(http.Handler) http.Handler) Router {
 	return im
 }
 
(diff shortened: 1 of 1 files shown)


VojtechVitek (approved):
LGTM

PR #876, opened 2023-11-22
Create a global pool to share contexts between all routes

rouzier:


VojtechVitek (commented):
Hi @rouzier,

This looks like a great contribution. Sorry for the late review!

Will this make any performance impact to chi? I suppose it could, if there were many sub-routers?

Can you please share any benchmark numbers with us?

(this PR needs a rebase now, which is our bad - sorry!)

PR #878, opened 2023-11-28
set a new route ctx if rctx is nil

shedyfreak:
PR to solve https://github.com/go-chi/chi/issues/839

### DIFF
--- a/middleware/url_format.go
+++ b/middleware/url_format.go
@@ -55,6 +55,10 @@ func URLFormat(next http.Handler) http.Handler {
 			path = rctx.RoutePath
 		}
 
+		if rctx == nil {
+			rctx = chi.NewRouteContext()
+		}
+
 		if strings.Index(path, ".") > 0 {
 			base := strings.LastIndex(path, "/")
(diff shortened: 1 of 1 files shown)


yurineydachin (approved):

VojtechVitek:
Hey @shedyfreak, @yurineydachin, @AndreiOslin,

I'm just checking back on this PR and I wonder -- how do you trigger the #839 panic? Under what circumstance is the route context nil?

Might be helpful to provide a test case for this too.

@shedyfreak what's your thoughts on https://github.com/go-chi/chi/pull/878#discussion_r1418718973 ?

VojtechVitek:
It looks like https://github.com/go-chi/chi/pull/841 was submitted first and has a test case too. I'll follow up there.

Thanks for the PR!

VojtechVitek (commented):
Hi, @shedyfreak

After all, I prefer this fix over https://github.com/go-chi/chi/pull/841.

Can we move `if rctx == nil {` check to the below `if` branch to avoid the panic per https://github.com/go-chi/chi/pull/878/files#r1418718973?

Thank you

PR #880, opened 2023-12-05
Support Content-Type headers with charset/boundary parameters

GocaMaric:
Handler Function Parameters: Changed the function parameters of the returned handler functions to use the http.Handler interface directly. This makes the code more idiomatic and allows for better compatibility with other middleware.

Inline Function Definitions: Eliminated the need for the intermediate fn variable in both SetHeader and AllowContentType functions. You can directly return the http.HandlerFunc in a more concise manner.

Code Comments: Added a comment to explain the splitting of the content type string to extract the MIME type. This improves code readability.

Code Formatting: Applied consistent code formatting to improve readability.

### DIFF
--- a/middleware/content_type.go
+++ b/middleware/content_type.go
@@ -6,36 +6,32 @@ import (
 )
 
 // SetHeader is a convenience handler to set a response header key/value
-func SetHeader(key, value string) func(next http.Handler) http.Handler {
+func SetHeader(key, value string) func(http.Handler) http.Handler {
 	return func(next http.Handler) http.Handler {
-		fn := func(w http.ResponseWriter, r *http.Request) {
+		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 			w.Header().Set(key, value)
 			next.ServeHTTP(w, r)
-		}
-		return http.HandlerFunc(fn)
+		})
 	}
 }
 
 // AllowContentType enforces a whitelist of request Content-Types otherwise responds
 // with a 415 Unsupported Media Type status.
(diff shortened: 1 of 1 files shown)


middleware/content_type.go
```diff
@@ -6,44 +6,40 @@ import (
...
+				// Skip check for empty content body
 				next.ServeHTTP(w, r)
 				return
 			}
 
-			s := strings.ToLower(strings.TrimSpace(r.Header.Get("Content-Type")))
-			if i := strings.Index(s, ";"); i > -1 {
-				s = s[0:i]
-			}
+			s := strings.ToLower(strings.TrimSpace(strings.Split(r.Header.Get("Content-Type"), ";")[0]))
```
VojtechVitek, line 34: This can panic.

VojtechVitek (changes requested):

middleware/content_type.go
VojtechVitek: https://go.dev/play/p/ISGDkoiv4Us

VojtechVitek (approved):
LGTM. Thanks for your contribution :)

middleware/content_type.go
VojtechVitek, line 34: This doesn't change the behavior (much).

The code is slightly cleaner and can trim whitespaces within first part of the header value.

I'll merge.

GocaMaric: Thanks for the feedback! I agree that trimming whitespaces and cleaning up the code makes it more efficient and readable. Appreciate you merging the changes!

PR #881, opened 2023-12-05
Update content_type.go

GocaMaric:


GocaMaric:
My last update could trigger PANIC if contentType is empty, but now should work well.

VojtechVitek:
Ref. https://github.com/go-chi/chi/pull/880

VojtechVitek:
Sorry I was wrong. I'll reopen #880.

PR #884, opened 2023-12-14
Add function to expose allowed methods for use in custom 405-handlers

### LINKED ISSUE #870: Feature: Expose Context.methodsAllowed for use by MethodNotAllowedHandler
The default MethodNotAllowed handler uses `Context.methodsAllowed` to set the "Allow" header in the response.

It would be nice to expose that list of allowed methods to a custom handler, as well.

Options I've considered:

1. Simply export that field of the `Context` struct. This would allow access when the `Context` value is retrieved from the `r.Context()`
2. Add a function that extracts the unexported field from a context, and returns it. Something like:

    ```golang
    func AllowedMethods(ctx context.Context) []string {
        rctx, ok := ctx.Value(RouteCtxKey).(*Context)
        if ok {
            return rctx.methodsAllowed
        }
        return nil
    }
    ```
3. Add a methodn on `Router` that takes a function that receives the list, and returns a handler. This seems really intrusive, though, and largely duplicates existing functionality:

    ```golang
    func (mx *Mux) CustomMethodNotAllowedHandler(func(allowedMethods []string) http.HandlerFunc
    ```

flimzy:
Fixes #870


context.go
```diff
@@ -34,6 +34,21 @@ func NewRouteContext() *Context {
...
+// WithRouteContext returns the list of methods allowed for the current
+// request, based on the current routing context.
+func AllowedMethods(ctx context.Context) []string {
+	if rctx := RouteContext(ctx); rctx != nil {
+		result := make([]string, 0, len(rctx.methodsAllowed))
+		for _, method := range rctx.methodsAllowed {
+			if method := methodTypString(method); method != "" {
+				result = append(result, method)
+			}
+		}
```
VojtechVitek, line 46: Hi, do you think it'd make sense to expose a method on `rctx` instead?

```go
rctx := chi.RouteContext(ctx)

var methods []string = rctx.AllowedMethods()
```

flimzy: That would serve my purpose just as well. Would you like me to update the PR?

VojtechVitek: sounds good :) thank you

flimzy: I've made the requested change, and rebased on a current version of the `master` branch.

PR #885, opened 2023-12-16
bump setup-go and checkout action versions

purificant:
This PR updates CI to use the latest versions of [actions/setup-go](https://github.com/actions/setup-go) and [actions/checkout](https://github.com/actions/checkout).

`check-latest: true` is to ensure the latest patch version is used.
`cache: false` is to keep the same default behaviour as in v3 and to avoid warning when looking for `go.sum` in the repository.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,11 +28,13 @@ jobs:
 
     steps:
     - name: Install Go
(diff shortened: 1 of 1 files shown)


purificant:
image
Example behavior: https://github.com/go-chi/chi/actions/runs/7232964360/job/19707719923

PR #887, opened 2023-12-20
Fixed incorrect comment about routing

### LINKED ISSUE #879: Incorrect Route Matching Comment in chi.go
The comment on line `54` of the `chi.go` file contains inaccurate information regarding route matching. It states:
```
// "/page/{other}/index" also matches "/page/intro/latest"
```
However, this is not accurate. The route `/page/{other}/index` expects a path component after `/page/` followed by `/index.` It does not match paths like `/page/intro/latest`.

#### Proposed Correction:
Update the comment on line `54` to accurately reflect the behavior of the `/page/{other}/index` route. For example:
```
// "/page/{other}/index" matches "/page/some_value/index" but not "/page/intro/latest"
```
This correction will provide users with accurate information about the route's matching behavior.

jtams:
In the chi.go file there are examples of routing behaviors. One comment was meant to demonstrate how other routes can match wildcard routes, but there was a typo that made it incorrect:

`// "/page/{other}/index" also matches "/page/intro/latest"`

Resolves #879

### DIFF
--- a/chi.go
+++ b/chi.go
@@ -51,7 +51,7 @@
 //	"/user/{name}" matches "/user/jsmith" but not "/user/jsmith/info" or "/user/jsmith/"
 //	"/user/{name}/info" matches "/user/jsmith/info"
 //	"/page/*" matches "/page/intro/latest"
-//	"/page/{other}/index" also matches "/page/intro/latest"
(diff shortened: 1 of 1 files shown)


VojtechVitek (approved):
LGTM

PR #890, opened 2023-12-27
middleware: add method to WrapResponseWriter for getting response writing time

vasayxtx:
This PR introduces a new method, `ElapsedWriteTime()`, to the `middleware.WrapResponseWriter`. This addition is particularly useful in scenarios where clients are slow or responses are large, necessitating a measure of the time taken to write the response.

Please note, that this change may break backward compatibility. An alternative could be to introduce a separate interface containing only this new method. However, this would require developers to perform type assertions in the client code, which may reduce clarity. So, your feedback on this approach is appreciated.


VojtechVitek (changes requested):
Thanks for the contribution! I think it's a cool feature and I see nice use-cases for telemetry.

However, I think it should be an opt-in feature, as I'm seeing some performance issues with the current approach.

VojtechVitek:
Perhaps a better approach would be to introduce a hook on first written byte, e.g.

```go
        start := time.Now()
        var firstByteWritten time.Time
        
	ww := middleware.NewWrapResponseWriter(w, r.ProtoMajor)
	ww.OnWriteHeader(func(statusCode int) {
		firstByteWritten = time.Now()
	}) 
	
	next.ServeHTTP(ww, r.WithContext(ctx))
	
	elapsed := time.Since(start)
	writeElapsed := time.Since(firstByteWritten)
```

And maybe we could generalize this even further. Thoughts?

PR #893, opened 2024-01-10
doc: add Sunset middleware in readme

thedevsaddam:
Include `Sunset` middleware in readme documentation so that user can get to know about the middleware


VojtechVitek (approved):
LGTM. Thank you!

PR #897, opened 2024-01-18
Extend Handle method to parse HTTP method in pattern

Spartan09:
Updated the `Handle` method in the `chi` router to support patterns that include an HTTP method. This enhancement allows users to specify the HTTP method directly in the route pattern, such as "GET /path". The change is backward compatible, allowing patterns without an HTTP method to be handled as before.

- Added logic to split the pattern into HTTP method and path.
- Validated the extracted HTTP method against known methods.
- Used the existing handle method for routing with the specified method.
- Ensured backward compatibility for patterns without an HTTP method.
- Included error handling for unrecognized HTTP methods.
- Added comprehensive tests to cover new functionality and edge cases.

### DIFF
--- a/mux.go
+++ b/mux.go
@@ -107,6 +107,20 @@ func (mx *Mux) Use(middlewares ...func(http.Handler) http.Handler) {
 // Handle adds the route `pattern` that matches any http method to
 // execute the `handler` http.Handler.
 func (mx *Mux) Handle(pattern string, handler http.Handler) {
+	parts := strings.SplitN(pattern, " ", 2)
+	if len(parts) == 2 {
+		methodStr := strings.ToUpper(parts[0])
+		path := parts[1]
+
+		method, ok := methodMap[methodStr]
+		if !ok {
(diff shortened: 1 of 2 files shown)


pkieltyka:
thanks for the PR :) will review

pkieltyka:
nice work :D LGTM

mux.go
```diff
@@ -107,6 +107,20 @@ func (mx *Mux) Use(middlewares ...func(http.Handler) http.Handler) {
...
 // execute the `handler` http.Handler.
 func (mx *Mux) Handle(pattern string, handler http.Handler) {
+	parts := strings.SplitN(pattern, " ", 2)
+	if len(parts) == 2 {
+		methodStr := strings.ToUpper(parts[0])
+		path := parts[1]
+
+		method, ok := methodMap[methodStr]
+		if !ok {
+			panic("chi: invalid HTTP method specified in pattern: " + methodStr)
```
robsonpeixoto, line 117: `methodStr` or `pattern`?

PR #901, opened 2024-02-12
feat(mux): support http.Request.PathValue in Go 1.22

### LINKED ISSUE #873: Feature: on Go 1.22, fill `(*http.Request).PathValue` with URLParam data
Hey y'all!
For Go 1.22, [this enhanced `ServeMux` routing proposal](https://github.com/golang/go/issues/61410) was accepted and pushed into the main tree, and `*http.Request` has some new methods, `SetPathValue` and `PathValue`. I think `r.PathValue(...)` could act as an alias for `chi.URLParam(r, ...)` when routing with chi. 

There would probably need to be some logic with `//go:build go1.22` that runs `SetPathValue` when inserting keys/values into chi.Context.URLParams for the corresponding http.Request. I have not dug deeper into chi's source code, but I think this would be a welcome addition for when Go 1.22 is released.

angelofallars:
Resolves #873

Adds support for getting path values with `http.Request.PathValue`.

In Go versions earlier than 1.22 that don't have that method this shouldn't break them due to conditional compilation.

`http.Request.PathValue` will act as an alias for `chi.URLParam`.

The only syntax in common with 1.22 pattern matching seems to be single-segment matches like `/b/{bucket}`. Other 1.22 features like remaining segment matches (`/files/{path...}`) and matching trailing slashes (`/exact/match/{$}`) are not yet supported. Since they are more complicated and overlap with already-existing Chi functionality, they should be addressed in a separate discussion/PR.

### DIFF
--- a/mux.go
+++ b/mux.go
@@ -454,6 +454,10 @@ func (mx *Mux) routeHTTP(w http.ResponseWriter, r *http.Request) {
 
 	// Find the route
 	if _, _, h := mx.tree.FindRoute(rctx, method, routePath); h != nil {
+		if supportsPathValue {
+			setPathValue(rctx, r)
+		}
+
 		h.ServeHTTP(w, r)
 		return
 	}
--- a/path_value.go
+++ b/path_value.go
@@ -1,0 +1,20 @@
+//go:build go1.22
+// +build go1.22
+
+package chi
+
+import "net/http"
+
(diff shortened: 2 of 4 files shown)


pkieltyka:
thanks again @angelofallars :)

angelofallars:
@pkieltyka You're welcome! :)

PR #902, opened 2024-03-08
feat: move param logic onto RouteParams type

willnorris:
This moves the logic for fetching URL parameter values from Context onto the RouteParams type itself. This is helpful if you are working solely with the params themselves and don't need the full context.  This also adds a very simple test for the fetching logic (later values overwrite previous values).

As a concrete use case, I'm working on a request decoder (to plug into go-chi/render) that populates struct fields based on route parameter values.  It seems more appropriate to have an API of:

``` go
func (d *PathDecoder) Decode(dst any, params *chi.RouteParams) error
```

rather than:
``` go
func (d *PathDecoder) Decode(dst any, ctx *chi.Context) error
```

### DIFF
--- a/context.go
+++ b/context.go
@@ -99,12 +99,7 @@ func (x *Context) Reset() {
 // URLParam returns the corresponding URL parameter value from the request
 // routing context.
 func (x *Context) URLParam(key string) string {
-	for k := len(x.URLParams.Keys) - 1; k >= 0; k-- {
-		if x.URLParams.Keys[k] == key {
-			return x.URLParams.Values[k]
-		}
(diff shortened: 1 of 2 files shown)


willnorris:
actually, I just noticed that chi calls `request.SetPathValue` when present, so I can have my decoder take a standard `http.Request` and have it support both chi and http.ServeMux.  I still think putting this logic onto RouteParams makes sense, but I don't actually need the change for my use case anymore.  Feel free to close this PR if you prefer.

pkieltyka:
Thanks for the idea

PR #903, opened 2024-03-09
middleware: fix typo in RealIP doc

l2dy:
`True-Client-IP` was introduced later on, so there are 3 headers that need to be trusted.

### DIFF
--- a/middleware/realip.go
+++ b/middleware/realip.go
@@ -22,7 +22,7 @@ var xRealIP = http.CanonicalHeaderKey("X-Real-IP")
 // RemoteAddr will see the intended value.
 //
 // You should only use this middleware if you can trust the headers passed to
-// you (in particular, the two headers this middleware uses), for example
+// you (in particular, the three headers this middleware uses), for example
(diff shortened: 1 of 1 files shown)


l2dy:
@pkieltyka Could you help review this?

middleware/realip.go
```diff
@@ -22,7 +22,7 @@ var xRealIP = http.CanonicalHeaderKey("X-Real-IP")
 // RemoteAddr will see the intended value.
 //
 // You should only use this middleware if you can trust the headers passed to
-// you (in particular, the two headers this middleware uses), for example
+// you (in particular, the three headers this middleware uses), for example
```
VojtechVitek, line 25: LGTM.

Can we list the header names in this comment explicitly?

VojtechVitek (approved):
LGTM

middleware/realip.go
l2dy: They are listed on line 17, so there is no need to duplicate here.

PR #908, opened 2024-04-09
feat(): add CF-Connecting-IP

n33pm:


middleware/realip.go
```diff
@@ -9,12 +9,13 @@ import (
 	"strings"
 )
 
+var cfConnectionIP = http.CanonicalHeaderKey("CF-Connecting-IP")
```
VojtechVitek, line 12: I found myself reimplementing realip middleware myself in the past just to support `CF-Connecting-IP`, which is available on Cloudflare free plan (note: `True-Client-IP` is Enterprise only).

Screen Shot 2024-04-09 at 12 48 46 PM

So I think this makes sense.

VojtechVitek: However, I'm thinking that perhaps we should make this configurable rather than hardcode more provider-specific headers into chi.

I think exposing a list of headers and iterating over them in a loop would probably make sense.

```go
var RealIPHeaders = []string{"True-Client-IP", "X-Forwarded-For", "X-Real-IP"}

func realIP(r *http.Request) string {
	for _, header := RealIPHeaders {
		// check if not empty
		// remove trailing `; *` string
		// check if valid IP
		return ip
	}
	return ""
}
```

It feels like easy change to me.

@pkieltyka thoughts?

middleware/realip.go
```diff
@@ -9,12 +9,15 @@ import (
 	"strings"
 )
 
-var trueClientIP = http.CanonicalHeaderKey("True-Client-IP")
-var xForwardedFor = http.CanonicalHeaderKey("X-Forwarded-For")
-var xRealIP = http.CanonicalHeaderKey("X-Real-IP")
+var DefaultRealIPHeaders = []string{
+	"CF-Connecting-IP", // Cloudflare free plan
```
VojtechVitek, line 13: I know this is useful header, but it feels very provider specific. Should we remove it and let the users define it?

middleware/realip.go
```diff
@@ -39,22 +42,34 @@ func RealIP(h http.Handler) http.Handler {
...
-	} else if xff := r.Header.Get(xForwardedFor); xff != "" {
-		i := strings.Index(xff, ",")
-		if i == -1 {
-			i = len(xff)
+// RealIPCustomHeader is a middleware that sets a http.Request's RemoteAddr to the results
+// of parsing the custom headers.
+//
+// usage:
+// r.Use(RealIPCustomHeader([]string{"X-CUSTOM-IP"}))
+// r.Use(RealIPCustomHeader(append(DefaultRealIPHeaders, "X-CUSTOM-IP")))
```
VojtechVitek, line 50: Feel free to use `CF-Connecting-IP` header in this example comment.

middleware/realip.go
```diff
@@ -39,22 +42,34 @@ func RealIP(h http.Handler) http.Handler {
...
+func RealIPCustomHeader(realIPHeaders []string) func(http.Handler) http.Handler {
```
VojtechVitek, line 51:
```suggestion
func RealIPFromHeaders(headers []string) func(http.Handler) http.Handler {
```

VojtechVitek (commented):
I think this is looking good. Few more suggestions.

middleware/realip.go
n33pm: what about `True-Client-IP`? it's provider specific too.

but if u want i can remove it.

VojtechVitek: let's leave it.. we need to keep `chi` backward compatible :)

VojtechVitek: (`True-Client-IP` is used by Akamai / CF / AWS and many others .. you're right - but I think it's good idea to keep it anyway)

middleware/realip.go
```diff
@@ -39,22 +41,34 @@ func RealIP(h http.Handler) http.Handler {
...
-	} else if xff := r.Header.Get(xForwardedFor); xff != "" {
-		i := strings.Index(xff, ",")
-		if i == -1 {
-			i = len(xff)
+// RealIPFromHeaders is a middleware that sets a http.Request's RemoteAddr to the results
+// of parsing the custom headers.
+//
+// usage:
+// r.Use(RealIPFromHeaders([]string{"CF-Connecting-IP"}))
+// r.Use(RealIPFromHeaders(append(DefaultRealIPHeaders, "CF-Connecting-IP")))
```
VojtechVitek, line 49: nit: `CF-Connecting-IP` needs to be prepended to the beginning of the slice

VojtechVitek (approved):
LGTM

I can test this if @pkieltyka gives it a go.

middleware/realip.go
```diff
@@ -9,9 +9,11 @@ import (
 	"strings"
 )
 
-var trueClientIP = http.CanonicalHeaderKey("True-Client-IP")
-var xForwardedFor = http.CanonicalHeaderKey("X-Forwarded-For")
-var xRealIP = http.CanonicalHeaderKey("X-Real-IP")
+var DefaultRealIPHeaders = []string{
+	"True-Client-IP",   // Cloudflare Enterprise plan
```
VojtechVitek, line 13: gofmt?

middleware/realip.go
```diff
@@ -9,9 +9,11 @@ import (
 	"strings"
 )
 
-var trueClientIP = http.CanonicalHeaderKey("True-Client-IP")
-var xForwardedFor = http.CanonicalHeaderKey("X-Forwarded-For")
-var xRealIP = http.CanonicalHeaderKey("X-Real-IP")
+var DefaultRealIPHeaders = []string{
```
pkieltyka, line 12: I don't love making package-level variables as any dependency could change the values and create side effects. I'd rather make it a const, or unexported and folks can pass their own with the new method

n33pm: you cannot declare an array/slice as const, so this would be close as possible.

```golang
func DefaultRealIPHeaders() []string {
	return []string{
		"True-Client-IP", // Cloudflare Enterprise plan
		"X-Real-IP",
		"X-Forwarded-For",
	}
}
```

or

```golang
func RealIPFromHeaders(realIPHeaders []string, addDefaultHeaders bool) func(http.Handler) http.Handler {
...
```

but is it before or after the custom headers. i would always export default header.

middleware/realip.go
```diff
@@ -39,22 +41,34 @@ func RealIP(h http.Handler) http.Handler {
...
-		i := strings.Index(xff, ",")
-		if i == -1 {
-			i = len(xff)
+// RealIPFromHeaders is a middleware that sets a http.Request's RemoteAddr to the results
+// of parsing the custom headers.
+//
+// usage:
+// r.Use(RealIPFromHeaders([]string{"CF-Connecting-IP"}))
+// r.Use(RealIPFromHeaders(append(DefaultRealIPHeaders, "CF-Connecting-IP")))
+func RealIPFromHeaders(realIPHeaders []string) func(http.Handler) http.Handler {
```
VojtechVitek, line 50: can we do variadic argument `headers ...string` for simple use?

middleware/realip.go
VojtechVitek: If I choose to have a full control of the headers, I wouldn't mind passing them all explicitly.

This is what I'd expect to do:

```diff
- r.Use(middleware.RealIP)
+ r.Use(middleware.RealIPFromHeaders("CF-Connecting-IP", "X-Real-IP", "X-Forwarded-For"))
```

I think we can keep the var unexported.

(even tho I don't mind the function returning it)

VojtechVitek (approved):

VojtechVitek (approved):
Code LGTM.

I'll give this a try and test.

VojtechVitek (approved):
```go
r.Use(middleware.RealIPFromHeaders("CF-Connecting-IP", "X-True-Client-IP", "X-Real-IP", "X-Forwarded-For"))
```
works as expected

VojtechVitek:
This feature was reverted and never released. See https://github.com/go-chi/chi/pull/966 for more details.

PR #912, opened 2024-05-07
reduce size of Context struct from 216 bytes to 208 bytes

juburr:
optimize memory alignment of `Context` struct

### DIFF
--- a/context.go
+++ b/context.go
@@ -74,9 +74,8 @@ type Context struct {
 	// patterns across a stack of sub-routers.
 	RoutePatterns []string
 
-	// methodNotAllowed hint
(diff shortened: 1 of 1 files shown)


VojtechVitek:
every byte counts :) thanks!

dandee (commented):
It removes original comment that gives insight what this flag does.

BTW: is "methodTyp" type name correct or a typo? Should it read "methodType"?

pkieltyka:
its intentional :) just because "type" is a keyword, so just been using "typ", in this case, of course it doesn't apply, but its there for consistency

PR #913, opened 2024-05-09
Remove deprecated of package`io/ioutil`

JeaNile:


VojtechVitek:
Chi supports Go 1.14 in go.mod right now -- is this gonna work?

https://github.com/go-chi/chi/blob/master/go.mod

Issue #916, opened 2024-06-01
High CPU usage by chi.setPathValue


aadi-1024:
Hello

I'm not sure if this is an issue with Chi or somewhere in my own application, but the `setPathValue()` function seems to be causing very high CPU usage for me. It seems to be due to the hashing taking place in the function(pardon if I'm wrong) as I could infer from the pprof profile output.

![image](https://github.com/go-chi/chi/assets/90773264/7ce85a33-fbb3-42e0-8745-c5f303089428)

The generated chart:

![image](https://github.com/go-chi/chi/assets/90773264/a133f24e-7894-4081-ae5d-e409ffdfe190)

Any help appreciated!

aadi-1024:
![image](https://github.com/go-chi/chi/assets/90773264/e25d253b-0709-4c38-b7ec-457d7f01dd3d)

could this be the issue? all the entries are just `"*"`

aadi-1024:
A bit more digging around and it seems mounting a group with a prefix "/" causes this. I'll look into a bit more to see if I can fix this

VojtechVitek:
Any other volunteers to look into this? Or should we close?

aadi-1024:
I'll be looking into this again once I can manage some time :). Till then I suppose we can close this, I'll open a new issue if I'm able to solve it. Or maybe can just wait for other volunteers to chime in.

PR #918, opened 2024-06-04
WIP: fix for high CPU and memory usage

aadi-1024:
WIP!!

I further narrowed down the problem in #916 and turns out it only occurs when you have a Mount with "/" prefix and are trying to access an endpoint that does not exist, i.e. it should return 404. Instead it goes into an infinite loop and just keeps on increasing the memory and CPU usage.

This is my first time with such a project and I'm having quite the trouble getting hang of the codebase; while this does work it breaks a few tests related to URL Params. A bit of help would be really appreciated.

### DIFF
--- a/mux.go
+++ b/mux.go
@@ -338,8 +338,14 @@ func (mx *Mux) Mount(pattern string, handler http.Handler) {
 	if subroutes != nil {
 		method |= mSTUB
 	}
-	n := mx.handle(method, pattern+"*", mountHandler)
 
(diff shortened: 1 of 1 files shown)


aadi-1024:
sorry, started working recently, cant find time and motivation for this

VojtechVitek:
Thanks @aadi-1024

aadi-1024:
sorry I wasn't any help!

PR #934, opened 2024-07-13
Add RequestIDWithGenerator middleware

szinn:
The current implementation of RequestID the issue if the RequestID is added to the response header, then the hostname will be leaked

This PR extends the middleware with the ability to use an external generator to create the RequestID.

As well, SetReqID is exposed, allowing other API adapters (e.g., gprc) to set the request ID if needed.


szinn:
The other requestId solution should work - with regards to hostname, if requestId is available in the response, then it probably should not be present

PR #940, opened 2024-07-26
Fix `SupressNotFound` making subrouters 404

### LINKED ISSUE #939: Sub-routers returning 404 when using `middleware.SupressNotFound` on parent router
# Description
It seems that sub-routers are not properly matched after I applied `middleware.SupressNotFound`.
I investigated this and apparently `rctx.RoutePath` is modified without being restored in `middleware.SupressNotFound`.
This should be reproducible from the snippet I posted below.

# Action
## Expected Behavior
- Go to `http://localhost:8889/example/sub/hello`
  - I should get "Hello World".
  - Console should print "Route path /sub/hello"

## Actual
- Go to `http://localhost:8889/example/sub/hello`
  - Instead I got "404 page not found" with 404 status. This is returned by the sub router, not the middleware.
  - Console printed "Route path /hello"

*Actual result can be obtained by removing `middleware.SupressNotFound`.*

## Code
```go
package main

import (
	"fmt"
	"net/http"

	"github.com/go-chi/chi/v5"
)

func main() {
	mr := chi.NewRouter()

	mr.Route("/example", func(r chi.Router) {
		r.Use(middleware.SupressNotFound(mr))
		r.Use(func(h http.Handler) http.Handler {
				rctx := chi.RouteContext(r.Context())
				fmt.Println("Route path", rctx.RoutePath)
				h.ServeHTTP(w, r)
			})
		})

		r.Get("/other-hello", func(w http.ResponseWriter, r *http.Request) {
			w.Write([]byte("The Other Hello World"))
		})

		r.Route("/sub", func(r chi.Router) {
			r.Get("/hello", func(w http.ResponseWriter, r *http.Request) {
				w.Write([]byte("Hello World"))
			})
		})

	})

	http.ListenAndServe("0.0.0.0:8889", mr)
}
```

ezraisw:
This fixes #939 by restoring `rctx.RoutePath` to its initial value, before `Match()` was called.

### DIFF
--- a/middleware/supress_notfound.go
+++ b/middleware/supress_notfound.go
@@ -16,11 +16,20 @@ func SupressNotFound(router *chi.Mux) func(next http.Handler) http.Handler {
 	return func(next http.Handler) http.Handler {
 		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 			rctx := chi.RouteContext(r.Context())
+
+			// Match() updates rctx.RoutePath.
+			// Save the initial value to restore it later on for the next in chain.
+			initialRoutePath := rctx.RoutePath
+
 			match := rctx.Routes.Match(rctx, r.Method, r.URL.Path)
 			if !match {
 				router.NotFoundHandler().ServeHTTP(w, r)
 				return
 			}
+
+			// Restore after modification.
+			rctx.RoutePath = initialRoutePath
+
 			next.ServeHTTP(w, r)
 		})
 	}


johnwook:
wow, I was wondering why. Looks great!

devhaozi:
I'm having the same problem, @pkieltyka could you take a look at this?

devhaozi:
@VojtechVitek could you take a look at this?

devhaozi:
@VojtechVitek why hasn't this PR been merged yet?

jay-babu:
This is a pretty glaring bug. I unfortunately spent a couple of hours before realizing this issue

middleware/supress_notfound.go
```diff
@@ -16,11 +16,20 @@ func SupressNotFound(router *chi.Mux) func(next http.Handler) http.Handler {
...
+			initialRctx := *rctx
+
 			match := rctx.Routes.Match(rctx, r.Method, r.URL.Path)
 			if !match {
 				router.NotFoundHandler().ServeHTTP(w, r)
 				return
 			}
+
+			// Restore after modification.
+			*rctx = initialRctx
```
VojtechVitek, line 31: The only thing that I'm unsure about is why does the `.Match()` method change the `rctx` in the first place.

This looks like a nice bugfix/workaround. However, can we look into the root cause in the `.Match()` method, please? I wouldn't expect it to change the route context at all.

VojtechVitek: Never mind -- this is a documented behavior: https://github.com/go-chi/chi/blob/23c395f8524a30334126ca16fb4d37b88745b9b9/mux.go#L361-L362

VojtechVitek: I think the right fix is to use `NewRouteContext()` per the above comment.

VojtechVitek: @ezraisw thank you for the fix. Would you mind updating the PR to use `NewRouteContext()` instead of updating the top-level route context, please?

ezraisw:
FYI, this middleware returns 404 instead of 405 for requests with valid paths that have incorrect HTTP methods. Supporting 405 response would require a bigger change within the `chi.Mux` implementation.

middleware/supress_notfound.go
```diff
@@ -16,11 +16,22 @@ func SupressNotFound(router *chi.Mux) func(next http.Handler) http.Handler {
...
+
+			// Match() updates rctx.
+			// Use a temporary route context to look ahead.
+			tctx := chi.NewRouteContext()
+
+			// rctx.Routes is the original *chi.Mux.
+			// This starts a new traversal from the top.
+
+			// Unfortunately, this makes valid paths with incorrect methods return 404 instead of 405.
+			// TODO: Identify invalid path vs valid path with invalid method.
```
VojtechVitek, line 28: > FYI, this middleware returns 404 instead of 405 for requests with valid paths that have incorrect HTTP methods. Supporting 405 response would require a bigger change within the chi.Mux implementation.

@ezraisw Thanks for calling this out. Can we document this in the middleware's top-level comment and remove the TODO? I don't think it's worth the API change at this point.

PR #948, opened 2024-09-03
Set up go module for `_examples/versions`

hongkuancn:
At least `github.com/go-chi/render` needs to be downloaded to run the `_examples/versions`


VojtechVitek (approved):
LGTM. Thank you

PR #953, opened 2024-09-17
Replace deprecated ioutil calls with io package

marcofranssen:


VojtechVitek (changes requested):
Chi is currently compatible with Go 1.14+. See https://github.com/go-chi/chi/blob/master/go.mod#L3

This PR would effectively drop support for older Go versions. It's not worth the changes.

VojtechVitek:
Hi, feel free to submit an issue if you're proposing to drop support for older Go versions.

PR #962, opened 2024-12-03
chore: delint ioutil usage

costela:
This also bumps the minimum supported version of Go to 1.16 (which itself has beel EOL for almost 3 years now :see_no_evil: )

Refs #963


go.mod
```diff
@@ -1,3 +1,3 @@
 module github.com/go-chi/chi/v5
 
-go 1.14
+go 1.16
```
VojtechVitek, line 3: This downgrades chi's compatibility with older Go versions.

Can you please raise an issue first, so we can discuss pros/cons of dropping support for Go 1.14-1.15 and gather feedback from the community?

I don't disagree btw. It might be the right time to start requiring Go 1.16+.

Thanks!

costela: sure! 👍🏻

costela: done here: https://github.com/go-chi/chi/issues/963

VojtechVitek (approved):
LGTM

PR #964, opened 2024-12-05
Add tracing middleware and helper function

c2h5oh:
To get tracing across services one has to propagate tracing headers downstream. 

This PR adds a middleware copies tracing headers into request context and provides a function to add those headers to any requests you are creating to other services without having to pass around the entire request - context is enough.

### DIFF
--- a/middleware/tracing.go
+++ b/middleware/tracing.go
@@ -1,0 +1,47 @@
+package middleware
+
+import (
+	"context"
+	"net/http"
+)
+
+var (
+	TracingCtxKey       = &contextKey{"TracingHeaders"}
+	defaultTraceHeaders = []string{
+		"X-Cloud-Trace-Context", // GCP load balancer
+		"X-Amzn-Trace-Id",       // AWS X-ray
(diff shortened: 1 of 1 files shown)


middleware/tracing.go
```diff
@@ -0,0 +1,49 @@
...
+		http.CanonicalHeaderKey("X-Amzn-Trace-Id"),       // AWS X-ray
+		http.CanonicalHeaderKey("CF-ray"),                // Cloudflare ray id
+		http.CanonicalHeaderKey("TraceId"),               // https://github.com/go-chi/traceid
+	}
+)
+
+func Tracing(traceHeaders ...string) func(next http.Handler) http.Handler {
+	if len(traceHeaders) == 0 {
+		traceHeaders = defaultTraceHeaders
+	}
```
Pipello, line 21: I would not put any defaults here but rather export the `defaultTraceHeaders` so from caller perspective it is clearer what is being done.

Pipello (commented):
Could you add a comments for the documentation? 

It would also be nice to add some unit tests

Issue #711, opened 2022-03-04
Problems with use of X-Forwarded-For and other headers to get the "real" client IP


adam-p:
go-chi/chi/middleware/RealIP and go-chi/httprate suffer from problems getting the "real" client IP. I'll try to break it down into categories.

RealIP isn't necessarily "security-related", but it depends on how the dev uses it. I would argue that rate limiting -- and so httprate -- _is_ security-related.

For a full explanation of the problems mentioned here, please see [my blog post](https://adam-p.ca/blog/2022/03/x-forwarded-for/) about it. (Here's the [chi-specific section](https://adam-p.ca/blog/2022/03/x-forwarded-for/#go-chichi), but all the rest is informative.)

## Headers are untrustworthy, unless added by a reverse proxy you control

 The RealIP middleware has [this comment](https://github.com/go-chi/chi/blob/c97bc988430d623a14f50b7019fb40529036a35a/middleware/realip.go#L24-L30):

> You should only use this middleware if you can trust the headers passed to you (in particular, the two [three, actually] headers this middleware uses), for example because you have placed a reverse proxy like HAProxy or nginx in front of Chi. If your reverse proxies are configured to pass along arbitrary header values from the client, or if you use this middleware without a reverse proxy, malicious clients will be able to make you very sad (or, depending on how you're using RemoteAddr, vulnerable to an attack of some sort).

This is a good warning, but a) there's no such warning on chi's [httprate.KeyByIP](https://github.com/go-chi/httprate/blob/463924d478ea0c19de7265b97371a59a7ebf5fdd/httprate.go#L25) (even though it does the same thing), and b) it's not a comprehensive enough warning.

Some reverse proxies, like AWS ALB, lets all header values through that it doesn't set itself. So it will let through `True-Client-IP` and `X-Real-IP`. An attacker can spoof either of those headers, and that will end up as the IP that is reported by RealIP or the IP that's gets rate-limited by httprate.

### Leftmost vs Rightmost

Also, `X-Forwarded-For` is only _appended to_ by reverse proxies, so the only IPs in it that are trustworthy are the ones that were added by reverse proxies that you control. httprate, however, takes the [_first_ IP in XFF](https://github.com/go-chi/httprate/blob/463924d478ea0c19de7265b97371a59a7ebf5fdd/httprate.go#L37). So an attacker can trivially spoof it. The leftmost XFF IP should never be used for security-related purposes.

### Switching to rightmost will introduce an XFF multiple-headers problem

[...]

Issue #708, opened 2022-02-21
Add more sophisticated realip middleware


jawnsy:
Thanks so much for this excellent library! I would like to propose the addition of a more sophisticated realip middleware, perhaps as a separate project, similar to [`httprate`](https://github.com/go-chi/httprate).

I'd be happy to open a pull request or contribute a repository, but wanted to check with you first. Thanks in advance for your consideration!

## Current behavior

The [included `realip` middleware](https://github.com/go-chi/chi/blob/master/middleware/realip.go) is a very simple implementation, and serves as a great example for implementing custom middlewares. However, there are some limitations (by design):

1. It allows spoofing if clients are able to access the chi router directly, bypassing the frontend reverse proxy
   
   If the client connects directly to the router and passes `X-Forwarded-For: 127.0.0.1`, then `chi` will update the `RemoteAddr` to `127.0.0.1`

1. It allows spoofing if the frontend reverse proxy does not appropriately sanitize (filter out) other headers
   
   If the client connects to a reverse proxy and passes `X-Forwarded-For: 127.0.0.1` and `X-Real-IP: 10.0.0.1`, and the reverse proxy is configured to ignore/replace `X-Forwarded-For`, but not configured to filter out `X-Real-IP`, then the proxy will pass through: `X-Forwarded-For: 100.100.100.100` (the real client address) as well as `X-Real-IP: 10.0.0.1`, and due to the search order, the untrusted `X-Real-IP` will take precedence over the correct `X-Forwarded-For` header.

1. It does not indicate whether the original client request protocol was HTTP or HTTPS
   
   If the client connects to the reverse proxy over HTTPS and the reverse proxy terminates TLS, then it is not possible to tell that the client connected over TLS, since [`X-Forwarded-Proto`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-Proto) is not considered by the middleware. This means that the server may use incorrect logic, omitting HTTPS-only headers like [`Strict-Transport-Security`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Strict-Transport-Security), avoiding [Secure cookies](https://developer.mozilla.org/en-US/docs/Web/HTTP/Cookies#restrict_access_to_cookies), etc.

1. It does not preserve the original `RemoteAddr` information for downstream clients
   
   If clients need to get the original `RemoteAddr` (the address/port of the reverse proxy), there is currently no way to get that information.

[...]

Issue #963, opened 2024-12-03
proposal: drop support for older go versions


costela:
Chi currently supports go down to 1.14, which has been EOL for almost 4 years.

This means keeping "code smells" around (e.g. `ioutil`) and blocking access to some newer features (e.g. `ResponseController`).

I propose bumping the minimal version to allow for more aggressive deprecations and feature adoption.

Concretely, I'd propose matching go's support policy of "latest+2", but would be happy with any other such dynamic approach, so that there's 
1) a clearly defined expectation of support
2) no need to periodically rehash the support discussion

WDYT?

VojtechVitek:
Good points, I agree. I'd vote to drop support for Go 1.17 and older.

Go 1.18 would be enough to replace the deprecated `io/ioutil` functions, and so we could start using `net/netip` pkg.

VojtechVitek:
> I'd propose matching go's support policy of "latest+2", but would be happy with any other such dynamic approach, so that there's

This is actually a great idea. Even if we go for "latest+4" to be more conservative, we'd still support Go 1.20 as of today, which should be fine for most of the users. 👍

c2h5oh:
I think we should go a little bit past EOL cutoff, because there are some environments that are slow to adopt new version of go and even slower to remove support for older ones, most notably Google App Engine.

Latest +3 or +4 sounds like a reasonable compromise.

costela:
> I think we should go a little bit past EOL cutoff, because there are some environments that are slow to adopt new version of go and even slower to remove support for older ones, most notably Google App Engine.

Just FTR, it seems app engine isn't that far from the upstream support timeline. Latest 1.23 is still in preview, but [1.20 has already been EOLd](https://cloud.google.com/appengine/docs/standard/lifecycle/support-schedule#go):
![image](https://github.com/user-attachments/assets/2422fec7-781f-441e-9101-aa63bd4fc812)

And just as a devil's advocate, I see two more options:
- Telling users on older runtimes to stick with a slightly older chi version if runtime support is more important than new features.
- Supporting latest+2 on the `master`, but keeping the door open for security-fix releases up until +X (via minor releases from other branches e.g. `5.1.x`). 

But in summary, yes, I'd be totally ok with just +4 overall as well.

pkieltyka:
good discussion. I think latest +4 is a good place to be in general. I’d say for short term if we broke the rule to +3 so we could bump up to 1.20 given that 1.24 is coming out soon and we’d then be at latest+4.

[...]

PR #967, opened 2024-12-15
feat: middleware.ClientIP, a replacement for middleware.RealIP

VojtechVitek:
A replacement for `middleware.RealIP` that closes the three open spoofing advisories against it:
- [GHSA-9g5q-2w5x-hmxf](https://github.com/go-chi/chi/security/advisories/GHSA-9g5q-2w5x-hmxf) — IP spoofing via XFF in `RemoteAddr` resolution (convto)
- [GHSA-rjr7-jggh-pgcp](https://github.com/go-chi/chi/security/advisories/GHSA-rjr7-jggh-pgcp) — RealIP allows IP spoofing via unvalidated XFF (rezmoss)
- [GHSA-3fxj-6jh8-hvhx](https://github.com/go-chi/chi/security/advisories/GHSA-3fxj-6jh8-hvhx) — IP spoofing in `middleware.RealIP` (Saku0512, Critical / 9.3)

It also addresses issues outlined at:
- https://github.com/go-chi/chi/issues/708
- https://adam-p.ca/blog/2022/03/x-forwarded-for/
- https://github.com/go-chi/chi/issues/711
- https://github.com/go-chi/chi/issues/453
- https://github.com/go-chi/chi/pull/908

`middleware.RealIP` is deprecated in this PR with pointers to the new API.

The deprecation only adds a `// Deprecated:` doc comment; the function keeps working for backward compatibility.

### Why a new middleware (not "fix RealIP in place")
`RealIP` has two unfixable design choices: it mutates `r.RemoteAddr`, and it tries to be a one-size-fits-all default by walking a hard-coded list of headers any client can supply. Per [adam-p's "The perils of the 'real' client IP"](https://adam-p.ca/blog/2022/03/x-forwarded-for/) (which calls chi out by name on this), there is no safe default — the user must pick their trust source explicitly.

### The new API
Four middlewares, two accessors. Pick exactly one middleware based on your
infrastructure, read the result with one of the two accessors:

```go
// One of the four. There is no safe default — pick exactly one.
func ClientIPFromHeader(trustedHeader string) func(http.Handler) http.Handler
func ClientIPFromXFF(trustedIPPrefixes ...string) func(http.Handler) http.Handler
func ClientIPFromXFFTrustedProxies(numTrustedProxies int) func(http.Handler) http.Handler
func ClientIPFromRemoteAddr(h http.Handler) http.Handler

// Read the result.
func GetClientIP(ctx context.Context) string         // for logs, rate-limit keys
func GetClientIPAddr(ctx context.Context) netip.Addr // for typed work
```

## Example usage:
```go
// Pick a single ClientIP middleware based on your deployment
  
// Cloudflare.
r.Use(middleware.ClientIPFromHeader("CF-Connecting-IP"))

// Nginx with ngx_http_realip_module.
r.Use(middleware.ClientIPFromHeader("X-Real-IP"))

// Apache with mod_remoteip.
r.Use(middleware.ClientIPFromHeader("X-Client-IP"))

// AWS CloudFront, or any proxy fleet with known CIDRs.
r.Use(middleware.ClientIPFromXFF(
    "13.32.0.0/15",   // CloudFront IPv4
    "52.46.0.0/18",   // CloudFront IPv4
    "2600:9000::/28", // CloudFront IPv6
))

// Behind exactly 2 trusted proxies with dynamic IPs (autoscaling pools,
// ephemeral containers, dynamic CDN edges).
r.Use(middleware.ClientIPFromXFFTrustedProxies(2))

// Server directly on the public internet, no proxy in front.
r.Use(middleware.ClientIPFromRemoteAddr)
```

And in your handler or downstream middleware:
```go
clientIP := middleware.GetClientIP(r.Context())
// log it, use it as a rate-limit key, etc.
```
---

Thanks to @adam-p, @c2h5oh, @rezmoss, @Saku0512, @convto, @Dirbaio, @jawnsy, @lrstanley, @mfridman, @n33pm, @pkieltyka for the prior discussions, detailed reviews, advisory reports, and test contributions that shaped this PR.

### DIFF
--- a/middleware/client_ip.go
+++ b/middleware/client_ip.go
@@ -1,0 +1,185 @@
+package middleware
+
+import (
+	"context"
+	"net"
+	"net/http"
+	"net/netip"
+	"strings"
+)
+
+var (
+	// clientIPCtxKey is the context key used to store the client IP address.
+	clientIPCtxKey = &contextKey{"clientIP"}
+)
+
+// ClientIPFromHeader parses the client IP address from a specified HTTP header
+// (e.g., X-Real-IP, CF-Connecting-IP) and injects it into the request context
+// if it is not already set. The parsed IP address can be retrieved using GetClientIP().
+//
+// The middleware validates the IP address to ignore loopback, private, and unspecified addresses.
+//
+// ### Important Notice:
+// - Use this middleware only when your infrastructure sets a trusted header containing the client IP.
+// - If the specified header is not securely set by your infrastructure, malicious clients could spoof it.
+//
+// Example trusted headers:
+// - "X-Real-IP"        - Nginx (ngx_http_realip_module)
+// - "X-Client-IP"      - Apache (mod_remoteip)
+// - "CF-Connecting-IP" - Cloudflare
+// - "True-Client-IP"   - Akamai, Cloudflare Enterprise
+// - "X-Azure-ClientIP" - Azure Front Door
+// - "Fastly-Client-IP" - Fastly
+func ClientIPFromHeader(trustedHeader string) func(http.Handler) http.Handler {
+	return func(h http.Handler) http.Handler {
+		fn := func(w http.ResponseWriter, r *http.Request) {
+			ctx := r.Context()
+
+			// Check if the client IP is already set in the context.
+			if _, ok := ctx.Value(clientIPCtxKey).(netip.Addr); ok {
+				h.ServeHTTP(w, r)
+				return
+			}
+
+			// Parse the IP address from the trusted header.
+			ip, err := netip.ParseAddr(r.Header.Get(trustedHeader))
+			if err != nil || ip.IsLoopback() || ip.IsUnspecified() || ip.IsPrivate() {
+				// Ignore invalid or private IPs.
+				h.ServeHTTP(w, r)
+				return
+			}
+
+			// Store the valid client IP in the context.
+			ctx = context.WithValue(ctx, clientIPCtxKey, ip)
+			h.ServeHTTP(w, r.WithContext(ctx))
+		}
+		return http.HandlerFunc(fn)
+	}
+}
+
+// ClientIPFromXFFHeader parses the client IP address from the X-Forwarded-For
+// header and injects it into the request context if it is not already set. The
+// parsed IP address can be retrieved using GetClientIP().
+//
+// The middleware traverses the X-Forwarded-For chain (rightmost untrusted IP)
+// and excludes loopback, private, unspecified, and trusted IP ranges.
+//
+// ### Important Notice:
+// - Use this middleware only when your infrastructure sets and validates the X-Forwarded-For header.
+// - Malicious clients can spoof the header unless a trusted reverse proxy or load balancer sanitizes it.
+//
+// Parameters:
+// - `trustedIPPrefixes`: A list of CIDR prefixes that define trusted proxy IP ranges.
+//
+// Example trusted IP ranges:
+// - "203.0.113.0/24"     - Example corporate proxy
+// - "198.51.100.0/24"    - Example data center or hosting provider
+// - "2400:cb00::/32"     - Cloudflare IPv6 range
+// - "2606:4700::/32"     - Cloudflare IPv6 range
+// - "192.0.2.0/24"       - Example VPN gateway
+//
+// Note: Private IP ranges (e.g., "10.0.0.0/8", "192.168.0.0/16", "172.16.0.0/12")
+// are automatically excluded by netip.Addr.IsPrivate() and do not need to be added here.
+func ClientIPFromXFFHeader(trustedIPPrefixes ...string) func(http.Handler) http.Handler {
+	// Pre-parse trusted prefixes.
+	trustedPrefixes := make([]netip.Prefix, len(trustedIPPrefixes))
+	for i, ipRange := range trustedIPPrefixes {
+		trustedPrefixes[i] = netip.MustParsePrefix(ipRange)
+	}
+
+	return func(h http.Handler) http.Handler {
+		fn := func(w http.ResponseWriter, r *http.Request) {
+			ctx := r.Context()
+
+			// Check if the client IP is already set in the context.
+			if _, ok := ctx.Value(clientIPCtxKey).(netip.Addr); ok {
+				h.ServeHTTP(w, r)
+				return
+			}
+
+			// Parse and split the X-Forwarded-For header(s).
+			xff := strings.Split(strings.Join(r.Header.Values("X-Forwarded-For"), ","), ",")
+		nextValue:
+			for i := len(xff) - 1; i >= 0; i-- {
+				ip, err := netip.ParseAddr(strings.TrimSpace(xff[i]))
+				if err != nil {
+					continue
+				}
+
+				// Ignore loopback, private, or unspecified addresses.
+				if ip.IsLoopback() || ip.IsPrivate() || ip.IsUnspecified() {
+					continue
+				}
+
+				// Ignore trusted IPs within the given ranges.
+				for _, prefix := range trustedPrefixes {
+					if prefix.Contains(ip) {
+						continue nextValue
+					}
+				}
+
+				// Store the valid client IP in the context.
+				ctx = context.WithValue(ctx, clientIPCtxKey, ip)
+				h.ServeHTTP(w, r.WithContext(ctx))
+				return
+			}
+
+			h.ServeHTTP(w, r)
+		}
+		return http.HandlerFunc(fn)
+	}
+}
+
+// ClientIPFromRemoteAddr extracts the client IP address from the RemoteAddr
+// field of the HTTP request and injects it into the request context if it is
+// not already set. The parsed IP address can be retrieved using GetClientIP().
+//
+// The middleware ignores invalid or private IPs.
+//
+// ### Use Case:
+// This middleware is useful when the client IP cannot be determined from headers
+// such as X-Forwarded-For or X-Real-IP, and you need to fall back to RemoteAddr.
+func ClientIPFromRemoteAddr(h http.Handler) http.Handler {
+	fn := func(w http.ResponseWriter, r *http.Request) {
+		ctx := r.Context()
+
+		// Check if the client IP is already set in the context.
+		if _, ok := ctx.Value(clientIPCtxKey).(netip.Addr); ok {
+			h.ServeHTTP(w, r)
+			return
+		}
+
+		// Extract the IP from request RemoteAddr.
+		host, _, err := net.SplitHostPort(r.RemoteAddr)
+		if err != nil {
+			h.ServeHTTP(w, r)
+			return
+		}
+
+		ip, err := netip.ParseAddr(host)
+		if err != nil {
+			h.ServeHTTP(w, r)
+			return
+		}
+
+		// Store the valid client IP in the context.
+		ctx = context.WithValue(ctx, clientIPCtxKey, ip)
+		h.ServeHTTP(w, r.WithContext(ctx))
+	}
+	return http.HandlerFunc(fn)
+}
+
+// GetClientIP retrieves the client IP address from the given context.
+// The IP address is set by one of the following middlewares:
+// - ClientIPFromHeader
+// - ClientIPFromXFFHeader
+// - ClientIPFromRemoteAddr
+//
+// Returns an empty string if no valid IP is found.
+func GetClientIP(ctx context.Context) string {
+	ip, ok := ctx.Value(clientIPCtxKey).(netip.Addr)
+	if !ok || !ip.IsValid() {
+		return ""
+	}
+	return ip.String()
+}
--- a/middleware/client_ip.go
+++ b/middleware/client_ip.go
@@ -119,9 +119,8 @@ func ClientIPFromXFFTrustedProxies(numTrustedProxies int) func(http.Handler) htt
 	}
 	return func(h http.Handler) http.Handler {
 		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-			xff := mergeXFF(r.Header.Values(xForwardedForHeader))
-			if i := len(xff) - numTrustedProxies; i >= 0 {
-				if ip, ok := parseHeaderAddr(xff[i]); ok {
+			if v := nthFromRightXFF(r.Header.Values(xForwardedForHeader), numTrustedProxies); v != "" {
+				if ip, ok := parseHeaderAddr(v); ok {
 					r = r.WithContext(context.WithValue(r.Context(), clientIPCtxKey, ip))
 				}
 			}
@@ -175,24 +174,39 @@ func GetClientIPAddr(ctx context.Context) netip.Addr {
 	return ip
 }
 
-// mergeXFF merges all X-Forwarded-For header values into a single ordered
-// list of trimmed entries (left-to-right, in the order received). Empty
-// entries are dropped. Entries are not validated as IPs here.
+// nthFromRightXFF returns the trimmed entry at position n from the right
+// end of the merged X-Forwarded-For chain (n=1 is the rightmost), or "" if
+// the chain has fewer than n non-empty entries.
 //
-// Merging all instances is required for security: per RFC 2616, multiple
-// XFF headers MUST be processed in order. Reading only the first or last
-// header value lets an attacker pick which value security logic sees by
-// sending duplicate headers.
-func mergeXFF(headers []string) []string {
-	out := make([]string, 0, len(headers))
-	for _, h := range headers {
-		for _, v := range strings.Split(h, ",") {
-			if v = strings.TrimSpace(v); v != "" {
-				out = append(out, v)
+// Empty/whitespace entries are dropped and do NOT count toward n, so an
+// attacker can't slide their chosen value into the trusted slot by
+// prepending commas.
+//
+// Multi-header merge is required for security per RFC 2616: multiple XFF
+// headers MUST be processed in order received; reading only the first or
+// last lets an attacker pick which value security logic sees by sending
+// duplicate headers. Lazy walk, zero allocations.
+func nthFromRightXFF(headers []string, n int) string {
+	for hi := len(headers) - 1; hi >= 0; hi-- {
+		h := headers[hi]
+		for h != "" {
+			var v string
+			if i := strings.LastIndexByte(h, ','); i >= 0 {
+				v, h = h[i+1:], h[:i]
+			} else {
+				v, h = h, ""
+			}
+			v = strings.TrimSpace(v)
+			if v == "" {
+				continue
+			}
+			n--
+			if n == 0 {
+				return v
 			}
 		}
 	}
-	return out
+	return ""
 }
 
 // rightmostUntrustedXFF walks all X-Forwarded-For header values right-to-left
@@ -202,10 +216,11 @@ func mergeXFF(headers []string) []string {
 // an unparseable hop is indistinguishable from a hostile or missing one and
 // we cannot safely keep walking past it.
 //
-// Walks the headers lazily rather than calling [mergeXFF] so the common case
-// — one trusted hop, the rightmost entry is the client — returns on the very
-// first iteration without materializing the full chain. Zero allocations
-// beyond what [netip.ParseAddr] does internally.
+// Walks the headers lazily so the common case — one trusted hop, rightmost
+// entry is the client — returns on the very first iteration without
+// materializing the full chain. Zero allocations beyond what
+// [netip.ParseAddr] does internally. See also [nthFromRightXFF], which uses
+// the same iteration shape with a count-down stop condition.
 func rightmostUntrustedXFF(headers []string, trustedPrefixes []netip.Prefix) (netip.Addr, bool) {
 	for hi := len(headers) - 1; hi >= 0; hi-- {
 		h := headers[hi]
--- a/middleware/client_ip.go
+++ b/middleware/client_ip.go
@@ -92,8 +92,7 @@ func ClientIPFromXFF(trustedIPPrefixes ...string) func(http.Handler) http.Handle
 // with explicit trusted CIDRs whenever you can.
 //
 // If the XFF chain has fewer than numTrustedProxies entries (header missing
-// or architecture changed), no client IP is set; chain [ClientIPFromRemoteAddr]
-// after this middleware if you want a fallback.
+// or architecture changed), no client IP is set and [GetClientIP] returns "".
 //
 // Panics at startup if numTrustedProxies < 1.
 func ClientIPFromXFFTrustedProxies(numTrustedProxies int) func(http.Handler) http.Handler {
--- a/middleware/client_ip.go
+++ b/middleware/client_ip.go
@@ -90,7 +90,7 @@ func ClientIPFromXFF(trustedIPPrefixes ...string) func(http.Handler) http.Handle
 	return func(h http.Handler) http.Handler {
 		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 			var found netip.Addr
-			walkXFF(r.Header.Values(xForwardedForHeader), func(v string) bool {
+			walkXFF(r.Header[xForwardedForHeader], func(v string) bool {
 				ip, ok := parseHeaderAddr(v)
 				if !ok {
 					return true // fail-closed; leave found unset
@@ -143,7 +143,7 @@ func ClientIPFromXFFTrustedProxies(numTrustedProxies int) func(http.Handler) htt
 		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 			n := numTrustedProxies
 			var entry string
-			walkXFF(r.Header.Values(xForwardedForHeader), func(v string) bool {
+			walkXFF(r.Header[xForwardedForHeader], func(v string) bool {
 				n--
 				if n == 0 {
 					entry = v
--- a/middleware/client_ip.go
+++ b/middleware/client_ip.go
@@ -192,17 +192,34 @@ func mergeXFF(headers []string) []string {
 	return out
 }
 
-// rightmostUntrustedXFF walks merged XFF right-to-left, skipping IPs that
-// match trustedPrefixes (and unparseable entries), and returns the first
-// remaining valid IP.
+// rightmostUntrustedXFF walks all X-Forwarded-For header values right-to-left
+// across the merged chain, skipping IPs that match trustedPrefixes (and
+// unparseable / empty entries), and returns the first remaining valid IP.
+//
+// Walks the headers lazily rather than calling [mergeXFF] so the common case
+// — one trusted hop, the rightmost entry is the client — returns on the very
+// first iteration without materializing the full chain. Zero allocations
+// beyond what [netip.ParseAddr] does internally.
 func rightmostUntrustedXFF(headers []string, trustedPrefixes []netip.Prefix) (netip.Addr, bool) {
-	xff := mergeXFF(headers)
-	for i := len(xff) - 1; i >= 0; i-- {
-		ip, ok := parseHeaderAddr(xff[i])
-		if !ok || inAnyPrefix(ip, trustedPrefixes) {
-			continue
+	for hi := len(headers) - 1; hi >= 0; hi-- {
+		h := headers[hi]
+		for h != "" {
+			var v string
+			if i := strings.LastIndexByte(h, ','); i >= 0 {
+				v, h = h[i+1:], h[:i]
+			} else {
+				v, h = h, ""
+			}
+			v = strings.TrimSpace(v)
+			if v == "" {
+				continue
+			}
+			ip, ok := parseHeaderAddr(v)
+			if !ok || inAnyPrefix(ip, trustedPrefixes) {
+				continue
+			}
+			return ip, true
 		}
-		return ip, true
 	}
 	return netip.Addr{}, false
 }
--- a/middleware/client_ip.go
+++ b/middleware/client_ip.go
@@ -8,178 +8,215 @@ import (
 	"strings"
 )
 
-var (
-	// clientIPCtxKey is the context key used to store the client IP address.
-	clientIPCtxKey = &contextKey{"clientIP"}
-)
+// clientIPCtxKey stores the client IP set by any of the ClientIPFrom* middlewares.
+var clientIPCtxKey = &contextKey{"clientIP"}
 
-// ClientIPFromHeader parses the client IP address from a specified HTTP header
-// (e.g., X-Real-IP, CF-Connecting-IP) and injects it into the request context
-// if it is not already set. The parsed IP address can be retrieved using GetClientIP().
+// ClientIPFromHeader stores the client IP read from a single-IP HTTP header
+// set by your reverse proxy. Read the IP with [GetClientIP].
 //
-// The middleware validates the IP address to ignore loopback, private, and unspecified addresses.
+// Use this when your reverse proxy sets one of these headers for every
+// request and overwrites any client-supplied value:
 //
-// ### Important Notice:
-// - Use this middleware only when your infrastructure sets a trusted header containing the client IP.
-// - If the specified header is not securely set by your infrastructure, malicious clients could spoof it.
+//   - X-Real-IP        — Nginx with ngx_http_realip_module
+//   - X-Client-IP      — Apache with mod_remoteip
+//   - CF-Connecting-IP — Cloudflare
 //
-// Example trusted headers:
-// - "X-Real-IP"        - Nginx (ngx_http_realip_module)
-// - "X-Client-IP"      - Apache (mod_remoteip)
-// - "CF-Connecting-IP" - Cloudflare
-// - "True-Client-IP"   - Akamai, Cloudflare Enterprise
-// - "X-Azure-ClientIP" - Azure Front Door
-// - "Fastly-Client-IP" - Fastly
+// DO NOT use this with headers your infrastructure does not overwrite
+// (True-Client-IP, X-Azure-ClientIP, Fastly-Client-IP by default, etc.).
+// Those can be supplied by the client and are trivially spoofable.
 func ClientIPFromHeader(trustedHeader string) func(http.Handler) http.Handler {
 	return func(h http.Handler) http.Handler {
-		fn := func(w http.ResponseWriter, r *http.Request) {
-			ctx := r.Context()
-
-			// Check if the client IP is already set in the context.
-			if _, ok := ctx.Value(clientIPCtxKey).(netip.Addr); ok {
+		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			if hasClientIP(r.Context()) {
 				h.ServeHTTP(w, r)
 				return
 			}
-
-			// Parse the IP address from the trusted header.
-			ip, err := netip.ParseAddr(r.Header.Get(trustedHeader))
-			if err != nil || ip.IsLoopback() || ip.IsUnspecified() || ip.IsPrivate() {
-				// Ignore invalid or private IPs.
-				h.ServeHTTP(w, r)
-				return
+			if ip, err := netip.ParseAddr(r.Header.Get(trustedHeader)); err == nil {
+				r = r.WithContext(context.WithValue(r.Context(), clientIPCtxKey, ip))
 			}
-
-			// Store the valid client IP in the context.
-			ctx = context.WithValue(ctx, clientIPCtxKey, ip)
-			h.ServeHTTP(w, r.WithContext(ctx))
-		}
-		return http.HandlerFunc(fn)
-	}
-}
-
-// ClientIPFromXFFHeader parses the client IP address from the X-Forwarded-For
-// header and injects it into the request context if it is not already set. The
-// parsed IP address can be retrieved using GetClientIP().
-//
-// The middleware traverses the X-Forwarded-For chain (rightmost untrusted IP)
-// and excludes loopback, private, unspecified, and trusted IP ranges.
-//
-// ### Important Notice:
-// - Use this middleware only when your infrastructure sets and validates the X-Forwarded-For header.
-// - Malicious clients can spoof the header unless a trusted reverse proxy or load balancer sanitizes it.
-//
-// Parameters:
-// - `trustedIPPrefixes`: A list of CIDR prefixes that define trusted proxy IP ranges.
-//
-// Example trusted IP ranges:
-// - "203.0.113.0/24"     - Example corporate proxy
-// - "198.51.100.0/24"    - Example data center or hosting provider
-// - "2400:cb00::/32"     - Cloudflare IPv6 range
-// - "2606:4700::/32"     - Cloudflare IPv6 range
-// - "192.0.2.0/24"       - Example VPN gateway
-//
-// Note: Private IP ranges (e.g., "10.0.0.0/8", "192.168.0.0/16", "172.16.0.0/12")
-// are automatically excluded by netip.Addr.IsPrivate() and do not need to be added here.
-func ClientIPFromXFFHeader(trustedIPPrefixes ...string) func(http.Handler) http.Handler {
-	// Pre-parse trusted prefixes.
-	trustedPrefixes := make([]netip.Prefix, len(trustedIPPrefixes))
-	for i, ipRange := range trustedIPPrefixes {
-		trustedPrefixes[i] = netip.MustParsePrefix(ipRange)
-	}
-
-	return func(h http.Handler) http.Handler {
-		fn := func(w http.ResponseWriter, r *http.Request) {
-			ctx := r.Context()
-
-			// Check if the client IP is already set in the context.
-			if _, ok := ctx.Value(clientIPCtxKey).(netip.Addr); ok {
-				h.ServeHTTP(w, r)
-				return
-			}
-
-			// Parse and split the X-Forwarded-For header(s).
-			xff := strings.Split(strings.Join(r.Header.Values("X-Forwarded-For"), ","), ",")
-		nextValue:
-			for i := len(xff) - 1; i >= 0; i-- {
-				ip, err := netip.ParseAddr(strings.TrimSpace(xff[i]))
-				if err != nil {
-					continue
-				}
-
-				// Ignore loopback, private, or unspecified addresses.
-				if ip.IsLoopback() || ip.IsPrivate() || ip.IsUnspecified() {
-					continue
-				}
-
-				// Ignore trusted IPs within the given ranges.
-				for _, prefix := range trustedPrefixes {
-					if prefix.Contains(ip) {
-						continue nextValue
-					}
-				}
-
-				// Store the valid client IP in the context.
-				ctx = context.WithValue(ctx, clientIPCtxKey, ip)
-				h.ServeHTTP(w, r.WithContext(ctx))
-				return
-			}
-
 			h.ServeHTTP(w, r)
-		}
-		return http.HandlerFunc(fn)
+		})
 	}
 }
 
-// ClientIPFromRemoteAddr extracts the client IP address from the RemoteAddr
-// field of the HTTP request and injects it into the request context if it is
-// not already set. The parsed IP address can be retrieved using GetClientIP().
+// ClientIPFromXFF stores the client IP read from the X-Forwarded-For header,
+// walking the chain right-to-left and skipping any IP that falls within one
+// of the given trusted CIDR prefixes. The first IP that is not trusted is
+// the client. Read it with [GetClientIP].
 //
-// The middleware ignores invalid or private IPs.
+// Use this when you sit behind one or more reverse proxies whose IP ranges
+// you can enumerate as CIDRs:
 //
-// ### Use Case:
-// This middleware is useful when the client IP cannot be determined from headers
-// such as X-Forwarded-For or X-Real-IP, and you need to fall back to RemoteAddr.
-func ClientIPFromRemoteAddr(h http.Handler) http.Handler {
-	fn := func(w http.ResponseWriter, r *http.Request) {
-		ctx := r.Context()
+//	r.Use(middleware.ClientIPFromXFF(
+//	    "13.32.0.0/15",   // CloudFront IPv4
+//	    "52.46.0.0/18",   // CloudFront IPv4
+//	    "2600:9000::/28", // CloudFront IPv6
+//	))
+//
+// Calling with no arguments returns the rightmost parseable XFF IP — safe
+// only if you have exactly one trusted hop directly in front of this server
+// (e.g., nginx on localhost).
+//
+// If you know the number of trusted proxies but not their IPs, use
+// [ClientIPFromXFFTrustedProxies] instead.
+//
+// Panics at startup if any prefix is invalid.
+func ClientIPFromXFF(trustedIPPrefixes ...string) func(http.Handler) http.Handler {
+	prefixes := make([]netip.Prefix, len(trustedIPPrefixes))
+	for i, p := range trustedIPPrefixes {
+		prefixes[i] = netip.MustParsePrefix(p)
+	}
+	return func(h http.Handler) http.Handler {
+		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			if hasClientIP(r.Context()) {
+				h.ServeHTTP(w, r)
+				return
+			}
+			if ip, ok := rightmostUntrustedXFF(r.Header.Values("X-Forwarded-For"), prefixes); ok {
+				r = r.WithContext(context.WithValue(r.Context(), clientIPCtxKey, ip))
+			}
+			h.ServeHTTP(w, r)
+		})
+	}
+}
 
-		// Check if the client IP is already set in the context.
-		if _, ok := ctx.Value(clientIPCtxKey).(netip.Addr); ok {
+// ClientIPFromXFFTrustedProxies stores the client IP read from the
+// X-Forwarded-For header, given the exact number of trusted reverse proxies
+// between this server and the public internet. It returns the IP at position
+// len(xff) - numTrustedProxies in the merged X-Forwarded-For list — the IP
+// added by the outermost of your trusted proxies, the only IP in the chain
+// that none of your proxies have allowed an attacker to forge. Read it with
+// [GetClientIP].
+//
+// Use this when:
+//   - You know exactly how many proxies you sit behind, AND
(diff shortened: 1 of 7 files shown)


convto (approved):
I’ve checked the implementation, and I completely agree with the approach.
It addresses the concerns beautifully, and I’m very grateful for the effort put into this!

middleware/client_ip.go
```diff
@@ -0,0 +1,185 @@
...
+			for i := len(xff) - 1; i >= 0; i-- {
+				ip, err := netip.ParseAddr(strings.TrimSpace(xff[i]))
+				if err != nil {
+					continue
+				}
+
+				// Ignore loopback, private, or unspecified addresses.
+				if ip.IsLoopback() || ip.IsPrivate() || ip.IsUnspecified() {
+					continue
+				}
```
c2h5oh, line 112: If trusted prefixes are set then we only care and/or can trust the first value past those trusted prefixes. No loopback, private pr unspecified addresses should exist past trusted prefixes unless headers have been tampered with.

1. Move trusted prefixes check block before this
2. Loopback/private/unspecified check should exit without setting client IP if any trusted prefixes are set

VojtechVitek:
> P.S.: This PR uses the "net/netip" package and assumes that we'll accept https://github.com/go-chi/chi/issues/963.

Rebased against master. The CI tests are now passing since we dropped support for Go 1.19 and older and thus can we use `"net/netip"` pkg.

@jawnsy @adam-p @convto @Dirbaio @pkieltyka @mfridman @lrstanley @n33pm @c2h5oh

I'd appreciate any more reviews 👀 🙏 .

middleware/client_ip.go
```diff
@@ -0,0 +1,185 @@
...
+// Example trusted IP ranges:
+// - "203.0.113.0/24"     - Example corporate proxy
+// - "198.51.100.0/24"    - Example data center or hosting provider
+// - "2400:cb00::/32"     - Cloudflare IPv6 range
+// - "2606:4700::/32"     - Cloudflare IPv6 range
+// - "192.0.2.0/24"       - Example VPN gateway
+//
+// Note: Private IP ranges (e.g., "10.0.0.0/8", "192.168.0.0/16", "172.16.0.0/12")
+// are automatically excluded by netip.Addr.IsPrivate() and do not need to be added here.
+func ClientIPFromXFFHeader(trustedIPPrefixes ...string) func(http.Handler) http.Handler {
```
c2h5oh, line 83: Having `numTrustedProxies` as an alternative to trusted prefixes would be a useful option too. If you control exactly n proxies, but not necessarily all possible proxies on a network (e.g. corporate intranet situation).

When set we should look only at nth rightmost value and only check if it's unspecified or loopback to reject - private is fine.

c2h5oh (changes requested):

middleware/client_ip.go
adam-p: I don't think the loopback/private/unspecified checks should be here at all. The user has specified that they trust certain reverse proxies, and I don't think we should have code here that second-guesses that. I think the first non-trusted IP should be returned regardless of what it is.

I would even prefer that there not be a check for IP validity at all (i.e., the netip.ParseAddr) call above, but it's obviously necessary for prefix checking. If the parse fails, I think it should trigger an error condition and not just be skipped.

These kinds of bad/incorrect IP failures indicate a configuration error. Which is pretty bad!

I'm even not sure I like "don't store the IP" as the way to handle error states. Panicking by default is unpalatable (although that's how I prefer to handle developer/configuration errors). Maybe there can be a flag or callback or something? I don't feel super strongly about it, though.

middleware/client_ip.go
```diff
@@ -0,0 +1,185 @@
...
+
+			// Check if the client IP is already set in the context.
+			if _, ok := ctx.Value(clientIPCtxKey).(netip.Addr); ok {
+				h.ServeHTTP(w, r)
+				return
+			}
+
+			// Parse the IP address from the trusted header.
+			ip, err := netip.ParseAddr(r.Header.Get(trustedHeader))
+			if err != nil || ip.IsLoopback() || ip.IsUnspecified() || ip.IsPrivate() {
```
adam-p, line 46: For the same reasons as I mention in reply to @c2h5oh's comment below, I don't think these checks should be here at all. (In short: The user has said this is trustworthy and we should treat it as such.)

middleware/client_ip.go
```diff
@@ -0,0 +1,185 @@
...
+		}
+		return http.HandlerFunc(fn)
+	}
+}
+
+// ClientIPFromRemoteAddr extracts the client IP address from the RemoteAddr
+// field of the HTTP request and injects it into the request context if it is
+// not already set. The parsed IP address can be retrieved using GetClientIP().
+//
+// The middleware ignores invalid or private IPs.
```
adam-p, line 137: As above: Why? The user has indicated that this is the value they want.

But also... I don't think you're actually doing the loopback/private/unspecified check here, so I don't think this comment is true?

middleware/client_ip.go
```diff
@@ -0,0 +1,185 @@
...
+// GetClientIP retrieves the client IP address from the given context.
+// The IP address is set by one of the following middlewares:
+// - ClientIPFromHeader
+// - ClientIPFromXFFHeader
+// - ClientIPFromRemoteAddr
+//
+// Returns an empty string if no valid IP is found.
+func GetClientIP(ctx context.Context) string {
+	ip, ok := ctx.Value(clientIPCtxKey).(netip.Addr)
+	if !ok || !ip.IsValid() {
```
adam-p, line 181: Depending on above discussions, the ip.IsValid check might not be right.

middleware/client_ip_test.go
```diff
@@ -0,0 +1,141 @@
...
+
+			if clientIP != tc.out {
+				t.Errorf("expected %v, got %v", tc.out, clientIP)
+			}
+		})
+	}
+}
+
+func TestClientIPFromXFFHeader(t *testing.T) {
+	tt := []struct {
```
adam-p, line 73: I would prefer to see tests that exercise the trusted range functionality.

adam-p (changes requested):

middleware/client_ip.go
c2h5oh: You will get private IPs here e.g. if you use nginx ingress in k8s

adam-p: Shouldn't it be on the user to know that and put the ingress IP prefixes in `trustedIPPrefixes`?

rezmoss:
test cases ip spoofing (feel free to edit/delete)
```
// TestClientIPFromXFFHeaderWithTrustedPrefixes tests the trusted CIDR prefix
func TestClientIPFromXFFHeaderWithTrustedPrefixes(t *testing.T) {
	tt := []struct {
		name            string
		xff             []string
		trustedPrefixes []string
		expected        string
	}{
		// single trusted proxy, proxy ip should be skipped
		{
			name:            "single_trusted_proxy",
			xff:             []string{"8.8.8.8, 203.0.113.50"},
			trustedPrefixes: []string{"203.0.113.0/24"},
			expected:        "8.8.8.8",
		},
		// multiple trusted proxies chain
		{
			name:            "multiple_trusted_proxies",
			xff:             []string{"1.1.1.1, 203.0.113.10, 198.51.100.5"},
			trustedPrefixes: []string{"203.0.113.0/24", "198.51.100.0/24"},
			expected:        "1.1.1.1",
		},
		// all ipss in trusted range,should return empty (no untrusted client)
		{
			name:            "all_ips_trusted",
			xff:             []string{"203.0.113.10, 203.0.113.20"},
			trustedPrefixes: []string{"203.0.113.0/24"},
			expected:        "",
		},
		// ipv6 trusted range e.g. cloudflare
		{
			name:            "ipv6_trusted_range",
			xff:             []string{"2001:db8::1, 2606:4700::1"},
			trustedPrefixes: []string{"2606:4700::/32"},
			expected:        "2001:db8::1",
		},
		// mixed ip and ipv6 trusted ranges
		{
			name:            "mixed_ipv4_ipv6_trusted",
			xff:             []string{"8.8.8.8, 2606:4700::1, 203.0.113.5"},
			trustedPrefixes: []string{"2606:4700::/32", "203.0.113.0/24"},
			expected:        "8.8.8.8",
		},
		// trusted prefix with longer chain
		{
			name:            "long_proxy_chain",
			xff:             []string{"1.2.3.4, 5.6.7.8, 203.0.113.1, 198.51.100.1"},
			trustedPrefixes: []string{"203.0.113.0/24", "198.51.100.0/24"},
			expected:        "5.6.7.8",
		},
	}

	for _, tc := range tt {
		t.Run(tc.name, func(t *testing.T) {
			req, _ := http.NewRequest("GET", "/", nil)
			for _, v := range tc.xff {
				req.Header.Add("X-Forwarded-For", v)
			}
			w := httptest.NewRecorder()

			r := chi.NewRouter()
			r.Use(ClientIPFromXFFHeader(tc.trustedPrefixes...))

			var clientIP string
			r.Get("/", func(w http.ResponseWriter, r *http.Request) {
				clientIP = GetClientIP(r.Context())
				w.Write([]byte("OK"))
			})
			r.ServeHTTP(w, req)

			if clientIP != tc.expected {
				t.Errorf("expected %q, got %q", tc.expected, clientIP)
			}
		})
	}
}

// TestXFFSpoofingPrevention tests ip spoofing attack scenarios in X-Forwarded-For context
func TestXFFSpoofingPrevention(t *testing.T) {
	tt := []struct {
		name            string
		xff             []string
		trustedPrefixes []string
		expected        string
		description     string
	}{
		{
			name:        "prepend_spoofed_loopback",
			xff:         []string{"127.0.0.1, 8.8.8.8"},
			expected:    "8.8.8.8",
			description: "Spoofed loopback ignored, real IP returned",
		},
		{
			name:        "prepend_spoofed_private",
			xff:         []string{"10.0.0.1, 8.8.8.8"},
			expected:    "8.8.8.8",
			description: "Spoofed private IP ignored, real IP returned",
		},
		// attacker prepends spoofed ip w trusted proxy in chain
		{
			name:            "spoof_with_trusted_proxy",
			xff:             []string{"127.0.0.1, 203.0.113.50, 198.51.100.1"},
			trustedPrefixes: []string{"198.51.100.0/24"},
			expected:        "203.0.113.50",
			description:     "Proxy skipped, spoofed loopback ignored, client IP returned",
		},
		// multiple xff headers,attacker injects first header
		{
			name:            "multi_header_injection_with_proxy",
			xff:             []string{"127.0.0.1", "8.8.8.8, 198.51.100.1"},
			trustedPrefixes: []string{"198.51.100.0/24"},
			expected:        "8.8.8.8",
			description:     "Attacker's header ignored, proxy chain processed correctly",
		},
		{
			name:        "multi_header_injection_loopback",
			xff:         []string{"127.0.0.1", "8.8.8.8"},
			expected:    "8.8.8.8",
			description: "Attacker injects loopback in separate header",
		},
		// rate limit bypass prevention,spoofed ipsshould be ignored
		{
			name:            "rate_limit_bypass_attempt",
			xff:             []string{"attacker.spoofed.ip, 8.8.8.8, 198.51.100.1"},
			trustedPrefixes: []string{"198.51.100.0/24"},
			expected:        "8.8.8.8",
			description:     "Invalid spoofed IP ignored, consistent client IP for rate limiting",
		},
	}

	for _, tc := range tt {
		t.Run(tc.name, func(t *testing.T) {
			req, _ := http.NewRequest("GET", "/", nil)
			for _, v := range tc.xff {
				req.Header.Add("X-Forwarded-For", v)
			}
			w := httptest.NewRecorder()

			r := chi.NewRouter()
			r.Use(ClientIPFromXFFHeader(tc.trustedPrefixes...))

			var clientIP string
			r.Get("/", func(w http.ResponseWriter, r *http.Request) {
				clientIP = GetClientIP(r.Context())
				w.Write([]byte("OK"))
			})
			r.ServeHTTP(w, req)

			if clientIP != tc.expected {
				t.Errorf("%s: expected %q, got %q", tc.description, tc.expected, clientIP)
			}
		})
	}
}

// TestClientIPMiddlewareChaining verifies that the first middleware to set
func TestClientIPMiddlewareChaining(t *testing.T) {
	t.Run("header_before_xff", func(t *testing.T) {
		req, _ := http.NewRequest("GET", "/", nil)
		req.Header.Set("CF-Connecting-IP", "1.1.1.1")
		req.Header.Set("X-Forwarded-For", "2.2.2.2")
		w := httptest.NewRecorder()

		r := chi.NewRouter()
		r.Use(ClientIPFromHeader("CF-Connecting-IP"))
		r.Use(ClientIPFromXFFHeader())

		var clientIP string
		r.Get("/", func(w http.ResponseWriter, r *http.Request) {
			clientIP = GetClientIP(r.Context())
			w.Write([]byte("OK"))
		})
		r.ServeHTTP(w, req)

		expected := "1.1.1.1"
		if clientIP != expected {
			t.Errorf("expected %q (from header), got %q", expected, clientIP)
		}
	})

	t.Run("xff_before_remoteaddr", func(t *testing.T) {
		req, _ := http.NewRequest("GET", "/", nil)
		req.Header.Set("X-Forwarded-For", "8.8.8.8")
		req.RemoteAddr = "192.0.2.1:1234"
		w := httptest.NewRecorder()

		r := chi.NewRouter()
		r.Use(ClientIPFromXFFHeader())
		r.Use(ClientIPFromRemoteAddr)

		var clientIP string
		r.Get("/", func(w http.ResponseWriter, r *http.Request) {
			clientIP = GetClientIP(r.Context())
			w.Write([]byte("OK"))
		})
		r.ServeHTTP(w, req)

		expected := "8.8.8.8"
		if clientIP != expected {
			t.Errorf("expected %q (from XFF), got %q", expected, clientIP)
		}
	})

	t.Run("empty_header_falls_through", func(t *testing.T) {
		req, _ := http.NewRequest("GET", "/", nil)
		req.Header.Set("X-Forwarded-For", "8.8.8.8")
		w := httptest.NewRecorder()

		r := chi.NewRouter()
		r.Use(ClientIPFromHeader("CF-Connecting-IP"))
		r.Use(ClientIPFromXFFHeader())

		var clientIP string
		r.Get("/", func(w http.ResponseWriter, r *http.Request) {
			clientIP = GetClientIP(r.Context())
			w.Write([]byte("OK"))
		})
		r.ServeHTTP(w, req)

		expected := "8.8.8.8"
		if clientIP != expected {
			t.Errorf("expected %q (from XFF fallback), got %q", expected, clientIP)
		}
	})

	t.Run("private_header_falls_through", func(t *testing.T) {
		req, _ := http.NewRequest("GET", "/", nil)
		req.Header.Set("X-Real-IP", "192.168.1.1") // private ignoredd
		req.Header.Set("X-Forwarded-For", "8.8.8.8")
		w := httptest.NewRecorder()

		r := chi.NewRouter()
		r.Use(ClientIPFromHeader("X-Real-IP"))
		r.Use(ClientIPFromXFFHeader())

		var clientIP string
		r.Get("/", func(w http.ResponseWriter, r *http.Request) {
			clientIP = GetClientIP(r.Context())
			w.Write([]byte("OK"))
		})
		r.ServeHTTP(w, req)

		expected := "8.8.8.8"
		if clientIP != expected {
			t.Errorf("expected %q (XFF after private rejected), got %q", expected, clientIP)
		}
	})
}

// TestPrivateIPWithTrustedPrefixes Should IsPrivate()/IsLoopback() checks be dropped when trusted prefixes are configured based  adam-p blog
func TestPrivateIPWithTrustedPrefixes(t *testing.T) {
	tt := []struct {
		name            string
		xff             string
		trustedPrefixes []string
		currentResult   string
		proposedResult  string
		description     string
	}{
		// k8s pod to pod: both ip are private
		{
			name:            "k8s_pod_to_pod",
			xff:             "10.244.1.50, 10.244.0.10",
			trustedPrefixes: []string{"10.244.0.0/24"},
			currentResult:   "",            // both private both skipped
			proposedResult:  "10.244.1.50", // only trusted prefix filters
			description:     "Pod client through service mesh sidecar",
		},
		// nginx ingress with private proxy ip
		{
			name:            "nginx_private_proxy",
			xff:             "203.0.113.50, 192.168.1.10",
			trustedPrefixes: []string{"192.168.1.0/24"},
			currentResult:   "203.0.113.50", // woks public client
			proposedResult:  "203.0.113.50", // dame
			description:     "Public client through private nginx",
		},
		// aws internal load balancer
		{
			name:            "aws_internal_lb",
			xff:             "8.8.8.8, 10.0.0.50",
			trustedPrefixes: []string{"10.0.0.0/16"},
			currentResult:   "8.8.8.8", // works LB skipped via trusted prefix
			proposedResult:  "8.8.8.8", // same
			description:     "Public client through internal LB",
		},
	}

	for _, tc := range tt {
		t.Run(tc.name, func(t *testing.T) {
			req, _ := http.NewRequest("GET", "/", nil)
			req.Header.Set("X-Forwarded-For", tc.xff)
			w := httptest.NewRecorder()

			r := chi.NewRouter()
			r.Use(ClientIPFromXFFHeader(tc.trustedPrefixes...))

			var clientIP string
			r.Get("/", func(w http.ResponseWriter, r *http.Request) {
				clientIP = GetClientIP(r.Context())
				w.Write([]byte("OK"))
			})
			r.ServeHTTP(w, req)

			t.Logf("Scenario: %s", tc.description)
			t.Logf("XFF: %s, Trusted: %v", tc.xff, tc.trustedPrefixes)
			t.Logf("Result: %q (current expected: %q, proposed: %q)",
				clientIP, tc.currentResult, tc.proposedResult)

			if clientIP != tc.currentResult && clientIP != tc.proposedResult {
				t.Errorf("unexpected: got %q", clientIP)
			}
			if clientIP == tc.proposedResult && tc.proposedResult != tc.currentResult {
				t.Logf("now matches")
			}
		})
	}
}

```

VojtechVitek:
Thanks everyone for the thorough reviews and feedback on this PR. I've been figuring out how to work through it all — the comments, open advisories, and decisions needed. With a bit of help from Opus 4.7, I've worked through everything and landed on a plan. Give it a read and let me know if anything stands out.
I'll execute on the plan and have a final commit ready to merge.

---

# PR #967 — ClientIP middleware: Plan to finish

A plan to address the open review feedback on
[PR #967](https://github.com/go-chi/chi/pull/967), close out the three
published security advisories against `middleware.RealIP`, and land a
small, safe, easy-to-use `ClientIP*` middleware family.

---

## Why this matters — three open advisories

Chi currently has **three published security advisories**, all
targeting the same root cause: `middleware.RealIP` blindly trusts
client-supplied headers.

| Advisory | Reporter | Severity | Status |
|---|---|---|---|
| [GHSA-9g5q-2w5x-hmxf](https://github.com/go-chi/chi/security/advisories/GHSA-9g5q-2w5x-hmxf) — IP spoofing via XFF in `RemoteAddr` resolution | convto | Moderate | Unpatched |
| [GHSA-rjr7-jggh-pgcp](https://github.com/go-chi/chi/security/advisories/GHSA-rjr7-jggh-pgcp) — RealIP allows IP spoofing via unvalidated XFF | rezmoss | Moderate | Unpatched |
| [GHSA-3fxj-6jh8-hvhx](https://github.com/go-chi/chi/security/advisories/GHSA-3fxj-6jh8-hvhx) — IP spoofing in `middleware.RealIP` | Saku0512 | **Critical (9.3)** | Unpatched |

All three reporters propose the same fix shape: rightmost-untrusted
traversal of `X-Forwarded-For` with an explicit trusted-CIDR list.
That is exactly what `ClientIPFromXFF` does (renamed from
`ClientIPFromXFFHeader` in this PR).

**This PR is the patch for all three advisories** and the
deprecation of `middleware.RealIP` is part of that patch.

---

## What the source-of-truth references say

[adam-p's "The perils of the 'real' client IP"](https://adam-p.ca/blog/2022/03/x-forwarded-for/)
is the definitive write-up. The points that drive our API:

1. **Rightmost-ish, never leftmost.** Leftmost is trivially spoofable.
2. **There's no sane default — so there should be no default.** The
   blog calls out chi by name on this. ("Conclusion #4")
3. **Defaults must not be insecure.** No auto-fall-back through a
   list of headers an attacker could supply. ("Conclusion #5")
4. **Any header you trust must be set/overwritten by your own
   infrastructure for every request** — otherwise it's
   attacker-controlled.
5. **Multiple `X-Forwarded-For` headers must be merged**, not just
   `Header.Get()`-ed.
6. **Behind multiple proxies, you need either trusted-proxy
   IPs/prefixes OR a trusted-proxy count** (rightmost-ish algorithm).
7. **`RemoteAddr` can be `ip:port`** — split correctly.
8. **Re-fronting attack** — even a "trusted" CDN IP can be
   attacker-controlled if a third-party distribution points at your
   origin. Mitigation is out-of-band (shared secret / client cert at
   the origin). Out of scope to fix in middleware; doc warning only.
9. **`Forwarded` (RFC 7239) has the same problems** and almost
   nobody uses it. Skip.

---

## Summary of reviewer feedback on PR #967

### c2h5oh (CHANGES_REQUESTED)

1. **`client_ip.go` L112** — when `trustedIPPrefixes` is set, the
   `IsLoopback / IsPrivate / IsUnspecified` rejection inside the XFF
   loop is wrong: "no loopback, private or unspecified addresses
   should exist past/in between trusted prefixes unless headers have
   been tampered with."
2. (Jun 2025) **Real-world hit**: nginx-ingress in k8s legitimately
   produces private IPs in the chain — the current code skips them
   and breaks lookups.
3. **New API idea**: a trusted-proxy-count knob (Traefik-style) for
   the case where you control exactly N proxies but not all possible
   proxies on a corporate network.

### adam-p (CHANGES_REQUESTED)

1. **L32 godoc** — `X-Azure-ClientIP`, `True-Client-IP`,
   `Fastly-Client-IP` are NOT trustworthy headers per his blog and
   shouldn't be listed as such.
2. **L46 and L112** — drop all the `IsLoopback / IsPrivate /
   IsUnspecified` second-guessing in both `ClientIPFromHeader` and
   `ClientIPFromXFFHeader`. The user has explicitly said what to
   trust; we shouldn't override that.
3. **L112** — a `ParseAddr` failure mid-chain indicates a
   configuration error; it shouldn't be silently skipped. Suggests
   an error / flag / callback (no strong preference).
4. **L137 godoc** (`ClientIPFromRemoteAddr`) — comment claims it
   "ignores invalid or private IPs", but it doesn't.
5. **L181 `GetClientIP`** — `ip.IsValid()` check may be redundant if
   we drop second-guessing.
6. **L73 tests** — add tests that exercise the trusted-range
   functionality.
7. (Jun 2025) Reaffirms: even nginx-ingress private IPs should be
   configured by the user via `trustedIPPrefixes`, not auto-filtered.

### rezmoss (proposed tests on PR #967)

Four ready-to-paste test functions:

- `TestClientIPFromXFFHeaderWithTrustedPrefixes` — CIDR behavior incl.
  IPv6, mixed v4/v6, long chains.
- `TestXFFSpoofingPrevention` — rightmost-untrusted defeats spoofing.
- `TestClientIPMiddlewareChaining` — header → xff → remoteaddr
  fallthrough.
- `TestPrivateIPWithTrustedPrefixes` — k8s, nginx, AWS internal LB.

### Saku0512 (follow-up [PR #1087](https://github.com/go-chi/chi/pull/1087) on top of #967)

Adds `TestClientIPFromXFFHeaderWithTrustedPrefixes` with boundary
tests: `.0` and `.255` of a `/24`, IP just outside, all-trusted →
empty, no-prefixes baseline. **Cleanly aligned with our API.** We
pull these in (or merge #1087 once we've landed the core changes).

### Convergence

Both blocking reviewers and all three advisory reporters
(rezmoss explicitly, convto and Saku0512 by algorithm) converge on
one thing: **stop second-guessing the user**. Drop the
`IsPrivate / IsLoopback / IsUnspecified` rejection and let the
explicit trust list be authoritative. This also unblocks the
legitimate k8s / nginx-ingress / internal-LB cases.

---

## Is the API right?

Final surface for this PR:

```go
// Trust a single-IP header set by your infrastructure.
func ClientIPFromHeader(trustedHeader string) func(http.Handler) http.Handler

// Trust XFF, walking right-to-left and skipping the listed trusted CIDRs.
// Zero args ⇒ rightmost parseable IP (safe only behind one trusted hop).
func ClientIPFromXFF(trustedIPPrefixes ...string) func(http.Handler) http.Handler

// Trust XFF, knowing the exact number of trusted proxies in front of you.
// Returns xff[len-numTrustedProxies], or empty if XFF has fewer entries.
// Panics if numTrustedProxies < 1.
func ClientIPFromXFFTrustedProxies(numTrustedProxies int) func(http.Handler) http.Handler

// Trust the TCP RemoteAddr (only safe when directly connected to the internet).
func ClientIPFromRemoteAddr(h http.Handler) http.Handler

// Accessors.
func GetClientIP(ctx context.Context) string         // empty if not set
func GetClientIPAddr(ctx context.Context) netip.Addr // zero value if not set
```

Scored against the blog + advisory checklist:

| Requirement | Status |
|---|---|
| Rightmost-untrusted algorithm | ✅ `ClientIPFromXFF` |
| No default header fallback list | ✅ separate middlewares; user composes |
| Merge multiple XFF headers | ✅ `Header.Values("X-Forwarded-For")` joined |
| Trusted CIDRs / IPs | ✅ variadic `trustedIPPrefixes ...string` |
| Trusted-proxy-COUNT alternative (Traefik-style) | ✅ `ClientIPFromXFFTrustedProxies(n)` |
| Validate before parsing | ✅ via `netip.ParseAddr` |
| Don't mutate `RemoteAddr` | ✅ context-based |
| Doesn't second-guess what the user trusts | 🟡 still rejects private/loopback — dropped in Phase 1 |
| Typed accessor for `netip.Addr` | ✅ `GetClientIPAddr(ctx)` |
| Strong security warnings in godoc | 🟡 insufficient — expanded in Phase 1 |
| Misleading "trusted" header examples | 🟡 lists Akamai/Azure/Fastly — removed in Phase 1 |
| Document re-fronting / dual-connection / network-change risks | ❌ added in Phase 1 |

**Why two XFF functions and not one?**

`ClientIPFromXFF(prefixes...)` and
`ClientIPFromXFFTrustedProxies(n)` configure trust in fundamentally
different ways:

| | What you specify | Best when |
|---|---|---|
| `ClientIPFromXFF(prefixes...)` | **IP ranges** of your proxies | You know the IPs (static fleet, on-prem, Cloudflare's published list) |
| `ClientIPFromXFFTrustedProxies(n)` | **Number** of proxies | You don't know the IPs but you know the hop count (autoscaling, ephemeral pools) |

They overlap intentionally for the trivial case: `(no prefixes)` and
`TrustedProxies(1)` both return the rightmost parseable IP. That's
fine — the user picks based on what they actually know about their
infra.

---

## Proposed plan

### Phase 1 — Code changes to `middleware/client_ip.go`

1. **`ClientIPFromHeader`** (adam-p L46, advisories)
   Drop the `IsLoopback || IsUnspecified || IsPrivate` rejection.
   Parse-fail still rejects. The user has chosen the header
   explicitly; if their infra writes a private IP there, that's the
   answer.

2. **Rename `ClientIPFromXFFHeader` → `ClientIPFromXFF`**
   Shorter, pairs nicely with `ClientIPFromRemoteAddr`. Drop the
   `IsLoopback || IsPrivate || IsUnspecified` rejection inside the
   loop (c2h5oh L112, adam-p L112, all three advisories). Walk
   right-to-left, skip values matching `trustedIPPrefixes`, return
   the first remaining IP. Pure rightmost-untrusted, no
   second-guessing.

   Keep silent-skip on `ParseAddr` failure mid-chain (chi has no
   logger here; buggy proxies do emit garbage; revisit in a
   follow-up if anyone asks).

3. **NEW: `ClientIPFromXFFTrustedProxies`** (c2h5oh's ask)
   Traefik-style trusted-proxy count variant. Algorithm: merge all
   XFF headers into a single list, then return
   `xff[len(xff) - numTrustedProxies]` if (a) that index exists and
   (b) the value parses as an IP. Otherwise don't set a client IP
   (downstream middleware like `ClientIPFromRemoteAddr` can be
   chained as a fallback). Panic at construction if
   `numTrustedProxies < 1`.

4. **`ClientIPFromRemoteAddr`** (adam-p L137)
   Fix the godoc — it currently claims "ignores invalid or private
   IPs" but doesn't filter private. Remove the false claim.

5. **Godoc cleanup for `ClientIPFromHeader`** (adam-p L32)
   Drop `X-Azure-ClientIP`, `True-Client-IP`, `Fastly-Client-IP`
   from the examples (they're not reliably infra-overwritten per
   the blog). Keep `X-Real-IP`, `X-Client-IP`, `CF-Connecting-IP`.
   Full text in Phase 1.5.

6. **NEW: `GetClientIPAddr`, refactor `GetClientIP`** (adam-p L181,
   user request)
   Add a typed accessor returning `netip.Addr`; `GetClientIP`
   becomes a thin wrapper around it. Keep the `IsValid()` check.
   Full text in Phase 1.5.

7. **NEW: package-level security guidance**
   Add a doc block at the top of `client_ip.go`. Full text in
   Phase 1.5 ("Choosing a ClientIPFrom* middleware"). The block
   also expands into per-function godocs that cover:
   - HTTPS required (plaintext can be modified in transit).
   - Firewall must restrict ingress to your trusted edge.
   - Re-fronting attack (third-party CDN distribution → your origin).
   - Network-architecture changes require re-checking config.

### Phase 1.5 — Godoc spec (exact text to drop into the code)

#### Package-level decision guide

Placed above the first `ClientIPFrom*` function in `client_ip.go`:

```go
// # Choosing a ClientIPFrom* middleware
//
// There is no safe default. Apply exactly ONE ClientIPFrom* middleware
// per request, based on your network setup:
//
//   - [ClientIPFromRemoteAddr] — this server is directly on the public
//     internet, with no reverse proxy in front.
//
//   - [ClientIPFromHeader] — your reverse proxy sets a dedicated
//     single-IP header (X-Real-IP, CF-Connecting-IP, X-Client-IP)
//     and overwrites any client-supplied value for every request.
//
//   - [ClientIPFromXFF] — you sit behind one or more reverse proxies
//     whose IP ranges you can enumerate (your VPC CIDR, Cloudflare's
//     published IP list, etc.).
//
//   - [ClientIPFromXFFTrustedProxies] — you sit behind a known, fixed
//     number of reverse proxies whose IPs are dynamic (autoscaling
//     pools, ephemeral containers).
//
// Read the resulting IP with [GetClientIP] (string) or
// [GetClientIPAddr] (netip.Addr).
```

#### `ClientIPFromHeader`

```go
// ClientIPFromHeader stores the client IP read from a single-IP HTTP
// header set by your reverse proxy. Read the IP with [GetClientIP].
//
// Use this when your reverse proxy sets one of these headers for every
// request and overwrites any client-supplied value:
//
//   - X-Real-IP        — Nginx with ngx_http_realip_module
//   - X-Client-IP      — Apache with mod_remoteip
//   - CF-Connecting-IP — Cloudflare
//
// DO NOT use this with headers your infrastructure does not overwrite
// (True-Client-IP, X-Azure-ClientIP, Fastly-Client-IP by default,
// etc.). Those can be supplied by the client and are trivially
// spoofable. See https://adam-p.ca/blog/2022/03/x-forwarded-for/ for
// details.
//
// Pick exactly one ClientIPFrom* middleware for your application; see
// the "Choosing a ClientIPFrom* middleware" section above.
```

#### `ClientIPFromXFF`

```go
// ClientIPFromXFF stores the client IP read from the X-Forwarded-For
// header, walking the chain right-to-left and skipping any IP that
// falls within one of the given trusted CIDR prefixes. The first IP
// that is not trusted is the client. Read it with [GetClientIP].
//
// Use this when you sit behind one or more reverse proxies whose IP
// ranges you can enumerate as CIDRs:
//
//   r.Use(middleware.ClientIPFromXFF(
//       "13.32.0.0/15",   // CloudFront IPv4
//       "52.46.0.0/18",   // CloudFront IPv4
//       "2600:9000::/28", // CloudFront IPv6
//   ))
//
// Calling with no arguments returns the rightmost parseable XFF IP —
// safe only if you have exactly one trusted hop directly in front of
// this server (e.g., nginx on localhost).
//
// If you know the number of trusted proxies but not their IPs, use
// [ClientIPFromXFFTrustedProxies] instead.
//
// Panics at startup if any prefix is invalid. Pick exactly one
// ClientIPFrom* middleware for your application.
```

#### `ClientIPFromXFFTrustedProxies`

```go
// ClientIPFromXFFTrustedProxies stores the client IP read from the
// X-Forwarded-For header, given the exact number of trusted reverse
// proxies between this server and the public internet. It returns the
// IP at position len(xff) - numTrustedProxies in the merged
// X-Forwarded-For list — the IP added by the outermost of your
// trusted proxies, the only IP in the chain that none of your proxies
// have allowed an attacker to forge. Read it with [GetClientIP].
//
// Use this when:
//   - You know exactly how many proxies you sit behind, AND
//   - Their IP addresses are dynamic (autoscaling proxy pools,
//     ephemeral containers, dynamic CDN edges) so listing CIDRs
//     with [ClientIPFromXFF] is impractical.
//
// WARNING: This variant is brittle to network architecture changes.
// If you add or remove a proxy level, numTrustedProxies silently
// becomes wrong and you may start trusting an attacker-supplied IP.
// Prefer [ClientIPFromXFF] with explicit trusted CIDRs whenever you
// can.
//
// If the XFF chain has fewer than numTrustedProxies entries (header
// missing or architecture changed), no client IP is set; chain
// [ClientIPFromRemoteAddr] after this middleware if you want a
// fallback.
//
// Panics at startup if numTrustedProxies < 1. Pick exactly one
// ClientIPFrom* middleware for your application.
```

#### `ClientIPFromRemoteAddr`

```go
// ClientIPFromRemoteAddr stores the client IP read from the TCP
// RemoteAddr of the incoming request — the IP address of whoever
// opened the connection to this server. Read it with [GetClientIP].
//
// Use this when this server is directly connected to the public
// internet with NO reverse proxy in front of it. Behind a reverse
// proxy, RemoteAddr is the proxy's IP, not the client's — use
// [ClientIPFromHeader] or [ClientIPFromXFF] instead.
//
// Pick exactly one ClientIPFrom* middleware for your application.
```

#### `GetClientIP` and `GetClientIPAddr`

```go
// GetClientIP returns the client IP as a string, as set by one of the
// ClientIPFrom* middlewares. Returns "" if no valid IP was set.
// Convenient for logging, rate-limit keys, etc.
func GetClientIP(ctx context.Context) string

// GetClientIPAddr returns the client IP as a netip.Addr, as set by one
// of the ClientIPFrom* middlewares. The returned Addr is the zero
// value if not set; use .IsValid() to check. Useful when you need
// typed work — prefix containment, Is4/Is6, etc. — without re-parsing.
func GetClientIPAddr(ctx context.Context) netip.Addr
```

### Phase 2 — Update `middleware/client_ip_test.go`

8. **Fix existing tests for `ClientIPFromHeader`**
   `invalid_loopback`, `invalid_zeroes`, `invalid_private_*` cases
   become **passing** cases (we no longer filter). Rename them and
   set `out` equal to `in`.

9. **Fix existing tests for `ClientIPFromXFF`** (renamed from
   `ClientIPFromXFFHeader`)
   Most stay; adjust any that depended on filtering. Update test
   function names to match.

10. **Adopt new tests**
    - Pull in Saku0512's [`TestClientIPFromXFFHeaderWithTrustedPrefixes`](https://github.com/go-chi/chi/pull/1087)
      (already prefix-based, includes boundary tests). Rename to
      `TestClientIPFromXFFWithTrustedPrefixes`. Either merge #1087
      after #967 lands, or include the same content here.
    - Pull in rezmoss's `TestXFFSpoofingPrevention` ✓ — still passes
      under the new algorithm because rightmost-untrusted naturally
      defeats prepended spoofs.
    - Pull in rezmoss's `TestClientIPMiddlewareChaining` ✓ — fix the
      `private_header_falls_through` subtest: with our changes
      `X-Real-IP: 192.168.1.1` is no longer ignored; rework the
      scenario to show that the first middleware to set wins.
    - Pull in rezmoss's `TestPrivateIPWithTrustedPrefixes` ✓ —
      simplify to assert only the `proposedResult` (now the only
      behavior).

11. **NEW: tests for `ClientIPFromXFFTrustedProxies`**
    - `N=1` returns rightmost (equivalence with zero-arg
      `ClientIPFromXFF()`).
    - `N=2` skips 1 rightmost, returns 2nd from right.
    - `N=3` skips 2 rightmost, returns 3rd from right.
    - `len(xff) < N` → empty (no IP set, downstream fallback works).
    - `len(xff) == N` → leftmost.
    - Spoofing attempt: attacker prepends fake IPs → still picks the
      correct slot, ignoring everything to its left.
    - Bad parse at target slot → empty (no IP set).
    - Construction with `N < 1` panics.

12. **NEW: tests for `GetClientIPAddr`**
    - Returns zero `netip.Addr` when not set (`!IsValid()`).
    - Round-trips against `GetClientIP` for set values.

### Phase 3 — Deprecate `middleware.RealIP`

13. **`middleware/realip.go`** — add a `// Deprecated:` godoc on
    `RealIP`:

    > Deprecated: RealIP is vulnerable to IP spoofing — see
    > GHSA-3fxj-6jh8-hvhx, GHSA-rjr7-jggh-pgcp,
    > GHSA-9g5q-2w5x-hmxf. Use [ClientIPFromXFF],
    > [ClientIPFromHeader] or [ClientIPFromRemoteAddr] instead, and
    > read the value with [GetClientIP].

    Keep the function working for backward compatibility. Removal
    can be considered for a future major release.

14. **README** — short section pointing migrators from `RealIP` to
    `ClientIP*`, with copy-paste recipes for the common deployment
    shapes (direct, behind nginx, behind Cloudflare, behind AWS
    ALB).

### Phase 4 — Update PR description

15. Remove the `X-Azure-ClientIP` example from the PR body.
16. Add an example for the new `ClientIPFromXFFTrustedProxies`
    variant.
17. Note that this PR is the patch for the three advisories listed
    above and that `middleware.RealIP` is being deprecated.

### Phase 5 — Publish the advisories

18. Once #967 is merged and a release cut, mark all three GHSAs as
    patched in that release. Update CVE info if assigned.

---

## Open questions — resolved

- **Q1 (Traefik-style proxy count):** ✅ INCLUDED in this PR as
  `ClientIPFromXFFTrustedProxies(numTrustedProxies int)`. See
  Phase 1 step 3.

- **Q2 (zero-arg `ClientIPFromXFF()`):** ✅ Kept valid → returns
  rightmost parseable IP. Documented as "safe only when you have
  exactly one trusted hop directly in front of you." Overlaps with
  `TrustedProxies(1)` by design; user picks based on whether they
  know IPs or proxy count.

- **Q3 (`ParseAddr` failure mid-chain):** ✅ Defer to a follow-up.
  v1 silently skips. Future extension point if anyone asks:

  ```go
  type XFFOpts struct {
      OnParseError func(r *http.Request, value string, err error)
  }
  func ClientIPFromXFFWithOptions(opts XFFOpts, trustedIPPrefixes ...string) ...
  ```

- **Q4 (typed accessor `GetClientIPAddr`):** ✅ INCLUDED in this PR.
  Returns `netip.Addr` (not `net.Addr` — that's an endpoint interface
  including a port). Essentially free since we already store
  `netip.Addr` in the context. `GetClientIP` becomes a thin wrapper.

---

## Out of scope (mentioned for completeness)

- **`go-chi/httprate.LimitByRealIP`** — separate repo. It also
  needs to switch to the new model. Open a follow-up issue there
  referencing this PR.
- **RFC 7239 `Forwarded` header support** — almost nobody uses it,
  same problems as XFF. Not worth the API surface.
- **Auto-verifying the immediate `RemoteAddr` against a trusted
  edge** — adam-p mentions this as a defense for the
  "behind-proxy-and-internet" anti-pattern. Better solved at the
  firewall/security-group layer than in middleware. Doc warning.

---

## Summary — the ClientIP API

Four middlewares, two accessors. There is no safe default — the user
picks exactly one middleware based on their network setup, and reads
the resulting IP with one of two accessors:

```go
// One of the four. Pick based on your infrastructure.
func ClientIPFromHeader(trustedHeader string) func(http.Handler) http.Handler
func ClientIPFromXFF(trustedIPPrefixes ...string) func(http.Handler) http.Handler
func ClientIPFromXFFTrustedProxies(numTrustedProxies int) func(http.Handler) http.Handler
func ClientIPFromRemoteAddr(h http.Handler) http.Handler

// One of the two. Pick based on what you need.
func GetClientIP(ctx context.Context) string         // for logs, rate-limit keys
func GetClientIPAddr(ctx context.Context) netip.Addr // for typed work
```

| Your setup | Use |
|---|---|
| Directly on the public internet, no proxy | `ClientIPFromRemoteAddr` |
| Behind nginx (`X-Real-IP`), Cloudflare (`CF-Connecting-IP`), Apache (`X-Client-IP`) | `ClientIPFromHeader(...)` |
| Behind one or more proxies whose IP ranges you can list | `ClientIPFromXFF(cidr1, cidr2, …)` |
| Behind a fixed number of proxies with dynamic IPs | `ClientIPFromXFFTrustedProxies(n)` |

The implementation is rightmost-untrusted, never second-guesses what
the user trusts, doesn't mutate `r.RemoteAddr`, and stores
`netip.Addr` in context. It is the patch for GHSA-3fxj-6jh8-hvhx,
GHSA-rjr7-jggh-pgcp, and GHSA-9g5q-2w5x-hmxf, and supersedes the
now-deprecated `middleware.RealIP`.

VojtechVitek:
@c2h5oh @adam-p @Saku0512 please review :)

middleware/client_ip.go
```diff
@@ -0,0 +1,200 @@
...
+//     is impractical.
+//
+// WARNING: This variant is brittle to network architecture changes. If you
+// add or remove a proxy level, numTrustedProxies silently becomes wrong and
+// you may start trusting an attacker-supplied IP. Prefer [ClientIPFromXFF]
+// with explicit trusted CIDRs whenever you can.
+//
+// If the XFF chain has fewer than numTrustedProxies entries (header missing
+// or architecture changed), no client IP is set; chain [ClientIPFromRemoteAddr]
+// after this middleware if you want a fallback.
```
Saku0512: Should this say to register `ClientIPFromRemoteAddr` **before** this middleware?

In chi, middlewares run in registration order, and each `ClientIPFrom*` middleware overwrites the context value when it finds a valid IP. If users follow the current "after this middleware" guidance, `ClientIPFromRemoteAddr` runs later and overwrites a valid XFF-derived client IP with the immediate peer address.

For fallback behavior, the safe order should be fallback first, preferred source last:

```go
r.Use(middleware.ClientIPFromRemoteAddr)
r.Use(middleware.ClientIPFromXFFTrustedProxies(2))
```

That way RemoteAddr is used only when XFF does not produce a client IP.

Saku0512 (changes requested):
Requesting changes for one documentation issue that can lead to incorrect client IP selection in real deployments. The implementation looks good, but the fallback ordering guidance for `ClientIPFromXFFTrustedProxies` appears reversed for chi's middleware execution order.

middleware/client_ip.go
VojtechVitek: Good catch on the inverted order — but rather than reverse the example, I've dropped the fallback-chaining note entirely. The whole point of 482e855 was to push users to pick exactly one ClientIPFrom* middleware; documenting a chaining recipe here works against that. The honest answer when XFF is too short is just GetClientIP() == "", and the caller decides what to do.

Thank you 👍

Saku0512 (approved):

middleware/client_ip.go
VojtechVitek: Addressed in 7ca76eecd95a36161f7ac48e5f3c14e46197d5b5.

Appreciate your help 🙏 🙌

Saku0512: thanks!!

VojtechVitek:
## Threat model coverage

| Advisory | Root cause in `RealIP` | How this PR closes it |
|---|---|---|
| [GHSA-3fxj-6jh8-hvhx](https://github.com/go-chi/chi/security/advisories/GHSA-3fxj-6jh8-hvhx) | Trusts arbitrary client-supplied headers | User explicitly picks one source; rightmost-untrusted on XFF; bare TCP peer for direct-internet |
| [GHSA-rjr7-jggh-pgcp](https://github.com/go-chi/chi/security/advisories/GHSA-rjr7-jggh-pgcp) | Auto-fallback through True-Client-IP / X-Real-IP / XFF | No default; each `ClientIPFrom*` reads exactly one opted-in source |
| [GHSA-9g5q-2w5x-hmxf](https://github.com/go-chi/chi/security/advisories/GHSA-9g5q-2w5x-hmxf) | Mutates `r.RemoteAddr` from leftmost XFF | Context-stored; rightmost-untrusted; `r.RemoteAddr` never mutated |

## adam-p blog checklist

Against [adam-p's "The perils of the 'real' client IP"](https://adam-p.ca/blog/2022/03/x-forwarded-for/):

| Requirement | Status |
|---|---|
| Rightmost-ish, never leftmost | ✅ |
| No safe default; no auto-fallback header list | ✅ |
| Trusted headers must be infra-overwritten | ✅ documented, traps named by product |
| Merge multiple XFF headers (RFC 2616) | ✅ |
| Single-IP header tolerates multi-value misconfig (last wins) | ✅ defense-in-depth |
| Trusted CIDRs *or* trusted-proxy count | ✅ both variants |
| `RemoteAddr` should be `ip:port` | ✅ |
| Don't mutate `r.RemoteAddr` | ✅ context-based |
| No private/loopback filtering (don't second-guess the user) | ✅ |
| Validate before parsing; fail-closed on garbage mid-chain | ✅ |
| Normalize v4-mapped IPv6 and IPv6 header zones | ✅ at every entry point |
| Typed accessor for `netip.Addr` | ✅ `GetClientIPAddr` |
| Re-fronting / dual-connection attack | 🟡 out of scope (firewall layer) |
| RFC 7239 `Forwarded` header | 🟡 deliberately skipped (same problems as XFF, almost no adoption) |

---

TL;DR: replaces `middleware.RealIP` with four orthogonal `ClientIPFrom*`
middlewares + two accessors. All three GHSAs closed and pinned by named
tests. `r.RemoteAddr` is never mutated. Migration recipes in README +
godoc.

middleware/logger.go
```diff
@@ -118,7 +120,11 @@ func (l *DefaultLogFormatter) NewLogEntry(r *http.Request) LogEntry {
 	cW(entry.buf, useColor, nCyan, "%s://%s%s %s\" ", scheme, r.Host, r.RequestURI, r.Proto)
 
 	entry.buf.WriteString("from ")
-	entry.buf.WriteString(r.RemoteAddr)
+	clientIP := GetClientIP(ctx)
+	if clientIP == "" {
+		clientIP = r.RemoteAddr
+	}
+	entry.buf.WriteString(clientIP)
```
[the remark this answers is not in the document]
VojtechVitek: Thanks — this is intentional, not a regression.

The format change is bounded to users who **opt in** to a `ClientIPFrom*`
middleware. For everyone else `GetClientIP(ctx)` returns `""` and we fall
back to `r.RemoteAddr`, preserving the existing `IP:port` shape exactly.

Users migrating from `middleware.RealIP` also see no Logger change: `RealIP`
already mutated `r.RemoteAddr` to a port-less XFF value (that's the
long-standing [#453](https://github.com/go-chi/chi/issues/453)), so their
`from` field was already IP-only.

Forcing `IP:port` on the new path would mean either fabricating a port
(misleading) or pasting `r.RemoteAddr`'s port — which is the *immediate
proxy hop's* TCP source port, not the client's — onto the client IP. That's
actively wrong for any downstream parser that reads the port.

Custom formatters that genuinely need a stable composite shape can always
compose it explicitly: `GetClientIP(r.Context())` + `r.RemoteAddr`.

VojtechVitek:
Thanks @copilot.

- **`Header.Get` → last `Values()`** is a real defense-in-depth gap; fixed
  in `a8e3dbf` after pinning the four bypass shapes with failing tests in
  `22f567f`.
- **Godoc inconsistency on `ClientIPFromXFF()` no-args** fixed in `e692997`.
- **O(n²) walkXFF claim** — pushing back with a benchmark
(`5684b03`):
```
BenchmarkWalkXFF/n=1          5.6 ns/op
BenchmarkWalkXFF/n=10        57.6 ns/op    (5.8 ns/entry)
BenchmarkWalkXFF/n=100      541.1 ns/op    (5.4 ns/entry)
BenchmarkWalkXFF/n=1000    5228   ns/op    (5.2 ns/entry)
BenchmarkWalkXFF/n=10000  55148   ns/op    (5.5 ns/entry)
BenchmarkWalkXFF_RightmostStop/n=10000   6.1 ns/op   -- truly constant
```

middleware/client_ip.go
```diff
@@ -0,0 +1,263 @@
...
+//
+// v4-mapped IPv6 (::ffff:a.b.c.d) folds to plain v4 and IPv6 zones are
+// stripped before storage.
+func ClientIPFromHeader(trustedHeader string) func(http.Handler) http.Handler {
+	header := http.CanonicalHeaderKey(trustedHeader)
+	return func(h http.Handler) http.Handler {
+		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			values := r.Header.Values(header)
+			if len(values) > 0 {
+				if ip, ok := parseHeaderAddr(values[len(values)-1]); ok {
```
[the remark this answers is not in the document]
VojtechVitek: Empirically verified: Go's `net/http.ReadRequest` strips leading AND
trailing OWS (spaces and tabs) from header values during wire parsing.
Six variations tested including double-whitespace and tab cases, all
produce a clean stored value. So for production traffic the concern is
moot. Programmatic injection via `r.Header.Add(key, " value ")` is the
only path that preserves the OWS, and that's user-code responsibility
(test fixtures, custom adapters). Leaving the trim out keeps the hot
path strictly minimal.

VojtechVitek:
Thanks to Copilot and Claude for additional hardening of this PR.

This PR is now ready for final human reviews & merge. ✅

rezmoss:
@VojtechVitek  one thing before mrg, middleware.realip itself is unchanged, i verified that x-forwarded-for 127.0.0.1 still spoofs r.remoteaddr, if these three gh get a "patched version", govulncheck & dependabot will mark every chi ver after that release as safe, even though code still using middleware.realip remains fully exploitable, suggest keeping them as "no fix, realip deprecated, migrate to clientipfrom" instead of assigning a patched ver

VojtechVitek:
@rezmoss I've marked `middleware.RealIP` as deprecated in this PR. Anything else we can do here? We can't remove it from the go-chi package completely due to the semver compatibility promise.

VojtechVitek:
Released in https://github.com/go-chi/chi/releases/tag/v5.3.0

rezmoss:
@VojtechVitek could request cve plz

scop:
Website docs have not been updated per this yet, https://github.com/go-chi/docs/issues/18

VojtechVitek:
https://github.com/go-chi/docs/issues/18 fixed.

Next, I will work on httprate pkg update.

PR #968, opened 2024-12-23
feat:() creating a new BasicMiddleware

YlanzinhoY:
I grouped middleware.RequestID, middleware.Logger, middleware.Recoverer, and middleware.RealIP into a single function called BasicMiddleware to simplify the process. Instead of adding each middleware individually, you can just call BasicMiddleware, and it works seamlessly! I also Updated the Readme.MD to show how to use this new function.

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,7 +88,9 @@ func main() {
   r.Use(middleware.RealIP)
   r.Use(middleware.Logger)
   r.Use(middleware.Recoverer)
-
+  // or you can use
+  middleware.BasicMiddlewares()
+  
   // Set a timeout value on the request context (ctx), that will signal
   // through ctx.Done() that the request has timed out and further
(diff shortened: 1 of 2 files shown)


VojtechVitek:
Hi @YlanzinhoY,

thanks for the contribution. However, we don't consider these four basic middleware anymore.

```go
r.Use(Logger)
r.Use(Recoverer)
r.Use(RequestID)
r.Use(RealIP)
```

- `Logger` and `Recoverer` can be replaced by `github.com/go-chi/httplog/v2` (or soon `github.com/go-chi/httplog/v3`) with better support for `log/slog` primitives and structured logging
- `RequestID` has other alternatives such as https://github.com/go-chi/traceid
- and `RealIP` is soon to be deprecated in favor of a more secure implementation

PR #971, opened 2025-02-04
Use strings.Cut in a few places

JRaspass:
Now that we're 1.20+ instead of 1.14+ we can use 1.18's strings.Cut in a few places to simplify code.


VojtechVitek (approved):
LGTM. Nice work!

david-littlefarmer (approved):
LGTM

PR #973, opened 2025-02-05
replace hardcoded methods and statuses with standard constants

pixel365:
Somewhere in the newer code constants from the net/http package are used, and in some of the old code hardcoded values.
I suggest replacing the hardcoded values ​​with constants.


VojtechVitek:
Closing per #495 #625.

Issue #951, opened 2024-09-15
type *http.Request has no field or method SetPathValue when use tinygo build to wasm


cxjava:
I am attempting to utilize the Chi router in a WASM module, but I encounter an error during the build process with TinyGo. I'm uncertain whether the issue originates from Chi or TinyGo. Would you be able to assist in diagnosing the problem?

```shell
tinygo build -o ./build/app0.wasm -target wasm 
# github.com/go-chi/chi/v5
../../../../go/pkg/mod/github.com/go-chi/chi/v5@v5.1.0/path_value.go:18:5: r.SetPathValue undefined (type *http.Request has no field or method SetPathValue)

```

go version :``` go version go1.23.1 darwin/amd64```
tinygo version: ``` tinygo version 0.33.0 darwin/amd64 (using go version go1.23.1 and LLVM version 18.1.2) ```

```go 
package main

import (
	"net/http"

	"github.com/go-chi/chi/v5"
)

func main() {
	r := chi.NewRouter()
	r.Get("/", func(w http.ResponseWriter, r *http.Request) {
		w.Write([]byte("welcome"))
	})
	http.ListenAndServe(":3000", r)
}
```

VojtechVitek:
@angelofallars would you like to take a look?

It looks like the build tags in [path_value.go](https://github.com/go-chi/chi/blob/master/path_value.go) and [path_value_fallback.go](https://github.com/go-chi/chi/blob/master/path_value_fallback.go) introduced in https://github.com/go-chi/chi/pull/901 should account for tinygo too?

cxjava:
Hi @angelofallars  Could you help to take a look? Thank you!

efraimbart:
I've created a pull request to resolve this issue https://github.com/go-chi/chi/pull/978

PR #978, opened 2025-02-12
Add support for TinyGo

efraimbart:
Resolves https://github.com/go-chi/chi/issues/951

### DIFF
--- a/path_value.go
+++ b/path_value.go
@@ -1,5 +1,6 @@
-//go:build go1.22
-// +build go1.22
+//go:build go1.22 && !tinygo
+// +build go1.22,!tinygo
+
 
 package chi
 
--- a/path_value_fallback.go
+++ b/path_value_fallback.go
@@ -1,5 +1,5 @@
-//go:build !go1.22
-// +build !go1.22
+//go:build !go1.22 || tinygo
+// +build !go1.22 tinygo
 
 package chi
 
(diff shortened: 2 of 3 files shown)


VojtechVitek (approved):
LGTM. Thank you!

Can someone help me test this with TinyGo?
I'll wait for few more reviews and we can merge 👍 

cc @angelofallars @cxjava

cxjava:
build as expected, no error.

----

BTW:
when I want to use middleware in tinygo, it will failed because of `https://github.com/go-chi/chi/blob/master/middleware/profiler.go`, not package `net/http/pprof`  for tinygo.
```
	r := chi.NewRouter()
	r.Use(middleware.CleanPath)
	r.Use(middleware.RealIP)
	r.Use(middleware.Logger)
	r.Use(middleware.Recoverer)
```
error output:
```log
../../../go/pkg/mod/github.com/go-chi/chi/v5@v5.2.1/middleware/profiler.go:6:2: 
package net/http/pprof is not in std (/Users/user/Library/Caches/tinygo/goroot-598d56643c1f2b0a17cb7ed7f5003addae1fc3ee5af904a1c5356bb7118a4988/src/net/http/pprof)
```
Maybe we also need to exclude `middleware/profiler.go` when use tinygo.
image

efraimbart:
Unfortunately I'm not familiar with middleware or what would happen if it's excluded.

cxjava:
No worry, I can create other PR to fix the middleware build.

VojtechVitek (approved):
Thanks, I'll merge both #978 and #982.

LGTM

PR #994, opened 2025-04-29
Refactor graceful shutdown example

mikereid1:
The current example for graceful shutdown is a bit convoluted and dated as it doesn't take advantage of `signal.NotifyContext`


OnatArslan:
#### Nice Work!!!!

PR #1001, opened 2025-05-26
Change install command format to code block

sglkc:
Improves developer experience by changing the install command format from single backtick (code) to double backticks (code block), providing copy to clipboard button.


VojtechVitek (approved):
LGTM

PR #1007, opened 2025-07-04
Update compress middleware support Brotli support documentation

sm3142:
### What
Update inline documentation on how to add Brotli support to a compress middleware instance.

### Why
The code sample in the current documentation is referencing a Brotli package that has been archived more than 5 years ago.

### How
Update the documentation to the use the Google's "reference" implementation Go module (github.com/google/brotli/go/cbrotli). PR also documents an alternative approach using a Cgo-free module, which can be easier to deal with if performance is not an absolute priority. Both modules are being actively maintained.

### DIFF
--- a/middleware/compress.go
+++ b/middleware/compress.go
@@ -136,13 +136,11 @@ func NewCompressor(level int, types ...string) *Compressor {
 //
 // For example, add the Brotli algorithm:
 //
-//	import brotli_enc "gopkg.in/kothar/brotli-go.v0/enc"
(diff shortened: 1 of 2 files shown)


VojtechVitek (approved):
Thanks @sm3142 — good catch, the kothar package has been archived for years, so this was overdue. I compile-tested both examples and they're correct. Merging.

One note for readers landing here: the cbrotli example needs cgo and the brotli C headers installed; the `andybalholm/brotli` builds with plain `go get`.

PR #1011, opened 2025-08-01
Optimize throttle middleware by avoiding unnecessary timer creation

vasayxtx:
- Only create time.NewTimer() when actually needed to wait
- Reduces allocations from 7 to 4 and memory usage by ~54%
- Improves performance by ~50% in high-throughput scenarios

**Benchmark results:**

Before:
```
goos: darwin
goarch: arm64
pkg: github.com/go-chi/chi/v5/middleware
cpu: Apple M1 Max
BenchmarkThrottle
BenchmarkThrottle-10    	 3349486	       382.3 ns/op	     456 B/op	       7 allocs/op
PASS
```

After:
```
goos: darwin
goarch: arm64
pkg: github.com/go-chi/chi/v5/middleware
cpu: Apple M1 Max
BenchmarkThrottle
BenchmarkThrottle-10    	 6310284	       186.9 ns/op	     208 B/op	       4 allocs/op
PASS
```

### DIFF
--- a/middleware/throttle.go
+++ b/middleware/throttle.go
@@ -83,12 +83,23 @@ func ThrottleWithOpts(opts ThrottleOpts) func(http.Handler) http.Handler {
 				return
 
 			case btok := <-t.backlogTokens:
-				timer := time.NewTimer(t.backlogTimeout)
-
 				defer func() {
 					t.backlogTokens <- btok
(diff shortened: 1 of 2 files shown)


VojtechVitek (approved):
LGTM

PR #1013, opened 2025-08-04
Allow multiple whitespace between method & pattern

JRaspass:
Allow separating the method and pattern with `[ \t]+`, this matches the stdlib change that was made in https://github.com/golang/go/commit/7b583fd

The rationale is the same as the stdlib issue, it allows you to visually line up patterns with variable-width methods:
```
r.Handle("GET    /my-route", handler1)
r.Handle("POST   /my-route", handler2)
r.Handle("DELETE /my-route", handler3)
```
In general aligning ourselves with the stdlib brings less surprise for end users.


VojtechVitek (approved):
LGTM. Thank you!

PR #1014, opened 2025-08-08
Fix/405 allow dedupe and param alignment

srpvpn:
Deduplicate and sort methods in Allow for 405 responses to avoid duplicates and unstable order. Restore param/value alignment when regex param branch doesn’t match to prevent panics with Go 1.22 PathValue. All tests pass. Change is backward-compatible; only stabilizes headers and fixes a runtime edge case.

### DIFF
--- a/mux.go
+++ b/mux.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"net/http"
+	"slices"
 	"strings"
 	"sync"
 )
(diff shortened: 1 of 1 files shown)


srpvpn:
Ok,  I'll keep an eye on #1017 and rebase once it's merged.

srpvpn:
@VojtechVitek Rebased onto the latest master, all checks are green now

Issue #608, opened 2021-03-25
Throttle middleware tests fail intermittently (quite often)


pkieltyka:
https://github.com/go-chi/chi/blob/master/middleware/throttle_test.go

It's often that when I submit a PR, and the code runs through github ci, it will fail -- https://github.com/go-chi/chi/actions 

I have to re-run the jobs to have it pass, it seems to be a timing issue. When running locally, it seems to work well pretty consistently, but I've seen the rare failing case.

VojtechVitek:
I didn't write those tests, but I can take a look if we have an example of the CI error.

pkieltyka:
I've just disabled the flaky test, and will rename this ticket..

https://github.com/go-chi/chi/commit/25eb15cdd4f644896eac2ac05d4e3e932f34a188

VojtechVitek:
@OtavioBernardes sound good!

Please run it with `go test -count=1000 -run=TestThrottleRetryAfter ./...`  locally to ensure the flakiness is really gone 👍

PR #1016, opened 2025-08-14
fix/608 - Fix flaky Throttle middleware test by synchronizing token usage

OtavioBernardes:
## Description:

The TestThrottleRetryAfter test was intermittently failing on CI due to a timing issue. The original test depended on real-time sleeps and concurrent goroutines completing at slightly different times, which made token consumption non-deterministic. As a result sometimes the test would incorrectly pass or fail depending on the Go scheduler or CI load.

## Problem:
- The test used time.Sleep in the handler and between request batches.
- Tokens in the throttler bucket were being released as soon as the handler finished, which could happen before the second batch of requests was issued, causing expected 429s to be missed.

## Solution:
Introduced a blockCh channel in the handler to control exactly when tokens are released. First batch of requests consumes all tokens and waits on blockCh, keeping the tokens occupied.

#608

### DIFF
--- a/middleware/throttle_test.go
+++ b/middleware/throttle_test.go
@@ -202,49 +202,47 @@ func TestThrottleMaximum(t *testing.T) {
 func TestThrottleRetryAfter(t *testing.T) {
 	r := chi.NewRouter()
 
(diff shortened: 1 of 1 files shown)


VojtechVitek (approved):
LGTM. Thank you for your contribution! 🎉

PR #1017, opened 2025-08-18
Bump minimum Go to 1.22 and use new features

JRaspass:
Go 1.21 brings us the builtin min func and slices.Contains.

Go 1.22 brings us range over int and fixed for loop scoping.

We were also able to drop the build tags from the path value code and inline the three-line function directly. This should still work on tinygo as they claim to support Go 1.24.


mux.go
```diff
@@ -467,8 +467,10 @@ func (mx *Mux) routeHTTP(w http.ResponseWriter, r *http.Request) {
 
 	// Find the route
 	if _, _, h := mx.tree.FindRoute(rctx, method, routePath); h != nil {
-		if supportsPathValue {
-			setPathValue(rctx, r)
+		// Set http.Request path values from our request context
+		for i, key := range rctx.URLParams.Keys {
+			value := rctx.URLParams.Values[i]
+			r.SetPathValue(key, value)
```
VojtechVitek, line 473: Looks like an unrelated change

tree.go
```diff
@@ -772,20 +770,14 @@ func patParamKeys(pattern string) []string {
 	}
 }
 
-// longestPrefix finds the length of the shared prefix
-// of two strings
-func longestPrefix(k1, k2 string) int {
-	max := len(k1)
```
VojtechVitek, line 778: interesting .. so this variable was named incorrectly - it meant "min" instead?

tree.go
```diff
@@ -772,20 +770,14 @@ func patParamKeys(pattern string) []string {
...
-	for i = 0; i < max; i++ {
+// longestPrefix finds the length of the shared prefix of two strings
+func longestPrefix(k1, k2 string) (i int) {
+	for i = 0; i < min(len(k1), len(k2)); i++ {
 		if k1[i] != k2[i] {
 			break
 		}
 	}
-	return i
+	return
```
VojtechVitek, line 780: This look good

VojtechVitek: One thing I noticed -- (unrelated to this PR) -- this function might not work well with unicode characters, as it works on bytes and not on runes .. the question is are we ever passing unicode strings into this function (e.g. `"/café"`, `"/café/1"`). Can this happen?

VojtechVitek (approved):
LGTM. Thank you!

I'll wait for more 👀 to review and then we can merge. 👍

benstigsen:
LGTM

PR #1019, opened 2025-08-27
Refactor to use atomic type

cuiweixie:


VojtechVitek (commented):
LGTM. Is there any benefit here beyond nicer syntax?

cuiweixie:
> LGTM. Is there any benefit here beyond nicer syntax?

Encapsulation of atomic behavior –
The atomic.Uint64 type ensures that all accesses go through its safe methods, preventing unsafe reads/writes that could cause data races.

Less error-prone –
With the old approach, it was possible to mistakenly read or write reqid directly without atomic operations. With atomic.Uint64, you can’t access the underlying value without using its atomic methods.

PR #1022, opened 2025-10-07
update reverseMethodMap in RegisterMethod

### LINKED ISSUE #1021: chi.Walk no longer sees routes with custom HTTP methods
In v5.2.3, `chi.Walk` no longer visits routes with custom HTTP methods. For example:
```
package main

import (
	"fmt"
	"net/http"

	"github.com/go-chi/chi/v5"
)

func main() {
	r := chi.NewMux()

	chi.RegisterMethod("QUACK")
	r.MethodFunc("QUACK", "/quack", func(w http.ResponseWriter, r *http.Request) {
		w.WriteHeader(http.StatusOK)
	})

	chi.Walk(r, func(method string, route string, handler http.Handler, _ ...func(http.Handler) http.Handler) error {
		fmt.Println(method, route)
		return nil
	})
}
```

I believe the change responsible for this is https://github.com/go-chi/chi/commit/9b9fb55def404397748a9fc7e044efe9db1d618e, which changed method name lookup to use `reverseMethodMap` instead of a utility function which iterated `methodMap`. As far as I can tell, the problem is that `reverseMethodMap` is not updated by `RegisterMethod`, so the lookup fails and the route is not added.

cixel:
Fixes #1021.

### DIFF
--- a/mux_test.go
+++ b/mux_test.go
@@ -1782,6 +1782,26 @@ func TestCustomHTTPMethod(t *testing.T) {
 	if _, body := testRequest(t, ts, "BOO", "/hi", nil); body != "custom method" {
 		t.Fatal(body)
 	}
+
+	var expectRoutes = map[string]string{
+		"GET": "/",
+		"BOO": "/hi",
+	}
+	Walk(r, func(method string, route string, handler http.Handler, _ ...func(http.Handler) http.Handler) error {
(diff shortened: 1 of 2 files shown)


VojtechVitek (approved):
LGTM. Thank you!

PR #985, opened 2025-02-23
Add pathvalue example to README and implement PathValue handler.

catatsuy:
This pull request includes updates to the `chi` examples, specifically adding a new example that demonstrates the usage of `PathValue` for retrieving URL parameters. The most important changes include the addition of a new example in the `_examples/README.md` file and the implementation of the `pathvalue` example in a new Go file.

New example addition:

* [`_examples/README.md`](diffhunk://#diff-f370007ac2f128eca652bd2acedeb98c49ad44e517ac7c8a9294668a7acb4bd1R15): Added a new entry for the `pathvalue` example, which demonstrates the usage of `PathValue` for retrieving URL parameters.

Implementation of the `pathvalue` example:

* [`_examples/pathvalue/main.go`](diffhunk://#diff-61eb4288dd3426b953600ac17e2cb836aa7b6855ae09f0c47f9ead6016400038R1-R25): Created a new Go file that includes a router setup with a handler to retrieve and respond with a URL parameter using `PathValue`.

refs: #984

PR #1025, opened 2025-10-16
docs: PathValue/SetPathValue with small examples (#984)

### LINKED ISSUE #984: Add documentation for PathValue and SetPathValue support in README
Thank you for maintaining chi!

The support for PathValue and SetPathValue was added in https://github.com/go-chi/chi/pull/901, which is a great feature. However, this is not mentioned in the README.

Some users might assume that chi does not support this feature and choose another router instead. Since this is part of Go's standard mechanism, I believe it should be recommended in the documentation.

Would it be possible to update the README to mention this support?

Thank you!

MichailBaldin:
Added a short README section under URL parameters describing the Go 1.22+ PathValue / SetPathValue support with a small example.
This clarifies that chi populates path values from the standard library and shows how to read or adjust them directly from *http.Request.

Closes #984

### DIFF
--- a/README.md
+++ b/README.md
@@ -325,6 +325,24 @@ func MyRequestHandler(w http.ResponseWriter, r *http.Request) {
   w.Write([]byte(fmt.Sprintf("hi %v, %v", userID, key)))
 }
 ```
+chi supports the path parameter mechanism introduced in Go’s standard library (Go 1.22+).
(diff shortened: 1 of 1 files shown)


VojtechVitek:
Thanks, but #984 was addressed by #985's example. The README snippet wouldn't compile. Closing.

PR #1026, opened 2025-10-20
docs: clarify With() middleware comment

ryutaKimu:
The original comment: // With adds inline middlewares for an endpoint handler. 

this does not indicate that the middleware is applied only to specific routes. It only states that the middleware is added to an endpoint handler, so it is not clear whether it affects all routes or just a subset. Adding "applying them only to the specific route(s) defined on it" makes this behavior explicit and easier for developers to understand.

### DIFF
--- a/chi.go
+++ b/chi.go
@@ -70,7 +70,8 @@ type Router interface {
 	// Use appends one or more middlewares onto the Router stack.
 	Use(middlewares ...func(http.Handler) http.Handler)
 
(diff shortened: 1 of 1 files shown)


VojtechVitek:
Sorry, I don't think this wording makes the godoc clearer -- "defined on it" is ambiguous, since With returns a new inline Router and the middlewares only apply to routes registered on that returned router. Closing for now.

PR #1028, opened 2025-10-23
Add ability to poison contexts, preventing reuse

BlakeWilliams:
I’m working with several teams to extract new services, replacing existing API endpoints one by one. To validate these changes we're comparing responses from the legacy ("control") service against the new ("candidate") service. We initially started implementing this in `chi`, but ran into race conditions due to `sync.Pool` reusing `chi.Context` between requests. This is due to us returning the "control" response immediately, followed by calling the "candidate" response in the background (which utilizes `context`) outside of the original req/res lifecycle.

We have some workarounds in place, but ideally chi would allow more control over context reuse for this use-case. I took a pretty basic swing at this, implementing a new `PreventReuse()` method on `Context` that marks the `Context` as `poisoned` and avoiding the `sync.Pool.Put` call appropriately.

I totally understand if this is out of scope or isn't an API you wanted to support, but I wanted to open a PR to get feedback and see if there's interest instead of assuming.

Either way, the good news here is that the new services are using chi. ;)

### DIFF
--- a/context.go
+++ b/context.go
@@ -76,6 +76,12 @@ type Context struct {
 
 	methodsAllowed   []methodTyp // allowed methods in case of a 405
 	methodNotAllowed bool
+	poisoned         bool // prevents returning to pool
+}
+
+// PreventReuse marks the routing context as being unable to be returned to the pool.
+func (x *Context) PreventReuse() {
+	x.poisoned = true
 }
 
 // Reset a routing context to its initial state.
--- a/context.go
+++ b/context.go
@@ -79,7 +79,8 @@ type Context struct {
 	poisoned         bool // prevents returning to pool
 }
 
-// PreventReuse marks the routing context as being unable to be returned to the pool.
+// PreventReuse marks the Context as being unable to be returned to the pool
+// and allowing it to be safely used outside of the running request.
 func (x *Context) PreventReuse() {
 	x.poisoned = true
 }
--- a/mux.go
+++ b/mux.go
@@ -88,7 +88,10 @@ func (mx *Mux) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 
 	// Serve the request and once its done, put the request context back in the sync pool
 	mx.handler.ServeHTTP(w, r)
-	mx.pool.Put(rctx)
+
+	if !rctx.poisoned {
+		mx.pool.Put(rctx)
+	}
 }
 
 // Use appends a middleware handler to the Mux middleware stack.
--- a/mux_test.go
+++ b/mux_test.go
@@ -1980,6 +1980,29 @@ func TestServerBaseContext(t *testing.T) {
 	}
 }
 
+func TestServerPoisonedContext(t *testing.T) {
+	r := NewRouter()
+	var firstContext *Context
+	r.Get("/first", func(w http.ResponseWriter, r *http.Request) {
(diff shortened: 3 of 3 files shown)


VojtechVitek:
Hi @BlakeWilliams,

Can you share an example reproducer, so we understand the the issue you're running into?

Are you keeping `chi.Context` around in a new goroutine outside of the Request scope?

BlakeWilliams:
Here's a small repro that's close enough to what we're doing:

```go
// main.go
package main

import (
	"context"
	"fmt"
	"log"
	"net/http"
	"time"

	"github.com/go-chi/chi/v5"
)

var request int

type Result struct {
	Status int
	Body   []byte
}

type Experiment struct {
	Control   func(context.Context) Result
	Candidate func(context.Context) Result
}

func NewExperiment(control, candidate func(context.Context) Result) *Experiment {
	return &Experiment{Control: control, Candidate: candidate}
}

func (e *Experiment) Run(ctx context.Context) Result {
	ctrl := e.Control(ctx)

	go func(ctrl Result) {
		// Exagerated delay to help showcase the race
		time.Sleep(500 * time.Millisecond)
		_ = e.Candidate(ctx)

		// log difference
	}(ctrl)

	return ctrl
}

func helloHandler(w http.ResponseWriter, r *http.Request) {
	name := chi.URLParam(r, "name") // captured before response returns

	rc := chi.RouteContext(r.Context())
	routePath := rc.RoutePattern()

	rc.PreventReuse()

	exp := NewExperiment(
		func(ctx context.Context) Result {
			return Result{
				Status: http.StatusOK,
				Body:   []byte("control hello " + name + "\n"),
			}
		},
		func(ctx context.Context) Result {
			rc := chi.RouteContext(r.Context())
			candidateRoutePath := rc.RoutePattern()

			if routePath != candidateRoutePath {
				fmt.Println("route paths do not match:", routePath, candidateRoutePath)
			}
			return Result{
				Status: http.StatusOK,
				Body:   []byte("candidate hello " + name + "\n"),
			}
		},
	)

	ctrl := exp.Run(r.Context())

	w.WriteHeader(ctrl.Status)
	_, _ = w.Write(ctrl.Body)
}

func main() {
	r := chi.NewRouter()

	r.Get("/hello/{name}", helloHandler)
	r.Get("/goodbye/{name}", helloHandler)

	addr := ":3005"
	log.Printf("listening on %s", addr)
	if err := http.ListenAndServe(addr, r); err != nil {
		log.Fatal(err)
	}
}
```

For `main` remove `rc.PreventReuse()`, but if running on this branch keep it to see the error message no longer being logged. I was running this to get the mismatch logs: `for i in {1..5}; do curl localhost:3005/hello/world & curl localhost:3005/goodbye/context; done; wait`

VojtechVitek:
Let's go for `rctx := r.Context().Value(chi.RouteCtxKey).(*Context)`.

Perhaps we can introduce a `.Clone()` method on the Context type. Would that work for you?

BlakeWilliams:
> Perhaps we can introduce a `.Clone()` method on the Context type. Would that work for you?

That would definitely work! We really just need a "blessed" way to work around the race condition. 

I still like the idea of excluding it from being put back into the pool since it eliminates a large potential for logic errors (e.g. we forget to add a field to `Clone`) but could definitely work with that.

PR #1033, opened 2025-11-05
Add Clone method to Context

BlakeWilliams:
This adds the ability to clone Chi contexts so that they may be used outside of the request/response lifecycle.

This is useful in scenarios where the context may be held _after_ a request has been served to a user, like asynchronous background tasks, mirror testing, etc.

This is a follow-up and replacement of https://github.com/go-chi/chi/pull/1028

### DIFF
--- a/context.go
+++ b/context.go
@@ -95,6 +95,22 @@ func (x *Context) Reset() {
 	x.parentCtx = nil
 }
 
+// Clone a routing context so that it may be used outside of the request/response lifecycle.
+func (c *Context) Clone() *Context {
+	clone := *c
+
+	clone.URLParams.Keys = append([]string(nil), c.URLParams.Keys...)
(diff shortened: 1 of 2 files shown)


VojtechVitek:
Thanks, this is the direction I wanted — copy over poisoning the pool. Three asks before merge:

1. **Document the timing constraint.** `Clone()` must be called inside the handler, before it returns. Cloning later from a background goroutine races with `Reset()` just like before. Let's spell this out in the doc comment, it's the one thing that will bite users.
2. **Nil out `parentCtx` in the clone.** Nothing reads it after routing, and keeping it pins the finished request's context in memory.
3. **Add a note on the `Context` struct** that new reference fields must be deep-copied in `Clone()` — you flagged this drift risk yourself in #1028.

Nit: use `x` as the receiver name to match the rest of the file.

VojtechVitek:
Merged in https://github.com/go-chi/chi/pull/1157.

PR #1038, opened 2025-12-15
Spelling and grammar fixes

scop:


VojtechVitek:
Thanks for the cleanup — the changes here are all correct ("an HTTP"/"an http.Request", the `it's` contractions, and `I.e.`).

One thing worth flagging: **the sweep is incomplete**, so the same mistakes it fixes still live elsewhere in the library (ignoring `_examples/`). If the goal is consistency, it'd be better to land this as one complete pass rather than leaving the codebase half-converted:

**`a http` → `an http` still remaining:**
- `chain.go` — 4× `a http.Handler` (whole file untouched)
- `mux.go:414` & `mux.go:439` — `a particular http method` (these are in the very comments this PR edits)
- `middleware/recoverer.go:18` — `a HTTP 500`

**`its` → `it's` still remaining:**
- `middleware/get_head.go:27` — `through its a GET route`
- `middleware/supress_notfound.go:14` — `its super annoying`
- `tree_test.go:173` — `its not possible`

**`ie.` → `I.e.` still remaining:**
- `middleware/profiler.go:14` — `ie.` (matches the `timeout.go` fix made here)

Would you mind extending the PR to cover these so the fix is consistent across the package? A quick `grep -rn "\ba http\b\|\ba HTTP\b" --include="*.go" .` (plus the `its`/`ie.` variants) catches the rest.

---
*Automated review by Claude.*

scop:
Addressed those and some others. Additionally, the filename middleware/supress_notfound.go would be correctly spelled suppress_notfound.go (two p's), but I did not do that here. Maybe you can take care of that if you'd like to see that?

scop:
@pkieltyka I'm wondering why this was closed without comment, was that intentional? I had just addressed comments raised by @VojtechVitek

PR #1046, opened 2026-02-07
Use strings.ReplaceAll where applicable

JRaspass:
We already use it in the codebase, it's been around since 1.12, and is considered best practice since it avoids hardcoding a magic number.


VojtechVitek (approved):
LGTM

PR #1048, opened 2026-02-11
Bump minimum Go to 1.23, always use request.Pattern

JRaspass:
Go 1.26 is out now so we can bump our minimum Go to 1.23 and simplify the code.


VojtechVitek (approved):
LGTM. Thank you.

PR #1050, opened 2026-02-14
Fix defaultLogEntry.Panic not respecting NoColor setting

### LINKED ISSUE #1042: middleware.Logger: defaultLogEntry.Panic() does not respect 'NoColor' of DefaultLogFormatter

doganarif:
`defaultLogEntry.Panic()` calls `PrintPrettyStack()` which hardcodes
`useColor = true`, ignoring the `NoColor` field on `DefaultLogFormatter`.

This threads the `useColor` flag from `defaultLogEntry` through to
`prettyStack.parse()`. The public `PrintPrettyStack` function keeps its
existing behavior (always colored).

Fixes #1042


VojtechVitek (approved):
LGTM

PR #1057, opened 2026-03-04
feat(middleware): compress all text/* types by default

acme:
- Replace explicit `text/html`, `text/css`, `text/plain`, `text/javascript` entries in `defaultCompressibleContentTypes` with a single `text/*` wildcard
- Fix `NewCompressor` to parse wildcard patterns in the defaults path, not only when types are explicitly passed
- Add test asserting `text/markdown` is gzip-compressed with default settings

I realise there have been a number of compression-related pull requests in the past, but hopefully this is small enough and has tests so is easy to merge in.

### DIFF
--- a/middleware/compress.go
+++ b/middleware/compress.go
@@ -14,10 +14,7 @@ import (
 )
 
 var defaultCompressibleContentTypes = []string{
-	"text/html",
-	"text/css",
-	"text/plain",
-	"text/javascript",
+	"text/*",
 	"application/javascript",
 	"application/x-javascript",
 	"application/json",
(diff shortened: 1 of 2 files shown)


JRaspass:
This needs a rebase but other than that seems like a really nice improvement that brings symmetry between user-supplied and default compression content types. I wonder if we might even be able to merge the loops now that they both support wildcards?

What do you think of this change @VojtechVitek ?

acme:
(Sorry, I messed up while trying to rebase, sigh)

VojtechVitek:
Thanks for the PR and the test, @acme — appreciated!

We decided to extend the explicit whitelist instead of going wildcard, see #1151. Two reasons:

1. **Performance** — exact entries keep `isCompressible()` on the single-map-lookup hot path for the common types.
2. **SSE** — `text/*` would silently start compressing `text/event-stream` by default, which can delay events behind buffering proxies.

If you want blanket coverage, `Compress(5, "text/*")` already works — the wildcard mechanism is there for explicit opt-in. Closing this one in favor of #1151.

VojtechVitek:
Thank you for the contribution!

VojtechVitek:
please review https://github.com/go-chi/chi/pull/1151

PR #1066, opened 2026-03-19
Add tests for panic behavior on malformed route patterns

fheikens:
## Summary

Adds test coverage for five existing panic guard paths in route pattern
validation that were not explicitly covered by tests. No production code was changed.

Covered cases:
- Unclosed `{` brace in route pattern (`tree.go:patNextSegment`)
- Duplicate parameter keys in pattern (`tree.go:patParamKeys`) — addresses the TODO at `tree_test.go:194`
- Invalid regexp in route parameter (`tree.go:addChild`)
- `Route()` called with nil function (`mux.go:Route`)
- `Mount()` called with nil handler (`mux.go:Mount`)

## Test plan
- [x] All five new tests pass: `go test -v -run TestPanic ./...`
- [x] Full test suite passes: `go test ./...`
- [x] No production code modified

### DIFF
--- a/panic_test.go
+++ b/panic_test.go
@@ -1,0 +1,61 @@
+package chi
+
+import (
+	"net/http"
+	"testing"
+)
+
+func TestPanicOnUnclosedBrace(t *testing.T) {
+	defer func() {
+		if recover() == nil {
+			t.Error("expected panic()")
+		}
+	}()
+
+	r := NewRouter()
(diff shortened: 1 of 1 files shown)


fheikens:
Good point — tightened the pattern to `{id:(abc}` (unclosed group) so the failure is unambiguously from `regexp.Compile`, not from route pattern parsing. Test name stays accurate.

PR #1068, opened 2026-03-25
Fix httpFancyWriter.ReadFrom double-counting BytesWritten with Tee

### LINKED ISSUE #1067: httpFancyWriter.ReadFrom double-counts BytesWritten when Tee is set
## Description

```go
	if f.basicWriter.tee != nil {
		return n, err
	}
}
```

## Reproduction

```go
		HeaderMap: make(http.Header),
		Body:      new(bytes.Buffer),
	}

	var teeBuf bytes.Buffer
	f.Tee(&teeBuf)

	input := "hello world"
	// n == 11, err == nil
}
```

## Fix

queelius:
## Summary
- Fix `httpFancyWriter.ReadFrom` double-counting bytes when a `Tee` writer is set
- `io.Copy(&f.basicWriter, r)` routes through `basicWriter.Write()` which already increments `bytes`; the additional `f.basicWriter.bytes += int(n)` after `io.Copy` caused `BytesWritten()` to return 2x the actual value
- Add regression test that verifies `BytesWritten()` accuracy after `ReadFrom` with `Tee`

Fixes #1067

## Test plan
- [x] New test `TestHttpFancyWriterReadFromWithTee` fails before fix, passes after
- [x] All existing tests pass (`go test ./...`)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/middleware/wrap_writer.go
+++ b/middleware/wrap_writer.go
@@ -208,8 +208,9 @@ func (f *http2FancyWriter) Push(target string, opts *http.PushOptions) error {
 
 func (f *httpFancyWriter) ReadFrom(r io.Reader) (int64, error) {
 	if f.basicWriter.tee != nil {
+		// io.Copy will call basicWriter.Write, which already
+		// increments basicWriter.bytes, so we must not add again.
 		n, err := io.Copy(&f.basicWriter, r)
-		f.basicWriter.bytes += int(n)
 		return n, err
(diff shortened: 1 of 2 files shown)


queelius:
Closing this to keep your queue clean. Happy to reopen if useful, or feel free to pick up any of it. Thanks!

PR #1070, opened 2026-03-25
Fix compress middleware Accept-Encoding parsing (substring + q=0)

### LINKED ISSUE #1069: Compress middleware: matchAcceptEncoding uses substring match and ignores q=0
## Description

The `matchAcceptEncoding` function in `middleware/compress.go` uses `strings.Contains` to match encoding names in the `Accept-Encoding` header. This has two problems:

### 1. Substring matching produces false positives

```go
func matchAcceptEncoding(accepted []string, encoding string) bool {
	for _, v := range accepted {
		if strings.Contains(v, encoding) {   // substring match, not exact
			return true
		}
	}
	return false
}
```

For example:
- `Accept-Encoding: br` incorrectly matches encoding `b` (since `"br"` contains `"b"`)
- `Accept-Encoding: bgzip` incorrectly matches encoding `gzip`

### 2. Quality value `q=0` is ignored

Per [RFC 9110 Section 12.5.3](https://httpwg.org/specs/rfc9110.html#field.accept-encoding), a quality value of `q=0` means the encoding is **not acceptable**. But `strings.Contains("gzip;q=0", "gzip")` returns `true`, so the middleware will compress with gzip even when the client explicitly rejects it.

## Reproduction

```go
// These all incorrectly return true:
matchAcceptEncoding([]string{"gzip;q=0"}, "gzip")   // should be false (q=0 = not acceptable)
matchAcceptEncoding([]string{"br"}, "b")             // should be false (not exact match)
matchAcceptEncoding([]string{"bgzip"}, "gzip")       // should be false (not exact match)
```

## Fix

Parse the encoding name properly by splitting on `;` to separate quality parameters, trimming whitespace, performing exact string comparison, and rejecting `q=0`.

queelius:
## Summary
- Fix `matchAcceptEncoding` to use exact encoding name comparison instead of `strings.Contains` substring matching
- Respect `q=0` quality values per RFC 9110 Section 12.5.3, which means "encoding not acceptable"
- Previously, `Accept-Encoding: gzip;q=0` would still compress with gzip, and `Accept-Encoding: br` could incorrectly match a hypothetical encoder named `b`

Fixes #1069

## Test plan
- [x] New `TestMatchAcceptEncoding` unit tests cover exact matching, whitespace handling, quality values, and substring rejection
- [x] New `TestCompressorRespectsQualityZero` integration test confirms end-to-end fix
- [x] All existing tests pass (`go test ./...`)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/middleware/compress.go
+++ b/middleware/compress.go
@@ -254,7 +254,10 @@ func matchAcceptEncoding(accepted []string, encoding string) bool {
 		if params != "" {
 			params = strings.TrimSpace(params)
 			if qval, ok := strings.CutPrefix(params, "q="); ok {
-				if q, err := strconv.ParseFloat(strings.TrimSpace(qval), 32); err == nil && q == 0 {
+				// Strip any trailing parameters after the quality value.
+				qval, _, _ = strings.Cut(qval, ";")
+				qval = strings.TrimSpace(qval)
+				if q, err := strconv.ParseFloat(qval, 32); err == nil && q == 0 {
 					continue
 				}
 			}
--- a/middleware/compress.go
+++ b/middleware/compress.go
@@ -9,6 +9,7 @@ import (
 	"io"
 	"net"
 	"net/http"
+	"strconv"
 	"strings"
 	"sync"
 )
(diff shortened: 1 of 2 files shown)


queelius:
Closing this to keep your queue clean. Happy to reopen if useful, or feel free to pick up any of it. Thanks!

PR #1071, opened 2026-03-26
Fix httpFancyWriter.ReadFrom double-counting BytesWritten with Tee

### LINKED ISSUE #1067: httpFancyWriter.ReadFrom double-counts BytesWritten when Tee is set

boinger:
## Summary

`httpFancyWriter.ReadFrom` double-counts `BytesWritten()` when a Tee writer is set.

When `tee != nil`, `ReadFrom` uses `io.Copy(&f.basicWriter, r)` which routes writes through `basicWriter.Write()`. `Write` already increments `basicWriter.bytes` on line 112. However, `ReadFrom` also adds `n` to `basicWriter.bytes` again on line 212, doubling the reported count.

The non-tee path is correct — `rf.ReadFrom` bypasses `basicWriter.Write`, so its manual increment is needed.

`http2FancyWriter` does not implement `ReadFrom`; HTTP/2 handles streaming at the framing layer.

## Fix

Remove the redundant `f.basicWriter.bytes += int(n)` on the tee path. `io.Copy` through `basicWriter.Write` handles byte counting.

## Test

Added `TestFancyWriterReadFromBytesCounting` with subtests for both the tee and non-tee paths, verifying `BytesWritten()` returns the correct count (not 2x).

Fixes #1067

### DIFF
--- a/middleware/wrap_writer.go
+++ b/middleware/wrap_writer.go
@@ -208,9 +208,8 @@ func (f *http2FancyWriter) Push(target string, opts *http.PushOptions) error {
 
 func (f *httpFancyWriter) ReadFrom(r io.Reader) (int64, error) {
 	if f.basicWriter.tee != nil {
-		n, err := io.Copy(&f.basicWriter, r)
-		f.basicWriter.bytes += int(n)
-		return n, err
+		// io.Copy calls basicWriter.Write, which already increments bytes.
+		return io.Copy(&f.basicWriter, r)
 	}
 	rf := f.basicWriter.ResponseWriter.(io.ReaderFrom)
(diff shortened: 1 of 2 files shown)


boinger:
Closing as duplicate of #1068, which was submitted by the issue author. Apologies for the noise — I missed the existing PR during my survey. The fix is identical.

PR #1085, opened 2026-04-13
middleware: fix httpFancyWriter.ReadFrom double-counting bytes with Tee

### LINKED ISSUE #1067: httpFancyWriter.ReadFrom double-counts BytesWritten when Tee is set

alliasgher:
## Summary

`httpFancyWriter.ReadFrom` in `middleware/wrap_writer.go` double-counted `BytesWritten()` when a Tee writer was set.

When `tee != nil`, the code called `io.Copy(&f.basicWriter, r)`, which routes every write through `basicWriter.Write`. `Write` already increments `basicWriter.bytes`. After the copy returned, the old code then did:

```go
f.basicWriter.bytes += int(n)   // BUG: already counted inside Write
```

…adding `n` a second time. The result: `BytesWritten()` returned exactly **twice** the actual number of bytes transferred.

## Fix

Remove the redundant `f.basicWriter.bytes += int(n)` line in the tee branch. The non-tee path (which delegates to the underlying `ReaderFrom`, bypassing `Write`) is not affected and still needs to add `n` itself.

## Test

Added `TestHttpFancyWriterReadFromByteCountWithTee` that reproduces the original bug (it fails before the fix and passes after). The full middleware test suite still passes.

Fixes #1067

### DIFF
--- a/middleware/wrap_writer.go
+++ b/middleware/wrap_writer.go
@@ -208,8 +208,10 @@ func (f *http2FancyWriter) Push(target string, opts *http.PushOptions) error {
 
 func (f *httpFancyWriter) ReadFrom(r io.Reader) (int64, error) {
 	if f.basicWriter.tee != nil {
+		// Route through basicWriter.Write so that data is also written to the
+		// tee writer. basicWriter.Write already increments basicWriter.bytes,
+		// so we must NOT add n again here (that would double-count).
 		n, err := io.Copy(&f.basicWriter, r)
-		f.basicWriter.bytes += int(n)
 		return n, err
 	}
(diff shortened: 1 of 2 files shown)


VojtechVitek (approved):
LGTM. Thank you for the fix!

PR #1086, opened 2026-04-13
middleware: honour NoColor in defaultLogEntry.Panic

### LINKED ISSUE #1042: middleware.Logger: defaultLogEntry.Panic() does not respect 'NoColor' of DefaultLogFormatter

alliasgher:
## Summary

`defaultLogEntry.Panic` called `PrintPrettyStack` unconditionally. That function hardcodes `useColor := true`, so the panic stack trace was always printed with ANSI colour codes even when `DefaultLogFormatter.NoColor` was set to `true`.

### Fix

- Add `PrintPrettyStackColor(rvr interface{}, useColor bool)` that accepts an explicit colour flag.
- Refactor `PrintPrettyStack` to delegate to it with `useColor=true` — no change for existing callers.
- Refactor `prettyStack.parse` to receive `useColor` rather than hardcode it.
- `defaultLogEntry.Panic` now passes `l.useColor`, which is derived from `DefaultLogFormatter.NoColor`.

Fixes #1042

### DIFF
--- a/middleware/recoverer.go
+++ b/middleware/recoverer.go
@@ -51,16 +51,14 @@ func Recoverer(next http.Handler) http.Handler {
 // for ability to test the PrintPrettyStack function
 var recovererErrorWriter io.Writer = os.Stderr
 
-// PrintPrettyStack prints a formatted, coloured stack trace to stderr.
 func PrintPrettyStack(rvr interface{}) {
-	PrintPrettyStackColor(rvr, true)
+	printPrettyStack(rvr, true)
 }
 
-// PrintPrettyStackColor prints a formatted stack trace to stderr.
-// When useColor is false ANSI colour codes are suppressed, which is useful
-// for terminals that do not support them (e.g. on Windows) or when output is
-// being captured.
-func PrintPrettyStackColor(rvr interface{}, useColor bool) {
+// printPrettyStack prints a formatted stack trace to stderr. When useColor is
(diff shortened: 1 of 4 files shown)


VojtechVitek:
Needs a rebase.

alliasgher:
Rebased onto master (01faae4). The wrap_writer fix that was in this branch already landed upstream, so this is now just the NoColor panic change.

middleware/recoverer.go
```diff
@@ -51,10 +51,19 @@ func Recoverer(next http.Handler) http.Handler {
...
+// PrintPrettyStack prints a formatted, coloured stack trace to stderr.
 func PrintPrettyStack(rvr interface{}) {
+	PrintPrettyStackColor(rvr, true)
+}
+
+// PrintPrettyStackColor prints a formatted stack trace to stderr.
+// When useColor is false ANSI colour codes are suppressed, which is useful
+// for terminals that do not support them (e.g. on Windows) or when output is
+// being captured.
+func PrintPrettyStackColor(rvr interface{}, useColor bool) {
```
VojtechVitek: Do we really need to export this new function?

alliasgher: No, good catch, only the same-package logger needs it. Unexported it to printPrettyStack in a9f52aa.

VojtechVitek:
Merged via https://github.com/go-chi/chi/pull/1131. Thank you!

PR #1087, opened 2026-04-18
Add trusted CIDR prefix tests for ClientIPFromXFFHeader

Saku0512:
Covers single/multiple trusted proxy chains, spoofing attempts, prefix boundary values, and no-prefix baseline as requested in PR #967.


VojtechVitek:
Thank you for the contribution @Saku0512 👍 I've included your tests as a base for this plan: https://github.com/go-chi/chi/pull/967#issuecomment-4470650467.

PR #1090, opened 2026-04-24
Refactor Chain function to use Middleware type

72sevenzy2:
Add middleware type to slightly enhance readability.

### DIFF
--- a/chain.go
+++ b/chain.go
@@ -2,8 +2,11 @@ package chi
 
 import "net/http"
 
+// middleware type
+type Middleware func(http.Handler) http.Handler
+
 // Chain returns a Middlewares type from a slice of middleware handlers.
-func Chain(middlewares ...func(http.Handler) http.Handler) Middlewares {
+func Chain(middlewares ...Middleware) Middlewares {
 	return Middlewares(middlewares)
 }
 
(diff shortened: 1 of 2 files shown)


JRaspass:
I actually had a similar change locally and was considering PRing it, one fairly important different is I made it a type alias rather than a wholly different type so existing code would keep working, I also defined it just before the `Middlewares` type in chi.go as that seemed a more logical place:
```diff
+// Middlware is a type alias to standard middleware handlers.
+type Middleware = func(http.Handler) http.Handler
+
 // Middlewares type is a slice of standard middleware handlers with methods
 // to compose middleware chains and http.Handler's.
-type Middlewares []func(http.Handler) http.Handler
+type Middlewares []Middleware
```

After that there's a few more places that can be updated to use it like the Router interface, WalkFunc, etc. Just search and replace!

Not sure if such a change is wanted but it definitely gets my vote for being more readable, so long as it's just a type *alias* and therefore won't require explicit conversion from existing callers!

72sevenzy2:
Thats actually smart, I never thought about that, thanks for giving me a heads up.

VojtechVitek:
I was always hoping for Middleware type to be defined in `net/http` stdlib package :)

I don't think chi is the right place to reinvent this, though.

72sevenzy2:
Oh I see.

PR #1093, opened 2026-05-15
middleware: stop double-counting BytesWritten in httpFancyWriter.ReadFrom

### LINKED ISSUE #1067: httpFancyWriter.ReadFrom double-counts BytesWritten when Tee is set

c-tonneslan:
Fixes #1067.

`httpFancyWriter.ReadFrom`'s tee branch ran `io.Copy(&f.basicWriter, r)` and then added `n` to `f.basicWriter.bytes`. But the copy goes through `basicWriter.Write`, which already does `b.bytes += n` at the end. So `BytesWritten()` returned twice the real number of bytes whenever a `Tee` writer was set.

The non-tee branch is unaffected — it uses `ResponseWriter.(io.ReaderFrom).ReadFrom` directly, which doesn't go through `basicWriter.Write`, so it still needs the explicit `+= n`.

Added a regression test exercising the tee path.

### DIFF
--- a/middleware/wrap_writer.go
+++ b/middleware/wrap_writer.go
@@ -208,9 +208,9 @@ func (f *http2FancyWriter) Push(target string, opts *http.PushOptions) error {
 
 func (f *httpFancyWriter) ReadFrom(r io.Reader) (int64, error) {
 	if f.basicWriter.tee != nil {
-		n, err := io.Copy(&f.basicWriter, r)
-		f.basicWriter.bytes += int(n)
-		return n, err
+		// io.Copy routes through basicWriter.Write, which already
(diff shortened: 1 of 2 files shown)


VojtechVitek:
Thank you for the contribution.

This PR looks like a duplicate of https://github.com/go-chi/chi/pull/1085, which was submitted first and merged today.

PR #1094, opened 2026-05-16
middleware: thread useColor through Panic so NoColor logger emits plain text

### LINKED ISSUE #1042: middleware.Logger: defaultLogEntry.Panic() does not respect 'NoColor' of DefaultLogFormatter

c-tonneslan:
Fixes #1042.

\`defaultLogEntry.Panic\` forwarded straight to \`PrintPrettyStack\`, which hard-coded \`useColor=true\`. When the surrounding terminal was a TTY, panic stack traces always came out with ANSI sequences even if the request logs above them honored \`NoColor=true\`. That's especially noisy on a non-color sink (Docker logs, file capture) where the rest of the request logger output is plain.

Added an internal \`printPrettyStackWithColor(rvr, useColor)\` helper. The exported \`PrintPrettyStack\` keeps its default colored behavior (the standalone \`Recoverer\` path doesn't have a logger to inherit a NoColor setting from). \`defaultLogEntry.Panic\` calls the helper with its existing \`useColor\` field, so panic output now matches the rest of the request log.

Added \`TestRequestLoggerPanicRespectsNoColor\` which flips the package-level \`IsTTY\` flag on for the duration of the test (otherwise \`cW\` would skip color emission entirely and the regression wouldn't trigger in CI).

### DIFF
--- a/middleware/logger.go
+++ b/middleware/logger.go
@@ -160,7 +160,7 @@ func (l *defaultLogEntry) Write(status, bytes int, header http.Header, elapsed t
 }
 
 func (l *defaultLogEntry) Panic(v interface{}, stack []byte) {
-	PrintPrettyStack(v)
+	printPrettyStackWithColor(v, l.useColor)
 }
 
 func init() {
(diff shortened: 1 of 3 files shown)


VojtechVitek:
Hi, https://github.com/go-chi/chi/issues/1042 was fixed via https://github.com/go-chi/chi/pull/1050 + https://github.com/go-chi/chi/pull/1131.

Thanks for your contribution.

PR #1097, opened 2026-05-17
fix: set Request.Pattern from RoutePattern()

### LINKED ISSUE #1089: Feature: on Go 1.23+, fill `(*http.Request).Pattern` with RoutePattern data
Similarly to https://github.com/go-chi/chi/issues/873

wyf027:
## Summary

On Go 1.23+, `http.Request.Pattern` is now populated with the full composed route pattern from `RouteContext.RoutePattern()` instead of the per-sub-router `routePattern` field.

## Tests

- Extended `TestPattern` with a nested sub-router case (`/accounts/{accountID}/hi`).

Fixes #1089

### DIFF
--- a/mux.go
+++ b/mux.go
@@ -472,7 +472,7 @@ func (mx *Mux) routeHTTP(w http.ResponseWriter, r *http.Request) {
 			value := rctx.URLParams.Values[i]
 			r.SetPathValue(key, value)
 		}
-		r.Pattern = rctx.routePattern
+		r.Pattern = rctx.RoutePattern()
 
 		h.ServeHTTP(w, r)
 		return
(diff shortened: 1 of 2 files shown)


ptman (approved):

mux.go
```diff
@@ -472,7 +472,7 @@ func (mx *Mux) routeHTTP(w http.ResponseWriter, r *http.Request) {
 			value := rctx.URLParams.Values[i]
 			r.SetPathValue(key, value)
 		}
-		r.Pattern = rctx.routePattern
+		r.Pattern = rctx.RoutePattern()
```
VojtechVitek, line 475: LGTM.

I realized that chi's route pattern might not map 1:1 with stdlib's (regex, * wildcards, no method prefix). But that's not a regression in this PR.

Thank you!

VojtechVitek (approved):

PR #1108, opened 2026-06-12
Add Go fuzz tests for OSS-Fuzz integration

canolgun:
Adds Go native fuzz targets for OSS-Fuzz integration. Chi (22K stars) is a popular Go HTTP router.

### DIFF
--- a/.github/workflows/cifuzz.yml
+++ b/.github/workflows/cifuzz.yml
@@ -6,25 +6,29 @@ on:
       - '.github/workflows/cifuzz.yml'
   push:
     branches: [main, master]
+
 permissions:
   contents: read
+  security-events: write
+
 jobs:
   fuzzing:
     runs-on: ubuntu-latest
     strategy:
       fail-fast: false
       matrix:
-        sanitizer: [address, memory]
+        sanitizer: [address]
     steps:
+    - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
     - name: Build Fuzzers (${{ matrix.sanitizer }})
       id: build
-      uses: google/oss-fuzz/infra/cifuzz/actions/build_fuzzers@ba0e2e0 # v1.0.0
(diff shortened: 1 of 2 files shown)


VojtechVitek (changes requested):
🤖 **Automated review by Claude** (via Claude Code, on behalf of a maintainer)

Thanks for taking the time to put this together — fuzzing chi's path/route parser is a genuinely worthwhile idea, and that's exactly the kind of input-handling code where fuzzing earns its keep. That said, as written this PR has a few issues that I think block merging in its current form.

### 1. The fuzz targets swallow every panic — this defeats the purpose

Both targets wrap the router call in:

```go
defer func() { _ = recover() }()
```

Go's fuzzer detects bugs primarily by catching panics/crashes. If the target recovers from them itself, the fuzzer sees a clean run and reports nothing — a real crash in chi's routing would be silently hidden. These should let panics propagate (or `recover()` only to re-fail via `t.Fatal`), otherwise the fuzzer can run forever and "find" nothing by construction.

### 2. The CIFuzz workflow can't run until chi is onboarded into OSS-Fuzz

`oss-fuzz-project-name: 'chi'` requires a corresponding project (project.yaml, Dockerfile, build.sh) to be merged into the [google/oss-fuzz](https://github.com/google/oss-fuzz) repo first. That's a separate PR to a different repo and isn't part of this change, so `build_fuzzers` will fail here on day one. The workflow shouldn't land before that onboarding exists.

### 3. CI cost / branch mismatch

The workflow runs 5 minutes of fuzzing on **every** PR touching any `.go` file, which is a meaningful ongoing CI burden. It also triggers on `main` while this repo's default branch is `master`.

### 4. `FuzzChiURLParam`'s first block tests nothing

```go
rctx := chi.NewRouteContext()
_ = rctx.URLParam(paramName)
```

A fresh route context has no params, so this always returns empty regardless of input.

### Suggestion

The salvageable core here is the plain `Fuzz*` functions, which are useful standalone via `go test -fuzz`. I'd suggest: drop the blanket `recover()`, drop (or gate) the CIFuzz workflow until OSS-Fuzz onboarding is actually done, and fix the no-op param block. Happy to reconsider once the fuzz targets can actually surface crashes.

Requesting changes for now.

PR #1110, opened 2026-06-13
Honor Discard() in httpFancyWriter.ReadFrom

DucMinhNe:
`httpFancyWriter.ReadFrom` only routes through `basicWriter.Write` when a tee writer is set. When it isn't, it streams straight to the original `ResponseWriter`'s `io.ReaderFrom` — bypassing the `discard` flag entirely.

So after a caller invokes `Discard()` (to suppress the proxied output and write the response itself later), any body sent via `ReadFrom`/`io.Copy` — e.g. `http.ServeContent`, `io.Copy(w, file)`, `http.ServeFile` — **still reaches the client**, defeating the purpose of `Discard()`. The plain `Write` path already respects discard; only the `ReadFrom` fast path missed it.

### Fix

Extend the slow-path condition to also cover discard, so `basicWriter.Write` (which honors both tee and discard) handles those cases. This mirrors the existing tee handling in the same function:

```go
if f.basicWriter.tee != nil || f.basicWriter.discard {
    return n, err
}
```

The fast `rf.ReadFrom` path is now taken only for pure pass-through (no tee, no discard), where it's safe.

### Test

Added `TestHttpFancyWriterReadFromHonorsDiscard`, which writes via `ReadFrom` after `Discard()` and asserts the original writer received nothing while `BytesWritten()` still reflects the consumed bytes. It fails on `master` (original receives all 11 bytes) and passes with the fix. Full `go test ./middleware/` passes; `gofmt`/`go vet` clean.

### DIFF
--- a/middleware/wrap_writer.go
+++ b/middleware/wrap_writer.go
@@ -207,10 +207,12 @@ func (f *http2FancyWriter) Push(target string, opts *http.PushOptions) error {
 }
 
 func (f *httpFancyWriter) ReadFrom(r io.Reader) (int64, error) {
-	if f.basicWriter.tee != nil {
-		// Route through basicWriter.Write so that data is also written to the
-		// tee writer. basicWriter.Write already increments basicWriter.bytes,
-		// so we must NOT add n again here (that would double-count).
+	if f.basicWriter.tee != nil || f.basicWriter.discard {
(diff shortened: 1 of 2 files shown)


VojtechVitek (approved):
LGTM. Thank you

PR #1029, opened 2025-10-27
Don't duplicate methods in Allow: header for 405 responses

flimzy:
Fixes #996

PR #1112, opened 2026-06-21
fix: deduplicate Allow header methods in 405 responses

### LINKED ISSUE #996: Duplicate HTTP methods in 405 Allow header
It seems like when routes are registered which 'overlap' due to their wildcards, their methods are repeated in the Allow header in 405 responses.

Minimal runnable example: https://go.dev/play/p/wm5Roz8MSRJ

```go
package main

import (
	"fmt"
	"log"
	"net/http"
	"net/http/httptest"

	"github.com/go-chi/chi/v5"
)

func main() {
	r := chi.NewRouter()

	r.Post("/article/1-2-3", func(w http.ResponseWriter, r *http.Request) {
		w.WriteHeader(http.StatusOK)

	})

	r.Post("/article/{a}", func(w http.ResponseWriter, r *http.Request) {
		w.WriteHeader(http.StatusOK)

	})

	r.Post("/article/{b}-{c}", func(w http.ResponseWriter, r *http.Request) {
		w.WriteHeader(http.StatusOK)
	})

	r.Post("/article/{b}-{c}-{d}", func(w http.ResponseWriter, r *http.Request) {
		w.WriteHeader(http.StatusOK)
	})

	server := httptest.NewServer(r)
	defer server.Close()

	// Send a request with an unsupported method
	res, err := http.Get(server.URL + "/article/1-2-3")
	if err != nil {
		log.Fatal(err)
	}

	fmt.Println("Status:", res.Status)
	fmt.Println("Header:", res.Header)
}
```

Outputs:

```
Status: 405 Method Not Allowed
Header: map[Allow:[POST POST POST POST] Content-Length:[0] Date:[Tue, 10 Nov 2009 23:00:00 GMT]]
```

The `POST` value in the `Allow` header is duplicated four times, which seems strange, and I think ideally it would only appear once.

LeSingh1:
## Problem

When multiple overlapping parameterized route patterns all register the same HTTP method, a request with an unregistered method produces duplicate entries in the `Allow` response header.

**Minimal reproduction** (from #996):

```go
r.Post("/article/1-2-3", ...)
r.Post("/article/{a}", ...)
r.Post("/article/{b}-{c}", ...)
r.Post("/article/{b}-{c}-{d}", ...)

// GET /article/1-2-3 → 405 with Allow: POST, POST, POST, POST
```

Each parameterized node that matched during route traversal independently appended `POST` to `rctx.methodsAllowed []methodTyp`. The `methodNotAllowedHandler` then wrote one `Allow` header value per slice entry — one per matched node — instead of one per distinct method.

## Fix

Replace `methodsAllowed []methodTyp` in `Context` with a `methodTyp` bitmask. Since `methodTyp` is already defined as a bitmask of power-of-two constants, each candidate node can OR its methods into the bitmask (`rctx.methodsAllowed |= endpoints`) rather than append to a slice. `methodNotAllowedHandler` then iterates `reverseMethodMap` and emits one `Allow` header entry for each bit that is set — guaranteeing uniqueness without any extra allocation.

## Changes

- `context.go`: `methodsAllowed []methodTyp` → `methodsAllowed methodTyp`; `Reset()` zeroes with `= 0`
- `tree.go`: two `append(rctx.methodsAllowed, ...)` → `rctx.methodsAllowed |= endpoints`
- `mux.go`: `MethodNotAllowedHandler` and `methodNotAllowedHandler` take a single `methodTyp` bitmask instead of `...methodTyp`; handler iterates `reverseMethodMap` by bitmask
- `mux_test.go`: `TestMethodNotAllowedNoDuplicates` — regression test from the exact scenario in #996

## Test plan

- [x] `TestMethodNotAllowedNoDuplicates` — asserts no duplicate values in `Allow` header and that `POST` is present
- [x] Existing `TestMethodNotAllowed` still passes (2 distinct methods, GET and HEAD, each appear once)
- [x] `go test ./...`

Fixes #996

### DIFF
--- a/context.go
+++ b/context.go
@@ -74,7 +74,7 @@ type Context struct {
 	// patterns across a stack of sub-routers.
 	RoutePatterns []string
 
-	methodsAllowed   []methodTyp // allowed methods in case of a 405
+	methodsAllowed   methodTyp // allowed methods bitmask in case of a 405
 	methodNotAllowed bool
 }
 
@@ -91,7 +91,7 @@ func (x *Context) Reset() {
 	x.routeParams.Keys = x.routeParams.Keys[:0]
 	x.routeParams.Values = x.routeParams.Values[:0]
 	x.methodNotAllowed = false
-	x.methodsAllowed = x.methodsAllowed[:0]
+	x.methodsAllowed = 0
 	x.parentCtx = nil
 }
 
--- a/mux.go
+++ b/mux.go
@@ -410,11 +410,11 @@ func (mx *Mux) NotFoundHandler() http.HandlerFunc {
 
 // MethodNotAllowedHandler returns the default Mux 405 responder whenever
 // a method cannot be resolved for a route.
-func (mx *Mux) MethodNotAllowedHandler(methodsAllowed ...methodTyp) http.HandlerFunc {
+func (mx *Mux) MethodNotAllowedHandler(methodsAllowed methodTyp) http.HandlerFunc {
 	if mx.methodNotAllowedHandler != nil {
 		return mx.methodNotAllowedHandler
 	}
-	return methodNotAllowedHandler(methodsAllowed...)
+	return methodNotAllowedHandler(methodsAllowed)
 }
 
 // handle registers a http.Handler in the routing tree for a particular http method
@@ -467,7 +467,7 @@ func (mx *Mux) routeHTTP(w http.ResponseWriter, r *http.Request) {
 	}
 	method, ok := methodMap[rctx.RouteMethod]
 	if !ok {
-		mx.MethodNotAllowedHandler().ServeHTTP(w, r)
+		mx.MethodNotAllowedHandler(0).ServeHTTP(w, r)
 		return
 	}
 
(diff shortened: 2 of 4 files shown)


VojtechVitek:
Thanks everyone. I confirmed the issue and reviewed these competing PRs:

| PR | Age | Approach |
|---|---|---|
| [#1029](https://github.com/go-chi/chi/pull/1029) | Oct 2025 | Dedupe at the source in `tree.go` (`slices.Contains` before append) |
| [#1040](https://github.com/go-chi/chi/pull/1040), [#1096](https://github.com/go-chi/chi/pull/1096), [#1124](https://github.com/go-chi/chi/pull/1124) | Dec 2025 – Jul 2026 | Near-identical: seen-map inside `methodNotAllowedHandler` |
| [#1047](https://github.com/go-chi/chi/pull/1047) | Feb 2026 | `slices.Sort` + `Compact` in the handler |
| [#1112](https://github.com/go-chi/chi/pull/1112) | Jun 2026 | Refactor `methodsAllowed` to a bitmask |

Let's go with [#1029](https://github.com/go-chi/chi/pull/1029), which looks like the cleanest zero-allocation fix.

Closing the other PRs as duplicates — thanks for the contributions, the approaches were all reasonable.

Thank you for your contribution!

PR #1113, opened 2026-06-22
Tidy build directives

JRaspass:
Remove the obsolete "// +build" form which hasn't been needed since Go 1.18, and remove the Go 1.23 build directives as that matches our minimum version now.


VojtechVitek (approved):
LGTM

PR #1118, opened 2026-06-29
middleware: handle Host routing via Request.Host

### LINKED ISSUE #691: Subdomain example in route_headers.go not working
I implemented the subdomain example provided in the [route_headers.go file](https://github.com/go-chi/chi/blob/master/middleware/route_headers.go). 
Unfortunately the example provided is not working. The issue is that sub.example.com goes to the main domain page instead of sub domain page. 
Can anyone please let me know how to fix the issue ? Or if the below code is working for them ?

I was testing on my local Ubuntu PC and following are the contents of /etc/hosts
_$ cat /etc/hosts
127.0.0.1       localhost
127.0.0.1       example.com
127.0.0.1       sub.example.com_

```
package main

import (
	"fmt"
	"net/http"

	"github.com/go-chi/chi/v5"
)


func main() {
    r := chi.NewRouter()
    rSubdomain := chi.NewRouter()

    r.Use(middleware.RouteHeaders().
        Route("Host", "example.com", middleware.New(r)).
        Route("Host", "*.example.com", middleware.New(rSubdomain)).
        Handler)

    r.Get("/", h)
    rSubdomain.Get("/", h2)

    http.ListenAndServe(":8080", r)
}

	fmt.Fprintf(w, "Main domain")
}

func h2(w http.ResponseWriter, r *http.Request) {
	fmt.Fprintf(w, "Sub domiain")
}
```

happysnaker:
## Summary
- read `Request.Host` for `RouteHeaders` rules targeting `Host`
- document the `net/http` host promotion behavior in the middleware docs
- add coverage for host matching with and without an explicit port

## Why
`net/http` promotes the HTTP Host header to `Request.Host` and removes it from `Request.Header`, so the existing Host-based example in `RouteHeaders` is misleading for normal browser traffic. This patch makes Host routing behave the way users expect and updates the docs/tests accordingly.

Closes #691

### DIFF
--- a/middleware/route_headers.go
+++ b/middleware/route_headers.go
@@ -1,6 +1,7 @@
 package middleware
 
 import (
+	"net"
 	"net/http"
 	"strings"
 )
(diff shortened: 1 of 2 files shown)


happysnaker:
Friendly follow-up: this patch keeps the API surface unchanged, but makes Host-based routing line up with how `net/http` actually exposes host information (`Request.Host` instead of `Request.Header["Host"]`).

It also adds host-with-port coverage, which seems useful for local development cases like `example.com:8080`.

happysnaker:
Resubmitted from a restored public fork as #1120. Closing this thread in favor of the new PR so the contribution stays reviewable from a proper fork relationship.

PR #1120, opened 2026-06-30
middleware: handle Host routing via Request.Host

### LINKED ISSUE #691: Subdomain example in route_headers.go not working

happysnaker:
## Summary
- read `Request.Host` for `RouteHeaders` rules targeting `Host`
- document the `net/http` host promotion behavior in the middleware docs
- add coverage for host matching with and without an explicit port

## Why
`net/http` promotes the HTTP Host header to `Request.Host` and removes it from `Request.Header`, so the existing Host-based example in `RouteHeaders` is misleading for normal browser traffic. This patch makes Host routing behave the way users expect and updates the docs/tests accordingly.

Closes #691

## Note
This PR supersedes #1118, which I had to close after the original fork relationship was broken during repo visibility changes on my side. The code/content is the same contribution path, now resubmitted from a proper public fork.

### DIFF
--- a/middleware/route_headers.go
+++ b/middleware/route_headers.go
@@ -1,6 +1,7 @@
 package middleware
 
 import (
+	"net"
 	"net/http"
 	"strings"
 )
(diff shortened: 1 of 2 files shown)


happysnaker:
Maintainer note: this patch intentionally targets the existing documented `Host` use case in `RouteHeaders` and keeps the change minimal by reading `Request.Host`, which is where Go promotes the HTTP Host header for normal server requests. Tests cover host matching with and without a port so behavior stays explicit.

happysnaker:
Keeping this warm: the intent of this patch is still to make the documented `RouteHeaders(...).Route("Host", ...)` usage work with normal Go server requests by reading `Request.Host`, including the common `host:port` case. Tests cover exact-host and wildcard-host matching with and without a port, and the change stays local to `middleware/route_headers.go`.

PR #1121, opened 2026-07-01
fix(middleware): honor all-types compress wildcard

### LINKED ISSUE #868: Brotli compression not working
```
import brotli_enc "gopkg.in/kothar/brotli-go.v0/enc"

compressor := middleware.NewCompressor(5, "text/html")
compressor.SetEncoder("br", func(w http.ResponseWriter, level int) io.Writer {
  params := brotli_enc.NewBrotliParams()
  params.SetQuality(level)
  return brotli_enc.NewBrotliWriter(params, w)
})
```

the above code shows error, but following seems working but compression not working what's wrong?

```
	r := chi.NewRouter() //chi.NewMux()
	//r.Use(middleware.Logger)

	compressor := middleware.NewCompressor(5, "/*") //"/*"
	compressor.SetEncoder("br", func(w io.Writer, level int) io.Writer {
		params := brotli_enc.NewBrotliParams()
		params.SetQuality(level)
		return brotli_enc.NewBrotliWriter(params, w)
	})
	r.Use(compressor.Handler) //
```

happysnaker:
## Summary
- make `NewCompressor(..., "/*")` behave as an actual all-types wildcard
- keep the existing wildcard parsing model and fix only the matching side
- add regression coverage for both `application/*` and `text/*` responses

## Problem
`NewCompressor(5, "/*")` is currently accepted as a valid wildcard pattern, but it silently does not work.

During setup, the pattern is normalized into an empty wildcard key via `strings.CutSuffix("/*", "/*")`, which is fine as an internal sentinel. The problem is in `isCompressible()`: wildcard matching only checks the top-level media type prefix (for example `text` or `application`) and never treats the empty key as a match-all sentinel.

That means the following configuration looks valid but compresses nothing:

```go
middleware.NewCompressor(5, "/*")
```

## Changes
- treat an empty wildcard key as "match any top-level media type" during content-type checks
- add a focused regression test showing that `"/*"` now compresses both `application/json` and `text/html; charset=utf-8`

The patch is intentionally narrow: it does not change the wildcard syntax rules, only makes the already-accepted `"/*"` form behave consistently.

Closes #868

### DIFF
--- a/middleware/compress.go
+++ b/middleware/compress.go
@@ -281,6 +281,9 @@ func (cw *compressResponseWriter) isCompressible() bool {
 		return true
 	}
 	if contentType, _, hadSlash := strings.Cut(contentType, "/"); hadSlash {
+		if _, ok := cw.contentWildcards[""]; ok {
+			return true
+		}
 		_, ok := cw.contentWildcards[contentType]
 		return ok
 	}
--- a/middleware/compress_test.go
+++ b/middleware/compress_test.go
@@ -169,6 +169,53 @@ func TestCompressorWildcards(t *testing.T) {
 	}
 }
 
+func TestCompressorWildcardAllTypes(t *testing.T) {
+	r := chi.NewRouter()
+	r.Use(NewCompressor(5, "/*").Handler)
+
+	r.Get("/json", func(w http.ResponseWriter, r *http.Request) {
+		w.Header().Set("Content-Type", "application/json")
+		w.Write([]byte(`{"ok":true}`))
+	})
+
+	r.Get("/html", func(w http.ResponseWriter, r *http.Request) {
+		w.Header().Set("Content-Type", "text/html; charset=utf-8")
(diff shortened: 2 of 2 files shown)


happysnaker:
Maintainer context:

- this keeps the existing wildcard syntax exactly as-is and only fixes the already-accepted `"/*"` form on the matching side
- `NewCompressor(..., "/*")` already normalizes into an empty wildcard key today, so the narrow intent here is just to make that sentinel behave consistently instead of silently matching nothing
- I kept the regression coverage focused on real response types (`application/json` and `text/html; charset=utf-8`) to show the intended runtime behavior without broadening the wildcard model further

VojtechVitek:
Hi, thanks for taking a pass at the fix.

However, I don't think we should allow the catchall `/*` here. It would allow users to misuse the middleware and compress already-compressed content like zip, jpeg, or png.

I'll follow-up with a PR that will disallow `/*`.

PR #1129, opened 2026-07-03
feat(mux): support http QUERY method ietf rfc10008

pjol:
Adds support for [IETF RFC 10008](https://datatracker.ietf.org/doc/rfc10008/) - The HTTP QUERY Method

As support is not yet enabled in native GoLang net/http, we have introduced a temporary method definition directly into the project [here](https://github.com/go-chi/chi/compare/master...pjol:chi:master?expand=1#diff-050830ac4334170e8d335e1b95519052a3461c5b57a4ca38a0531e35a8f3d388R39). Once enabled in net/http, a clean 1-1 swap with our temporary method should be simple.

### DIFF
--- a/README.md
+++ b/README.md
@@ -221,6 +221,7 @@ type Router interface {
 	Patch(pattern string, h http.HandlerFunc)
 	Post(pattern string, h http.HandlerFunc)
 	Put(pattern string, h http.HandlerFunc)
+	Query(pattern string, h http.HandlerFunc)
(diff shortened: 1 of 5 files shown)


VojtechVitek (approved):
LGTM. Thank you!

I'm going to unexport the constant and merge via https://github.com/go-chi/chi/pull/1132

VojtechVitek:
Merged via https://github.com/go-chi/chi/pull/1132

PR #1061, opened 2026-03-14
respect NoColor setting in defaultLogEntry.Panic()

ImpulseB23:
Fixes #1042.

`defaultLogEntry.Panic()` called `PrintPrettyStack()` which hardcodes `useColor = true`, ignoring the formatter's NoColor setting.

PR #1041, opened 2025-12-31
respect defaultLogEntry.useColor when call it's Panic(...)

wsw0108:
This PR fix #1042

PR #1131, opened 2026-07-05
middleware: document printPrettyStack and harden NoColor panic test

### LINKED ISSUE #1042: middleware.Logger: defaultLogEntry.Panic() does not respect 'NoColor' of DefaultLogFormatter

VojtechVitek:
Follow-up to #1050, which fixed the core regression from #1042 (`defaultLogEntry.Panic()` ignored `DefaultLogFormatter.NoColor` because `PrintPrettyStack` hardcodes `useColor = true`).

This carries over the two remaining pieces from the other PRs on that issue, on top of the merged fix:

- **Doc comment** on the unexported `printPrettyStack`, explaining the `useColor` behavior (from #1086).
- **Hardened `TestRecovererNoColor`**: force `IsTTY = true` (saved/restored) so `cW` would emit ANSI codes if `useColor` were true. Without this the assertion is a no-op on non-TTY runners (e.g. CI) and passes even against the buggy code path — verified it fails when the bug is reintroduced. (from #1061)

It preserves @alliasgher's original #1086 commits via a merge commit so authorship is retained.

Supersedes #1041, #1061, #1086.

Fixes #1042.

### DIFF
--- a/middleware/logger.go
+++ b/middleware/logger.go
@@ -166,7 +166,7 @@ func (l *defaultLogEntry) Write(status, bytes int, header http.Header, elapsed t
 }
 
 func (l *defaultLogEntry) Panic(v interface{}, stack []byte) {
-	PrintPrettyStackColor(v, l.useColor)
+	printPrettyStack(v, l.useColor)
 }
 
 func init() {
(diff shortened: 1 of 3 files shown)


VojtechVitek (commented):
LGTM.

Thank you @alliasgher for https://github.com/go-chi/chi/pull/1086.

PR #1137, opened 2026-07-22
docs(middleware): fix xml case label in URLFormat example

sonnemusk:
Godoc used `case "xml:"` which never matches format `xml`.


sonnemusk:
Closing — pure docs/typo cleanup. Going to focus on behavioral fixes instead.

PR #1138, opened 2026-07-22
fix(middleware): parse comma-separated Content-Encoding values

sonnemusk:
Split Content-Encoding on commas; fix multi-value tests.


sonnemusk:
Closing after a second quality pass: too thin / low user impact. Prefer fewer high-signal fixes. Sorry for the noise.

PR #1139, opened 2026-07-22
fix(middleware): accept quoted charset in Content-Type

sonnemusk:
Strip quotes from charset= so charset="utf-8" is accepted.


sonnemusk:
Closing after a second quality pass: too thin / low user impact. Prefer fewer high-signal fixes. Sorry for the noise.

Issue #750, opened 2022-09-23
chi.Walk missing middlewares


ebabani:
Hi,

I was trying to use chi.Walk to get a report of all routes and the middlewares used by every route. I noticed that some of the middlewares were missing from some of the nested routes when using groups.

Example routes:

```
	r := chi.NewRouter()

	r.Use(middleware.RequestID)

	r.Get("/A", func(w http.ResponseWriter, r *http.Request) {})

	r.Group(func(r chi.Router) {
		r.Use(middleware.Timeout(2500 * time.Millisecond))

		r.Get("/B", func(w http.ResponseWriter, r *http.Request) {})

		r.Route("/C", func(r chi.Router) {

			// Walk on the below route does not show the Timeout middleware
			r.Get("/D", func(w http.ResponseWriter, r *http.Request) {})
		})
	})
```

In the example above, I expected the `C/D` route to have both RequestId and Timeout middlewares. but chi.Walk only shows RequestID.

It's possible I'm missing something here and this is behaving as expected.

Go playground link: https://go.dev/play/p/HKh7cA35Bgx

VojtechVitek:
I guess the issue is that `r.Route()` creates a new sub-router (unlike `r.Group()` or `r.Handle()`) and `Walk()` doesn't reach out to its middleware stack.

https://github.com/go-chi/chi/blob/v0.9.0/mux.go#L169

ebabani:
It does seem related to the `Group` to me. I updated the example with some nested routes, and for those it shows the correct middleware chain.

https://go.dev/play/p/fia_kU8Lzqs

pkieltyka:
Group() is used for inlining middleware, where as Rotue() mount a subrouter -- it's a subtle difference, but there is one. The middlewares however will be there along the C/D route.

If I recall -- the inline middleware stack is computed and reduced in memory, and we don't keep around in memory the middleware stack, which is why we can't walk it. We could definitely add it, it just means a bit more code. It would be safe to keep, but would only be useful for walking like in this case.

ganicus:
> In terms of functionality all is good, it was just surprising behaviour when I hit this.
> 
> For context, I am using the walk function to write tests agains the router we're using. I was writing checks to ensure certain endpoints were always behind an auth middleware. I'm not super familiars with the chi internals, but if you can point me in the general area I can take a look and create a PR.

@ebabani what did you end up doing for testing with Route() and Group()?

ebabani:
Hey @ganicus,

[...]

PR #1146, opened 2026-08-02
Fix Walk for routes with handlers and subrouters

### LINKED ISSUE #830: chi.Walk missing routes
Similarly to https://github.com/go-chi/chi/issues/750, I'm trying to use chi.Walk() to get a report of all routes. However, I found that some routes are missing when I use Route() as well as e.g. Get() with the same pattern: https://go.dev/play/p/7Ntz1yMoXrz

Interestingly, the route _is_ visited by the walk function when you change `r.Route("/bar", ...` to `r.Route("/bar/", ...`.

Since the `GET /foo/bar` request is handled in both cases, I also expect chi.Walk() to report the route in both cases.

deepakganesh78:
Fixes #830

## Reproduction
Register a handler and a subrouter on the same Route() pattern:

```go
r.Route("/foo", func(r chi.Router) {
    r.Route("/bar", func(r chi.Router) {
        r.Get("/{id}", handler)
    })
    r.Get("/bar", handler)
})
```

`GET /foo/bar` is routable, but `chi.Walk` only reported `GET /foo/bar/{id}` and omitted `GET /foo/bar`.

## Root cause
`walk` returned immediately after descending into `route.SubRoutes`, so any real handlers stored on that same route entry were skipped. The route tree also uses stub endpoints for mount connectors; those synthetic handlers need to remain hidden from public route listings.

## Fix
Continue walking handlers after subroutes, while keeping subrouter middleware propagation separate. `routes()` now filters synthetic stub handlers but preserves the internal `*` handler used to carry inline middleware into subrouters.

## Compatibility
No API changes and no new dependencies. This only makes `Walk`/`Routes` report an already-routable handler that was previously omitted; synthetic mount connector routes remain hidden.

## Validation
- Regression test `go test . -run TestWalkRouteWithHandlerAndSubrouter -count=1` fails without the fix: `expected Walk to include GET /foo/bar, got map[GET /foo/bar/{id}:true]`.
- `go build ./...` passed.
- `go vet ./...` passed.
- `gofmt -l .` produced no output.
- `go test ./...` passed (`github.com/go-chi/chi/v5`, `github.com/go-chi/chi/v5/middleware`).

### DIFF
--- a/tree.go
+++ b/tree.go
@@ -7,6 +7,7 @@ package chi
 import (
 	"fmt"
 	"net/http"
+	"reflect"
 	"regexp"
 	"slices"
 	"sort"
@@ -629,8 +630,9 @@ func (n *node) routes() []Route {
 	rts := []Route{}
 
 	n.walk(func(eps endpoints, subroutes Routes) bool {
-		if eps[mSTUB] != nil && eps[mSTUB].handler != nil && subroutes == nil {
-			return false
+		var stubHandler http.Handler
+		if eps[mSTUB] != nil {
+			stubHandler = eps[mSTUB].handler
 		}
 
 		// Group methodHandlers by unique patterns
@@ -650,12 +652,13 @@ func (n *node) routes() []Route {
 
 		for p, mh := range pats {
 			hs := make(map[string]http.Handler)
+
 			if mh[mALL] != nil && mh[mALL].handler != nil {
 				hs["*"] = mh[mALL].handler
 			}
 
 			for mt, h := range mh {
-				if h.handler == nil {
+				if h.handler == nil || sameHandler(h.handler, stubHandler) {
 					continue
 				}
 				if m, ok := reverseMethodMap[mt]; ok {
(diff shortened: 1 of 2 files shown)


VojtechVitek (changes requested):
Thanks for digging into #830, @deepakganesh78 — the stub-vs-real-handler diagnosis is right, and `sameHandler` avoids a real footgun (`http.HandlerFunc` isn't `==`-comparable).

Two gaps, found by checking `Routes()` directly rather than just `Walk`:

1. `hs["*"] = mh[mALL].handler` isn't run through the `sameHandler` check just below it, so every `Mount()`/`Route()` now leaks two extra `Route` entries (the exact pattern + its trailing slash) pointing at the internal mount handler. Invisible via `chi.Walk` only because `walk()` always skips `"*"` — but visible to anything calling `Routes()` directly.
2. The `walk()` change (dropping `continue`, forking `subrouteMw`) looks unnecessary — reverting just that half still passes your test and the rest of the suite.

Opened #1148 with both fixed, keeping your test and the `sameHandler` approach. Its description also credits everyone who touched #830/#750 before this, including the folks who flagged this earlier: @sauerbraten, @ganicus, @lai0xn on #830, and @ebabani, @pkieltyka on #750.

VojtechVitek:
Superseded by https://github.com/go-chi/chi/pull/1148.

Thank you for your contribution!

PR #1148, opened 2026-08-09
fix: don't drop handlers that collide with a Mount()/Route() pattern

### LINKED ISSUE #830: chi.Walk missing routes

VojtechVitek:
Fixes #830: a handler sharing a pattern with a `Route()`/`Mount()` call was invisible to `Walk()`/`Routes()`, because the mount's internal stub handler caused the whole node to be hidden.

```go
r.Route("/foo", func(r chi.Router) {
    r.Route("/bar", func(r chi.Router) { r.Get("/{id}", h) })
    r.Get("/bar", h) // routable, but invisible to Walk
})
```

Builds on #1146 (thanks @deepakganesh78 for the diagnosis, repro, and the `sameHandler` idea), with two corrections found by testing `Routes()` directly instead of only `Walk`:

- Only hides the stub's `"*"` slot on leaf nodes — subroute nodes still need it, since `walk()` reads `Handlers["*"]` to pick up `With()` middleware from a `*ChainHandler`. #1146 hides it everywhere, which leaks two extra `Route` entries per `Mount()`/`Route()` call.
- Drops #1146's unrelated `walk()` change (`continue`/`subrouteMw`) — reverting just that half still passes everything.

Also adds a regression test for #750 (middlewares on a `Group()` not reaching a `Route()` mounted inside it) — already fixed on master, this just locks it in.

**Credits** — reported, diagnosed, or attempted a fix on these two issues:
- #830: [@sauerbraten](https://github.com/sauerbraten) (reported), [@ganicus](https://github.com/ganicus), [@lai0xn](https://github.com/lai0xn), [@deepakganesh78](https://github.com/deepakganesh78) (#1146)
- #750: [@ebabani](https://github.com/ebabani) (reported), [@pkieltyka](https://github.com/pkieltyka), [@ganicus](https://github.com/ganicus)

**Testing**: `TestWalkRouteWithHandlerAndSubrouter`, `TestRoutesHidesMountStub`, `TestWalkMiddlewaresAcrossGroupAndRoute`; `go build`/`vet`/`gofmt -l`/`test ./...` all clean.

### DIFF
--- a/tree_test.go
+++ b/tree_test.go
@@ -622,13 +622,9 @@ func TestRoutesHidesMountStub(t *testing.T) {
 			r.Get("/{id}", handler)
 		})
 
-		// Routes() is not recursive: r.Route("/api", ...) mounts a distinct
-		// subrouter, so r.Routes() sees only the "/api/*" connector to it, not
-		// "/api/{id}". SubRoutes is itself a Routes, so it can be walked
-		// further — either manually via SubRoutes.Routes(), or by passing it
-		// straight to Walk() to recurse into just that subtree. Walk() itself
-		// already does this recursion across the whole tree, which is how it
-		// correctly reports "/api/{id}" (see TestWalkRouteWithHandlerAndSubrouter).
+		// Routes() isn't recursive: /api mounts a separate subrouter, so only
+		// its "/api/*" connector shows here, not "/api/{id}". SubRoutes can be
+		// walked further, the same way Walk() does internally.
 		routes := r.Routes()
 		if len(routes) != 1 {
 			t.Fatalf("expected exactly 1 route, got %d: %+v", len(routes), routes)
--- a/tree_test.go
+++ b/tree_test.go
@@ -622,6 +622,13 @@ func TestRoutesHidesMountStub(t *testing.T) {
 			r.Get("/{id}", handler)
 		})
 
+		// Routes() is not recursive: r.Route("/api", ...) mounts a distinct
+		// subrouter, so r.Routes() sees only the "/api/*" connector to it, not
+		// "/api/{id}". SubRoutes is itself a Routes, so it can be walked
+		// further — either manually via SubRoutes.Routes(), or by passing it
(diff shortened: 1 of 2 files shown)


tree_test.go
```diff
@@ -577,3 +579,137 @@ func TestWalkInlineMiddlewaresAcrossSubrouter(t *testing.T) {
...
+		// Routes() isn't recursive: /api mounts a separate subrouter, so only
+		// its "/api/*" connector shows here, not "/api/{id}". SubRoutes can be
+		// walked further, the same way Walk() does internally.
+		routes := r.Routes()
+		if len(routes) != 1 {
+			t.Fatalf("expected exactly 1 route, got %d: %+v", len(routes), routes)
+		}
+		if routes[0].Pattern != "/api/*" || routes[0].SubRoutes == nil {
+			t.Fatalf("expected the single route to be the /api/* subroute, got %+v", routes[0])
+		}
```
VojtechVitek, line 634: FYI -- I was surprised to learn that chi.Walk() doesn't walk the sub-routes automatically

Follow-up: https://github.com/go-chi/chi/issues/1149

VojtechVitek (commented):
LGTM

tree_test.go
VojtechVitek: Never mind. This test is not testing `.Walk()` (recursive) .. it's only testing `.Routes()` (non-recursive).

This is by design. It's not an issue. Going to close #1149.

PR #1150, opened 2026-08-10
test: make throttle retry-after tests deterministic and cover the cancel path

adisetiawanx:
Following up on #1016, this tightens up the throttle retry-after tests in two ways.

**The existing test is slow and timing dependent.** It sleeps a full second inside the handler for every request, then spins up a real httptest server and talks to it over the network. It passes, but running it repeatedly to shake out flakes takes a while, and the outcome can shift with how the runtime schedules goroutines.

**It only covers one of the two retryAfterFn paths.** The middleware calls RetryAfterFn with `ctxDone=true` when a request sitting in the backlog gets its context cancelled, and with `ctxDone=false` when the limit is just hit. That cancel path was never asserted.

What changed:

- Rewrote the first test to hold the limit tokens open with a channel instead of sleeping. Requests beyond the limit are rejected immediately, so the assertions stay the same no matter how things are scheduled.
- Added `TestThrottleRetryAfterCancelled`, which lets a backlogged request's context be cancelled and checks that the cancel variant of Retry-After is returned.

On the rewrite: `go test -count=1000 -run TestThrottleRetryAfter ./middleware/` completes in under a second. The old version took roughly a second per iteration.

### DIFF
--- a/middleware/throttle_test.go
+++ b/middleware/throttle_test.go
@@ -1,6 +1,7 @@
 package middleware
 
 import (
+	"context"
 	"io"
 	"net/http"
 	"net/http/httptest"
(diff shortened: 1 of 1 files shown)


VojtechVitek:
Thanks for the PR — the channel-based approach is the right way to test this, and covering the `ctxDone=true` path is a genuinely useful addition, that path had zero coverage until now.

That said, to be fair: the flakiness motivation is a bit overstated. #1016 fixed the actual flake a year ago and we haven't seen a failure since. The speed-up is nice, but what makes this PR worth merging is the new coverage — and right now the rewrite actually *drops* two assertions:

1. **The `Retry-After: 3600` header check is gone.** The rewritten test only collects `rec.Code`, so after this PR the `ctxDone=false` header value is asserted nowhere. A test named `TestThrottleRetryAfter` has to assert Retry-After. Please pass the recorder through the channel and check the header.
2. **The 200s are no longer verified.** The old test asserted the 5 accepted requests succeed; the new one discards their recorders. Two lines to restore.

One more note, no change needed: in `TestThrottleRetryAfterCancelled`, `cancel()` can fire before the second request even enters the middleware, so either cancel branch (early or backlogged) may run. Both return `ctxDone=true` so the assertion holds either way — a short comment acknowledging this would help the next reader.

Optional, if you want to make the coverage complete: the backlog-timeout path (`ctxDone=false` via `BacklogTimeout`) is untested in both the old and new version. A third small test with a short timeout would close that gap.

Fix 1 and 2 and this is good to go 👍

VojtechVitek:
Requesting review from @OtavioBernardes, as the author of https://github.com/go-chi/chi/pull/1016.

adisetiawanx:
Thanks, good catches. Both are in now.

The rewrite really did drop the Retry-After assertion and never checked that the held requests came back 200. Both are asserted again now, the recorders are collected over channels so the assertions run after the throttled phase ends.

For the cancelled test I added the comment about the two possible cancel paths so the next reader does not have to wonder.

Also added a small test for the backlog timeout branch (ctxDone=false) since neither version covered it.

PR #1127, opened 2026-07-03
feat(middleware): add text/xml and application/xml to default compressible types

VojtechVitek:
## Summary

Supersedes #1126 with the test fix applied.

XML content compresses very effectively (typically 70-90% reduction) and is commonly served by web applications, APIs, and RSS/Atom feeds. This adds `application/xml` and `text/xml` to the default compressible content types list.

Note that `application/atom+xml` and `application/rss+xml` were already present — but generic XML was not.

## Changes

- Added `application/xml` and `text/xml` to `defaultCompressibleContentTypes` (original commit by @happysnaker from #1126)
- Updated `TestCompressorWildcards/defaults` expected `typesCount` from 10 to 12 to match the new defaults — this was the failing assertion in #1126

Fixes #920. Closes #1126.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1151, opened 2026-08-15
feat(middleware): add text/markdown, text/csv, text/vtt to default compressible types

VojtechVitek:
Extends `defaultCompressibleContentTypes` with three common text types:

- `text/markdown`
- `text/csv`
- `text/vtt`

This is the explicit-whitelist alternative to #1057, which proposed a `text/*` wildcard. Keeping exact entries preserves the single-map-lookup hot path in `isCompressible()` and avoids silently compressing `text/event-stream` (SSE) by default. Users who want blanket coverage can still opt in with `Compress(5, "text/*")`.

Follows the same pattern as #1127.


VojtechVitek (commented):
LGTM

PR #1157, opened 2026-08-20
Add Clone method to Context

VojtechVitek:
This includes #1033 by @BlakeWilliams as-is (rebased onto current master, commit authorship preserved), plus a follow-up commit applying the review feedback:

- `Clone()` detaches `parentCtx`, so the clone doesn't pin the finished request's context.
- Doc comment spells out that `Clone()` must be called before the handler returns.
- `Context` struct notes that new reference fields must be deep-copied in `Clone()`.

Credit to @BlakeWilliams for the implementation and the design discussion in #1028.

Closes #1028
Closes #1033

### DIFF
--- a/context.go
+++ b/context.go
@@ -95,6 +95,22 @@ func (x *Context) Reset() {
 	x.parentCtx = nil
 }
 
+// Clone a routing context so that it may be used outside of the request/response lifecycle.
+func (c *Context) Clone() *Context {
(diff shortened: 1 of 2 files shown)


VojtechVitek (commented):
LGTM.

Thanks @BlakeWilliams for https://github.com/go-chi/chi/pull/1033.

PR #942, opened 2024-08-09
Fix ContentCharset middleware to skip processing if ContentLength == 0 

noazaj:
This PR fixes issue #746 by updating the ContentCharset middleware to skip processing if ContentLength is 0.

PR #1158, opened 2026-08-20
middleware: ContentCharset should skip requests without a body

### LINKED ISSUE #746: ContentCharset middleware should be skipped if ContentLength == 0
I could be wrong about this but I feel like the `ContentCharset` middleware should skip processing if the `ContentLength == 0`, just like the `AllowContentType` and `AllowContentEncoding` middleware. I also feel like it for consistency it should be `AllowContentCharset`? Some of the functions names in the `content_charset.go` file are confusing as well as they are named `contentEncoding`, etc.

I can create a PR if you want.

VojtechVitek:
Cherry-picked from #942 by @noazaj (fixes #746), plus a fix to its tests.

`ContentCharset` used to 415 any request without a body — e.g. every GET — unless it carried a matching charset header. Now it passes bodyless requests (`ContentLength == 0`) through, consistent with `AllowContentType` and `AllowContentEncoding`. Chunked requests (`ContentLength == -1`) are still validated.

The original PR's tests didn't pass: the `contentLength` table field was never applied to the request, so every case ran with `ContentLength == 0` and the three 415 cases failed. Tests now drive `ContentLength` from a real request body, and cover the bodyless request with no `Content-Type` header — the exact scenario from #746.

### DIFF
--- a/middleware/content_charset.go
+++ b/middleware/content_charset.go
@@ -15,6 +15,11 @@ func ContentCharset(charsets ...string) func(next http.Handler) http.Handler {
 
 	return func(next http.Handler) http.Handler {
 		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			if r.ContentLength == 0 {
+				next.ServeHTTP(w, r)
+				return
+			}
+
 			if !contentEncoding(r.Header.Get("Content-Type"), charsets...) {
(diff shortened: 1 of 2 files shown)


VojtechVitek (commented):
LGTM.

Built on top of @noazaj's commit/PR and fixed the tests. Thank you!

PR #1159, opened 2026-08-26
fix: panic on Mount() when a router is mounted onto itself

### LINKED ISSUE #843: Undefined Route Request Leads to Infinite Loop in route matching
I was implementing an HTTP server for my service and noticed that when I use 'chi' as shown below, the router does not respond to undefined routes. Additionally, I suspect the router gets stuck in an infinite loop, leading to a memory leak in the service.

```go
package main

import (
	"fmt"
	"github.com/go-chi/chi/v5"
	"net/http"
	_ "net/http/pprof"
)

func main() {
	r := chi.NewRouter().With(middleware.Logger)
	r.Mount("/", r.Group(func(r chi.Router) {
		r.Get("/ping", func(writer http.ResponseWriter, request *http.Request) {
			writer.WriteHeader(http.StatusOK)
			writer.Write([]byte("pong"))
		})
	}))

	fmt.Println("%v", http.ListenAndServe(":8080", r))
}
```

For the above-mentioned code, `GET /ping` request will be handled properly, but router never returns a response for an undefined request path.

sk-sanjay01:
## Problem

Mounting a router (or an inline mux derived from it via `With()`/`Group()`, which
shares the same underlying route tree) onto itself causes an infinite loop on any
request that doesn't match a defined route, instead of falling through to a 404:

```go
r := chi.NewRouter().With(middleware.Logger)
r.Mount("/", r.Group(func(r chi.Router) {
    r.Get("/ping", func(w http.ResponseWriter, r *http.Request) {
        w.Write([]byte("pong"))
    })
}))
```

`GET /ping` responds fine, but any undefined path hangs forever. Confirmed this is a
genuine busy-spin, not a deadlock: the process's CPU time climbed ~45s over ~3
wall-clock seconds while a single bad request was in flight, and a goroutine dump on
timeout shows the request cycling forever through
`Mux.ServeHTTP → mountHandler → Mux.ServeHTTP → routeHTTP → ...` with no state change.

Fixes #843.

## Root cause

`With()` and `Group()` return a new inline `*Mux` that shares the *same* underlying
`tree` pointer as the original router rather than copying it. So
`r.Mount("/", r.Group(fn))` wires a mount whose target handler routes through the
exact same tree the mount itself lives in.

In `Mux.ServeHTTP`, once a routing context already exists on the request (i.e. we're
inside a mount), it calls `mx.handler.ServeHTTP(w, r)` directly. `nextRoutePath()`
recomputes `RoutePath` as `"/" + <wildcard capture>` — for a mount at `"/"`, the
capture is the entire remaining path, so `RoutePath` ends up identical to what it was
before. Combined with the shared tree, a request that doesn't match `/ping` falls
through to the same `"/*"` mount node, calls the same `mountHandler`, recomputes the
same `RoutePath`, and re-enters routing with zero state change — an infinite loop
that pins one goroutine per bad request forever (a resource-exhaustion concern for a
router commonly used on public-facing services).

## Fix

`Mount()` now panics immediately — at registration time, not request time — when the
handler being mounted is a `*Mux` that shares the mounting router's tree:

```
chi: attempting to Mount() a router onto itself on '<pattern>'
```

This follows the same runtime-safety philosophy chi already uses elsewhere in
`Mount()` (e.g. panicking on a nil handler or on mounting over an existing pattern),
and mirrors the precedent set in #1148, which fixed a related `Mount()` edge case
with a similar "catch misuse loudly at registration time" approach plus a regression
test in the same style.

The check generalizes beyond the literal repro: it keys off tree-pointer identity, so
it also catches a plain `r.Mount("/", r)` self-mount, and any depth of `With()`/
`Group()` chaining, not just the specific pattern reported in the issue.

## Testing

- Added `TestMountingSelf` in `mux_test.go`, covering both the literal self-mount and
  the `With()`/`Group()` variant from the issue. Verified it fails (no panic) against
  the unpatched code and passes after the fix.
- Full test suite passes (`go test ./...` across `chi` and `chi/middleware`), `go vet`
  clean, `goimports -l` clean.
- Manually rebuilt the reporter's exact snippet against the patched module: it now
  panics immediately at startup instead of hanging on the first unmatched request.
- Verified no regressions: checked all `Mount()` call sites and `*Mux` construction
  paths in `_examples/`, `middleware/`, and the test suite — none share a tree pointer
  except the buggy self-mount case, so no existing legitimate usage trips the new
  panic.

### DIFF
--- a/mux.go
+++ b/mux.go
@@ -306,13 +306,9 @@ func (mx *Mux) Mount(pattern string, handler http.Handler) {
 	// Assign sub-Router's with the parent not found & method not allowed handler if not specified.
 	subr, ok := handler.(*Mux)
 
-	// Provide runtime safety for ensuring a router (or an inline mux derived
-	// from it via With()/Group(), which shares the same underlying tree) isn't
-	// mounted onto itself. Without this check, a request that doesn't match
-	// any defined route recurses through the same mount forever: routeHTTP
-	// falls through to the mount's wildcard, mountHandler hands the request
-	// back to a handler backed by the very same tree, and RoutePath never
-	// changes, so the routing never terminates. See issue #843.
+	// Provide runtime safety for ensuring a router isn't mounted onto itself,
+	// which includes an inline mux derived from it via With()/Group() since
+	// that shares the same underlying tree. See issue #843.
 	if ok && subr.tree == mx.tree {
 		panic(fmt.Sprintf("chi: attempting to Mount() a router onto itself on '%s'", pattern))
 	}
--- a/mux.go
+++ b/mux.go
@@ -305,6 +305,18 @@ func (mx *Mux) Mount(pattern string, handler http.Handler) {
 
 	// Assign sub-Router's with the parent not found & method not allowed handler if not specified.
 	subr, ok := handler.(*Mux)
+
+	// Provide runtime safety for ensuring a router (or an inline mux derived
+	// from it via With()/Group(), which shares the same underlying tree) isn't
+	// mounted onto itself. Without this check, a request that doesn't match
+	// any defined route recurses through the same mount forever: routeHTTP
+	// falls through to the mount's wildcard, mountHandler hands the request
+	// back to a handler backed by the very same tree, and RoutePath never
+	// changes, so the routing never terminates. See issue #843.
+	if ok && subr.tree == mx.tree {
+		panic(fmt.Sprintf("chi: attempting to Mount() a router onto itself on '%s'", pattern))
+	}
+
 	if ok && subr.notFoundHandler == nil && mx.notFoundHandler != nil {
 		subr.NotFound(mx.notFoundHandler)
 	}
--- a/mux_test.go
+++ b/mux_test.go
@@ -1513,6 +1513,43 @@ func TestMountingSimilarPattern(t *testing.T) {
 	}
 }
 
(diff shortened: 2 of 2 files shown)


mux.go
```diff
@@ -305,6 +305,18 @@ func (mx *Mux) Mount(pattern string, handler http.Handler) {
...
 	// Assign sub-Router's with the parent not found & method not allowed handler if not specified.
 	subr, ok := handler.(*Mux)
+
+	// Provide runtime safety for ensuring a router (or an inline mux derived
+	// from it via With()/Group(), which shares the same underlying tree) isn't
+	// mounted onto itself. Without this check, a request that doesn't match
+	// any defined route recurses through the same mount forever: routeHTTP
+	// falls through to the mount's wildcard, mountHandler hands the request
+	// back to a handler backed by the very same tree, and RoutePath never
+	// changes, so the routing never terminates. See issue #843.
```
VojtechVitek: Can you make this comment more concise please?

VojtechVitek (commented):
LGTM, minor feedback

mux.go
sk-sanjay01: Done — trimmed it down to match the style of the other checks in Mount(). Thanks!

VojtechVitek (approved):
LGTM

VojtechVitek:
Thank you!

PR #1163, opened 2026-08-30
Remove unused Context.parentCtx field

Kvnpsiddhartha:
## What

`Context.parentCtx`  has been assigned in `Mux.ServeHTTP` but has never been read.

This started as a doc-only fix for the field's stale comment. @VojtechVitek  asked in
review why not remove the field outright if it's unused - agreed, so this
now does that instead:

- Removes `parentCtx` from `Context`.
- Removes the assignment in `Mux.ServeHTTP`.
- Removes the nil-outs in `Reset` and `Clone`.
- Drops the now-dead parts of `TestContext_Clone` (the field init, the
  detachment assertion, and the `"context"` import that only existed for
  that assertion).

Checked for other open PRs/issues touching `parentCtx` and for whether
current `master` had already removed it — neither, so this should be the
complete cleanup. Also rebased onto `master` to pick up unrelated commits
that landed since this PR was opened.

No behaviour change: the field was write-only. `gofmt`/`go vet` clean,
`make test` (router + middleware, both `-race`) passes.


context.go
```diff
@@ -48,9 +48,9 @@ var (
 type Context struct {
 	Routes Routes
 
-	// parentCtx is the parent of this one, for using Context as a
-	// context.Context directly. This is an optimization that saves
-	// 1 allocation.
+	// parentCtx is the request's context, set in Mux.ServeHTTP. It is
+	// currently unused: it backed the directContext optimization that
+	// was reverted in #581.
 	parentCtx context.Context
```
VojtechVitek: if it's unused, why don't we remove it?

Kvnpsiddhartha:
@VojtechVitek  Thank you for your review; pushed the removal. Field, the ServeHTTP assignment, both
Reset/Clone nil-outs, and the now-dead bits of TestContext_Clone (including
the "context" import that only existed for that test) are gone.

Checked first for other open PRs/issues on this and for whether current
master had already picked it up; nothing did, so this should be the full
cleanup. Also rebased onto master to pick up the commits that landed since
I opened this. gofmt/go vet clean, `make test` (-race) passes.

PR #1170, opened 2026-09-01
fix(middleware): make Connection upgrade check token-aware and case-insensitive

AleksaMCode:
Updated `middleware.Recoverer` to treat the `Connection` header as a case-insensitive, comma-separated token list when detecting upgrade requests, instead of requiring an exact "Upgrade" string match. This avoids incorrectly writing `500` for valid upgrade handshakes such as `Connection: keep-alive, Upgrade` or lowercase variants.

RFC references:

- [RFC 6455: The WebSocket Protocol](https://datatracker.ietf.org/doc/html/rfc6455)
  - https://www.rfc-editor.org/rfc/rfc6455.html?#section-4.2.1
- [RFC 7230: HTTP/1.1 Message Syntax and Routing](https://datatracker.ietf.org/doc/html/rfc7230)
  - https://www.rfc-editor.org/info/rfc7230/#section-6.1


AleksaMCode:
@VojtechVitek could you take a look when you find some time. Thanks

PR #1172, opened 2026-09-07
fix(middleware): make Connection upgrade check token-aware and case-insensitive

favourthemaster:
Fixes #1170

**Problem**
`middleware.Recoverer` checked `r.Header.Get("Connection") != "Upgrade"` with exact string match. This incorrectly writes `500` for valid WebSocket upgrade handshakes such as:

- `Connection: keep-alive, Upgrade`
- `Connection: upgrade` (lowercase)
- `Connection: keep-alive, upgrade`
- Multiple header values

**Solution**
Parse `Connection` header per RFC 6455 §4.2.1 and RFC 7230 §6.1:

- Use `Header.Values("Connection")` to handle multiple headers
- Split each value by `,`
- `TrimSpace` + `EqualFold` for case-insensitive token match on `"upgrade"`
- Only skip `500` when upgrade token present; otherwise write 500

**Testing**
- `go test ./... ` PASS (chi 4.5s, middleware 26s) locally on go1.26.5
- Manual httptest: 8/8 cases PASS (exact, lowercase, keep-alive variants, empty, non-upgrade)
- RFC refs tested against real upgrade scenarios

Tested fix locally - ready for review


AleksaMCode:
This doesn't fix a PR (which doesn't make sense), but it's a slightly modified copy of  #1170

PR #1174, opened 2026-09-08
middleware: record response status when flushing

jakezwang:
Calling `Flush()` before writing a response commits HTTP 200, but `WrapResponseWriter.Status()` remains 0. This causes `RequestLogger` to log status 0 for streamed responses.

Use `maybeWriteHeader()` in all four flush-capable wrappers to record the implicit status while preserving an explicitly written status. Extend the regression tests to cover each wrapper, subsequent writes, and repeated flushes.

Validation: `make test` (including the race detector), `go vet ./...`, and `goimports`. A local HTTP/1.1 and HTTP/2 check also confirmed correct logging after an initial flush, including after 103 Early Hints.


VojtechVitek (approved):
LGTM. Thank you

PR #1176, opened 2026-09-10
fix(middleware): exact Accept-Encoding matching and honor q=0

### LINKED ISSUE #1069: Compress middleware: matchAcceptEncoding uses substring match and ignores q=0

shaikn6:
## What

Make `middleware`'s `matchAcceptEncoding` do exact coding-name matching and honor `q=0` in the `Accept-Encoding` header.

## Why / evidence

`middleware/compress.go` matched accepted encodings with `strings.Contains(v, encoding)`:

```go
func matchAcceptEncoding(accepted []string, encoding string) bool {
	for _, v := range accepted {
		if strings.Contains(v, encoding) {
			return true
		}
	}
	return false
}
```

Two bugs, reported in #1069:

1. Substring false positives. `Accept-Encoding: br` matches encoder `b`; `Accept-Encoding: bgzip` matches `gzip`.
2. `q=0` ignored. Per RFC 9110 section 12.5.3 a quality value of `0` means "not acceptable", but `strings.Contains("gzip;q=0", "gzip")` is `true`, so the middleware gzips a response the client explicitly refused.

Confirmed against `main` (ae6be74): a table test exercising `gzip;q=0`, `br`/`b`, and `bgzip`/`gzip` fails before the change and passes after.

## Change

`middleware/compress.go` only. `matchAcceptEncoding` now trims each header element, splits the coding name off the parameters on the first `;`, compares the name exactly, and (new helper `acceptsQValue`) parses the `q` parameter, treating `q=0` (any numeric spelling) as not acceptable. Single concern - no change to 405/`Allow` handling or `Content-Encoding` parsing.

## Verified

```
$ go test ./middleware/ -run TestMatchAcceptEncoding -v
--- PASS: TestMatchAcceptEncoding (0.00s)
PASS
ok  	github.com/go-chi/chi/v5/middleware

$ go test ./...
ok  	github.com/go-chi/chi/v5
ok  	github.com/go-chi/chi/v5/middleware

$ gofmt -l middleware/compress.go middleware/compress_test.go   # no output
$ go vet ./middleware/                                          # clean
```

Fixes #1069


Voyagerroc-Lab (approved):
The parser now matches a complete coding token and honors q=0, addressing both substring false positives and explicit refusal. The table tests cover exact names, whitespace, positive and zero quality values. The PR reports reproductions that fail on the base and pass after the change. Review performed with OpenAI Codex at the account owners request.

PR #1185, opened 2026-09-18
middleware: honor Discard when flushing

LeulTew:
`Discard()` still forwards `Flush()` to the original response writer, committing HTTP 200 before buffering middleware can set its final status and headers. Suppress the transport flush while retaining the wrapper's status bookkeeping in all four flush-capable wrappers.

Validated with Go 1.27.1 on Windows and Ubuntu, including Linux race checks. Go 1.24-1.26 and Windows race checks remain unverified.

<details>
<summary>Implementation and validation</summary>

This follows the `Discard` contract introduced in #926 and subsequently applied to `ReadFrom` in #1110: middleware can capture a response and write to the original writer after its handler returns.

The four existing `Flush` methods share a private helper that records the implicit status before checking discard mode. No public interface, constructor selection, byte-counting, or tee-write behavior changes.

The existing flush regression test now covers all four wrappers, implicit and explicit statuses, passthrough and tee controls, and discard with and without a tee. It checks repeated flushing through `http.ResponseController`, byte counts, captured bodies, and the caller's ability to write the final response.

On unchanged master, all 16 discard cases fail while the 16 passthrough/tee controls pass. They all pass with the fix. Separate local HTTP/1.1 and HTTP/2 checks exercise GET/HEAD, direct flushing and `ResponseController.Flush`, and informational responses; the caller's final status and post-handler header are retained only with the fix.

Passing checks:

- Full Go tests and `go vet` on Windows and Ubuntu.
- Direct `go test -race` and `go build` on Ubuntu.
- The writer/logger regression group repeated 25 times and the public HTTP checks repeated 10 times with a fixed shuffle seed.
- `goimports` v0.37.0 on the changed files and `git diff --check`.

All Go checks used `GOMAXPROCS=2`, `-p=1`, and, for tests, `-parallel=1`.

</details>


VojtechVitek (approved):
Looks great! Thank you.

LGTM

